Issue #16, opened 2026-06-26
Investigation/prototype: compile & run cognee-rs on WebAssembly (minimal-adapter config)


GraDKh:
# Investigation/prototype: can cognee-rs compile & run on WebAssembly in a minimal-adapter configuration?

## Summary

Investigate and prototype whether the cognee-rust pipeline can be compiled to
**WebAssembly** and run end-to-end (`add` → `cognify` → `search`) in a
**minimal adapter configuration** — i.e. picking the lightest pure-Rust /
remote-HTTP backend for each trait so the build contains no native C/C++ code.

This is a **spike**: the deliverable is a concrete answer ("it builds and runs
to *here*; these are the exact walls") plus a thin working prototype, not a
production WASM SDK. The output should be a short report (or roadmap doc) and a
minimal build target that we can iterate on.

## Motivation

A WASM build unlocks two things the project already cares about:

- **Edge / browser SDK** — cognee already targets edge devices (Android is a
  first-class target); a browser/WASM client extends that to web apps and
  serverless/WASI runtimes.
- **A "remote-thin-client" SDK** — run the pipeline logic in-browser against a
  remote [`cognee-http-server`](crates/http-server/) for the heavy backends.

## Background — what the dependency audit found

A prior audit (verified against `Cargo.lock`, not just `Cargo.toml`) mapped
which backends are WASM-clean and which are hard blockers. The
**trait-based architecture** ([docs/architecture.md](docs/architecture.md))
means each blocker is swappable in principle:

| Backend (trait) | Default impl | Native code | Pure-Rust / remote alternative present? |
|---|---|---|---|
| **Graph** (`GraphDBTrait`) | `lbug` / Ladybug | **C++** (`cxx`, `cxx-build`, `cmake`) | ❌ only `MockGraphDB` (in-memory, `testing`) or Postgres (network) |
| **Relational DB** (`IngestDb`/`SearchHistoryDb`/`DeleteDb`) | `sqlx-sqlite` | **C** (`libsqlite3-sys`) — incl. `sqlite::memory:` | ❌ no pure-Rust metadata store shipped |
| **Embedding** (`EmbeddingEngine`) | `ort` / ONNX | **C++** | ✅ `OpenAICompatibleEmbeddingEngine` / `OllamaEmbeddingEngine` (pure-Rust HTTP) |
[...]

PR #25, opened 2026-06-28
spike(wasm): Config 1 logic-only crates run on wasm32 (browser-verified)

rahulm820:
Addresses #16 (mirrored as topoteretes/cognee#3513 )
 
@GraDKh — this lands on your **preferred** target, not the fallback: it builds **and runs** on `wasm32-unknown-unknown` with JS glue (via `wasm-bindgen`), verified in a **real headless browser** — no WASI needed.
 
## 1. The Problem
 
WebAssembly (`wasm32-unknown-unknown` + JS, for the browser) is a target edge runtime for cognee-rs, but the workspace was never built for wasm and several foundational assumptions break there:
 
- **Randomness** — wasm has no OS RNG; `getrandom` (two major versions, via `rand 0.8` and `uuid`) fails to link without an explicit JS backend.
- **Async runtime** — `tokio`'s `rt-multi-thread`/`fs` features don't compile on wasm.
- **Filesystem** — code streaming local files via `tokio::fs` has no wasm equivalent.
- **Time** — `chrono::Utc::now()` traps as `unreachable` at *runtime* (invisible to the compiler).
- **Storage coupling** — `cognee-storage` is filesystem-shaped at the *trait* level.
- **Verification gap** — building isn't running; and running under Node isn't the same as running in a real browser, which is what the owner actually wants.
## 2. The Scope
 
**Config 1 = logic-only feasibility, on the owner's *preferred* target.** Prove the pure-logic crates (`cognee-utils`, `cognee-models`, `cognee-chunking`) **compile and run** on `wasm32-unknown-unknown` with JS glue (`wasm-bindgen`) — verified in a real headless browser, not just Node — using the smallest target-gated changes and **zero native-build impact**.
 
**Out of scope (Config 2, not done):** wasm-clean `cognee-storage` (decouple `StorageWriter` from `tokio::fs::File`), an in-memory relational store (`IngestDb`/`SearchHistoryDb`/`DeleteDb` over `HashMap`), an in-memory `GraphDBTrait`, a `reqwest` Fetch transport, and a `wasm` feature on `cognee-lib` mirroring the existing android-default pattern. `cognee-storage` is **excluded** from wasm here, not made wasm-clean.
 
## 3. The Solution — 6 walls, fixed minimally and target-gated
 
| # | Wall | Fix |
|---|------|-----|
| 1 | `getrandom` (two versions) | `js` feature (0.2 via `rand`) + `wasm_js` feature & `--cfg getrandom_backend="wasm_js"` (0.4 via `uuid`) |
| 2 | `uuid` needs wasm randomness | `uuid` `js` feature on wasm |
| 3 | `tokio` rejects `rt-multi-thread`/`fs`; workspace inheritance can't drop features | per-crate target split: native = workspace tokio, wasm = `["sync","macros","io-util","rt","time"]` |
| 4 | `cognee-models` streams local files via `tokio::fs` | cfg-gate the `FilePath` arm off wasm (returns `Unsupported`, mirroring `Url`/`S3Path`) |
| 5 | `cognee-chunking` → `cognee-storage` fs-coupled at trait level | cfg-gate `cognify_pipeline` + storage/tokio deps off wasm; pure primitives stay |
| 6 | Runtime trap: `chrono::Utc::now()` → `unreachable` | `chrono` `wasmbind` feature on wasm (routes through `js_sys::Date`) |
 
**Verification (the new increment):** added a `wasm-bindgen-test` smoke test that *runs* the chunking primitives + token counting in a wasm host. The assertion bodies are shared (`wasm_smoke/mod.rs`) by two runners — **Node** (`wasm.rs`) and **headless Chrome** (`wasm_browser.rs` via `wasm_bindgen_test_configure!(run_in_browser)`) — so they can't drift. The browser runner proves the owner's exact target: the wasm artifact + its JS glue executing in a real browser.
 
Every change is `cfg(target_arch = "wasm32")`-gated, so the native build is untouched by construction.
 
### In the first commit (`8fa4dde`, 7 files, already committed)
 
- `.cargo/config.toml` — wasm32 rustflags (`getrandom_backend`) + wasm test runner
- `crates/models/Cargo.toml` — tokio target split + chrono `wasmbind` on wasm (Wall 6)
- `crates/models/src/data_input.rs` — cfg-gate `FilePath` fs streaming off wasm
- `crates/utils/Cargo.toml` — getrandom/uuid/tokio wasm shims + tokio split
- `crates/chunking/Cargo.toml` — gate `cognee-storage` + tokio off wasm; wasm test dev-dep
- `crates/chunking/src/lib.rs` — gate `cognify_pipeline` off wasm
- `crates/chunking/tests/wasm.rs` — original Node smoke test
### This commit
 
**Modified:**
- `crates/chunking/tests/wasm.rs` — refactored to delegate to the shared module
**Added:**
- `crates/chunking/tests/wasm_browser.rs` — headless-browser runner (`run_in_browser`)
- `crates/chunking/tests/wasm_smoke/mod.rs` — shared assertion bodies for both runners
- `docs/spike-wasm-config1.md` — full spike report (walls 1–6, acceptance, Config-2 sizing) — *was never committed in `8fa4dde`; lands here*
## 5. Environment
 
- **Host:** WSL Ubuntu 24.04, x86_64. Rust toolchain pinned via `rust-toolchain.toml` (plain `cargo` selects it).
- **Build:** `wasm32-unknown-unknown` target (`rustup target add wasm32-unknown-unknown`).
- **Test (both runners):** `wasm-bindgen-cli` **0.2.126** — must match the locked `wasm-bindgen`.
- **Node runner:** Node on PATH (ran on 18 and 20; **20+ recommended** — an earlier wasm-bindgen/Node-18 combo crashed V8 on the externref glue).
- **Browser runner:** **Chrome for Testing 150.0.7871.24** + matching **chromedriver 150**, run headless; `CHROMEDRIVER` env points at the driver, `google-chrome` symlinked onto PATH for discovery; under WSL Chrome needs `--no-sandbox` (the runner's default headless args include it).
## 6. Test Results — all green
 
| Check | Result |
|-------|--------|
| wasm32 build (models, utils, chunking `--features tiktoken`) | ✅ |
| Native `cargo check` (all 4 crates) + `-p cognee-chunking --all-targets` | ✅ clean |
| `cargo fmt --check` | ✅ clean |
| Node runner — default / tiktoken | ✅ 2 passed / ✅ 3 passed |
| Headless Chrome runner — default / tiktoken | ✅ 2 passed / ✅ 3 passed |
 
### Reproduce locally
 
```bash
# build
cargo build -p cognee-models -p cognee-utils --target wasm32-unknown-unknown
cargo build -p cognee-chunking --no-default-features --features tiktoken \
    --target wasm32-unknown-unknown
 
# Node runner (needs wasm-bindgen-cli matching the locked version + Node 20+)
cargo test -p cognee-chunking --target wasm32-unknown-unknown --test wasm
cargo test -p cognee-chunking --features tiktoken \
    --target wasm32-unknown-unknown --test wasm
 
# Headless-browser runner (needs a WebDriver + browser; runs headless by default)
export CHROMEDRIVER=/path/to/chromedriver        # google-chrome on PATH
cargo test -p cognee-chunking --target wasm32-unknown-unknown --test wasm_browser
cargo test -p cognee-chunking --features tiktoken \
    --target wasm32-unknown-unknown --test wasm_browser
```
 
## Notes for reviewers
 
The browser test needs a WebDriver + browser present (Chrome for Testing + chromedriver in this spike). It's not wired into CI here — happy to add a setup step, or mark it `#[ignore]`-by-default, whichever you'd prefer.
 
cc @GraDKh — this is Config 1, now with browser verification included.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,10 +237,18 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install Rust (stable) + wasm32 target
+      - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: wasm32-unknown-unknown
+
+      # rust-toolchain.toml pins the repo to 1.91, and that pin takes precedence
+      # over @stable — so `cargo build --target wasm32-unknown-unknown` runs under
+      # 1.91, not stable. The wasm32 std component must be on the *pinned*
+      # toolchain. `rustup target add` honors the rust-toolchain.toml override
+      # (exactly as scripts/check_all.sh does locally); the action's `targets:`
+      # input adds it to @stable instead, which is why CI hit
+      # `error[E0463]: can't find crate for std`.
+      - name: Add wasm32 target to the pinned toolchain
+        run: rustup target add wasm32-unknown-unknown
 
       - name: Install Node (for the wasm-bindgen test runner)
         uses: actions/setup-node@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,12 +259,27 @@ jobs:
           cargo build -p cognee-models -p cognee-utils --target wasm32-unknown-unknown
           cargo build -p cognee-chunking --no-default-features --features tiktoken --target wasm32-unknown-unknown
 
-      # Drift guard: type-check every wasm test target (wasm.rs, wasm_browser.rs
-      # and the shared wasm_smoke module). --no-run needs no runner, so it works
-      # before wasm-bindgen-cli is installed. A renamed DocumentChunk field or a
-      # changed chunk_text signature breaks here instead of slipping through.
+      # Drift guard: type-check every wasm test target. --no-run needs no runner,
+      # so it works before wasm-bindgen-cli is installed.
       - name: Type-check wasm test targets (drift guard)
-        run: cargo test -p cognee-chunking --features tiktoken --target wasm32-unknown-unknown --no-run
+        run: |
+          # utils/models have no wasm-specific tests, but this compiles their wasm
+          # *test* build — which is what this PR's tokio dev-dep split (utils) and
+          # the cfg(not(wasm32)) gates on the retry (utils) and data_input (models)
+          # test modules exist to keep green. `cargo build` only covers the lib, so
+          # without this a regression to those gates stays invisible until someone
+          # runs the wasm tests by hand.
+          cargo test -p cognee-utils -p cognee-models --target wasm32-unknown-unknown --no-run
+          # chunking under BOTH feature configs: the shared wasm_smoke module has a
+          # #[cfg(feature = "tiktoken")] arm, and the no-feature build of
+          # tests/wasm.rs is otherwise only ever exercised by the live Node step
+          # below (which depends on the wasm-bindgen-cli install). A renamed
+          # DocumentChunk field or a changed chunk_text signature breaks here.
+          # Always spell chunking's wasm feature set explicitly with
+          # --no-default-features [--features tiktoken] — identical to the build
+          # step and check_all.sh — so the checks can't diverge if `default` grows.
+          cargo test -p cognee-chunking --no-default-features --target wasm32-unknown-unknown --no-run
+          cargo test -p cognee-chunking --no-default-features --features tiktoken --target wasm32-unknown-unknown --no-run
 
       # wasm-bindgen-test-runner must match the resolved wasm-bindgen crate
       # exactly or it refuses to run; read the version from the just-generated lock.
@@ -274,10 +289,10 @@ jobs:
           echo "Installing wasm-bindgen-cli $WB_VER"
           cargo install wasm-bindgen-cli --version "$WB_VER" --locked
 
-      - name: Run Node smoke tests (default + tiktoken)
+      - name: Run Node smoke tests (no-features + tiktoken)
         run: |
-          cargo test -p cognee-chunking --target wasm32-unknown-unknown --test wasm
-          cargo test -p cognee-chunking --features tiktoken --target wasm32-unknown-unknown --test wasm
+          cargo test -p cognee-chunking --no-default-features --target wasm32-unknown-unknown --test wasm
+          cargo test -p cognee-chunking --no-default-features --features tiktoken --target wasm32-unknown-unknown --test wasm
 
   # ── Stage 2: Build and run tests (depends on lint) ────────────────────
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,6 +222,63 @@ jobs:
       - name: Check (MSRV)
         run: cargo +1.91.0 check --all-targets
 
+  # ── wasm32 Config-1 lane: logic crates compile + run on wasm32 ────────
+  # The wasm smoke-test files are #![cfg(target_arch = "wasm32")], so the native
+  # lint/check lanes compile them to empty crates and never type-check them.
+  # This lane builds the logic crates for wasm32-unknown-unknown, type-checks
+  # every wasm test target (so a DocumentChunk-field or chunk_text-signature
+  # drift fails CI), and runs the Node smoke test. The headless-browser runner
+  # stays manual — it needs a WebDriver + browser; see docs/spike-wasm-config1.md.
+  # No protoc/cmake/lbug here: the three logic crates pull none of them.
+  wasm:
+    name: wasm32 (Config-1 logic crates)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Rust (stable) + wasm32 target
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+
+      - name: Install Node (for the wasm-bindgen test runner)
+        uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      - name: Cache cargo/target (wasm)
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: wasm-v1
+
+      # Build first so Cargo.lock exists (it is gitignored) before we pin the
+      # runner to the resolved wasm-bindgen version below.
+      - name: Build logic crates (wasm32)
+        run: |
+          cargo build -p cognee-models -p cognee-utils --target wasm32-unknown-unknown
+          cargo build -p cognee-chunking --no-default-features --features tiktoken --target wasm32-unknown-unknown
+
+      # Drift guard: type-check every wasm test target (wasm.rs, wasm_browser.rs
+      # and the shared wasm_smoke module). --no-run needs no runner, so it works
+      # before wasm-bindgen-cli is installed. A renamed DocumentChunk field or a
+      # changed chunk_text signature breaks here instead of slipping through.
+      - name: Type-check wasm test targets (drift guard)
+        run: cargo test -p cognee-chunking --features tiktoken --target wasm32-unknown-unknown --no-run
+
+      # wasm-bindgen-test-runner must match the resolved wasm-bindgen crate
+      # exactly or it refuses to run; read the version from the just-generated lock.
+      - name: Install wasm-bindgen-test-runner (matching the locked version)
+        run: |
+          WB_VER=$(sed -n '/^name = "wasm-bindgen"$/{n;s/^version = "\(.*\)"$/\1/p}' Cargo.lock)
+          echo "Installing wasm-bindgen-cli $WB_VER"
+          cargo install wasm-bindgen-cli --version "$WB_VER" --locked
+
+      - name: Run Node smoke tests (default + tiktoken)
+        run: |
+          cargo test -p cognee-chunking --target wasm32-unknown-unknown --test wasm
+          cargo test -p cognee-chunking --features tiktoken --target wasm32-unknown-unknown --test wasm
+
   # ── Stage 2: Build and run tests (depends on lint) ────────────────────
   test:
     name: Test (stable)
--- a/crates/chunking/tests/wasm_smoke/mod.rs
+++ b/crates/chunking/tests/wasm_smoke/mod.rs
@@ -1,0 +1,53 @@
+//! Shared smoke-test bodies for the Config-1 wasm spike.
+//!
+//! These plain functions hold the actual assertions; they are wrapped with
+//! `#[wasm_bindgen_test]` by both runners:
+//!
+//! * `wasm.rs`         — executes them under **Node** (the default runner).
+//! * `wasm_browser.rs` — executes them in a **headless browser** (via
+//!   `wasm_bindgen_test_configure!(run_in_browser)` + a WebDriver).
+//!
+//! Keeping the bodies here means the two harnesses can never drift apart. This
+//! file lives in a `tests/` subdirectory, so cargo does not treat it as its own
+//! test target — it is only compiled when `mod`-included by a wasm test file.
+
+use cognee_chunking::{NAMESPACE_OID, TokenCounter, WordCounter, chunk_text};
+
+pub fn word_counter() {
+    assert_eq!(WordCounter.count_tokens("hello wasm world"), 3);
+    assert_eq!(WordCounter.count_tokens(""), 0);
+}
+
+pub fn chunk_text_smoke() {
+    let counter = WordCounter;
+    // NAMESPACE_OID is a valid Uuid; reuse it as the document id so the test
+    // needs no direct uuid dependency.
+    let doc = NAMESPACE_OID;
+    let text = "First paragraph of the spike.\n\n\
+                Second paragraph has a few more words than the first one does.";
+
+    let chunks = chunk_text(doc, text, 8, &counter);
+
+    assert!(!chunks.is_empty(), "expected at least one chunk");
+    for (i, c) in chunks.iter().enumerate() {
+        assert_eq!(
+            c.chunk_index, i,
+            "chunks must be indexed sequentially from 0"
+        );
+        assert!(!c.text.is_empty(), "chunk text should be non-empty");
+        assert!(c.chunk_size > 0, "chunk size should be counted");
+        assert_eq!(c.document_id, doc, "chunk should carry its document id");
+    }
+}
+
+#[cfg(feature = "tiktoken")]
+pub fn tiktoken_counter() {
+    use cognee_chunking::TikTokenCounter;
+
+    // The cl100k_base BPE tables are bundled in the binary (pure Rust) — this
+    // exercises that they load and encode under wasm with no filesystem/network.
+    let counter = TikTokenCounter::cl100k_base().expect("cl100k_base BPE loads in wasm");
+    let n = counter.count_tokens("Hello, world!");
+    assert!((3..=6).contains(&n), "expected 3-6 cl100k tokens, got {n}");
+    assert_eq!(counter.count_tokens(""), 0);
+}
--- a/crates/models/Cargo.toml
+++ b/crates/models/Cargo.toml
@@ -20,4 +20,18 @@ schemars.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 uuid.workspace = true
+
+# tokio: the workspace dep carries rt-multi-thread + fs, which fail to compile on
+# wasm32-unknown-unknown (the runtime has no OS threads or filesystem). Workspace
+# inheritance can only add features, not drop them, so wasm gets a direct tokio
+# spec limited to the wasm-supported set. See docs/spike-wasm-config1.md.
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio.workspace = true
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+tokio = { version = "1", default-features = false, features = ["sync", "macros", "io-util", "rt", "time"] }
+# chrono's `Utc::now()` (used by DataPoint::new) calls the system clock, which
+# traps as `unreachable` on wasm32. The `wasmbind` feature routes it through
+# `js_sys::Date` instead. The feature is additive, so the base `chrono.workspace`
+# dep just gains it via unification on wasm. See docs/spike-wasm-config1.md.
+chrono = { workspace = true, features = ["wasmbind"] }
--- a/crates/models/Cargo.toml
+++ b/crates/models/Cargo.toml
@@ -21,15 +21,15 @@ serde.workspace = true
 serde_json.workspace = true
 uuid.workspace = true
 
-# tokio: the workspace dep carries rt-multi-thread + fs, which fail to compile on
-# wasm32-unknown-unknown (the runtime has no OS threads or filesystem). Workspace
-# inheritance can only add features, not drop them, so wasm gets a direct tokio
-# spec limited to the wasm-supported set. See docs/spike-wasm-config1.md.
+# tokio is only used here for local-filesystem streaming (data_input.rs), which is
+# cfg'd out on wasm32 — so wasm needs no tokio at all. Gate the dep to non-wasm:
+# the workspace tokio carries rt-multi-thread + fs, neither of which compiles on
+# wasm32-unknown-unknown, and a plain (ungated) dep would try to build them there.
+# See docs/spike-wasm-config1.md.
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio.workspace = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-tokio = { version = "1", default-features = false, features = ["sync", "macros", "io-util", "rt", "time"] }
 # chrono's `Utc::now()` (used by DataPoint::new) calls the system clock, which
 # traps as `unreachable` on wasm32. The `wasmbind` feature routes it through
 # `js_sys::Date` instead. The feature is additive, so the base `chrono.workspace`
--- a/crates/utils/Cargo.toml
+++ b/crates/utils/Cargo.toml
@@ -21,33 +21,33 @@ tracing.workspace = true
 # Random (for jitter)
 rand.workspace = true
 
-# UUID generation
+# Cross-platform async timer for retry backoff (retry.rs). Works on both native
+# and wasm32: its `wasm-bindgen` feature routes through setTimeout on wasm and is
+# a no-op on native (which uses a timer thread). This is why retry needs no tokio
+# and no in-code cfg — tokio's timer would never fire on wasm32-unknown.
+futures-timer.workspace = true
+
+# UUID generation. The `js` wasm randomness source is enabled on the workspace
+# uuid dep (a no-op on native), so v4 works on wasm with no per-crate override.
 uuid = { workspace = true, features = ["v4", "v5"] }
 
 # Secret redaction (cognee_utils::redact)
 regex = { workspace = true }
 
-# tokio: see crates/models/Cargo.toml — the workspace dep's rt-multi-thread + fs
-# features don't compile on wasm32, and inheritance can't drop them, so wasm gets
-# a direct tokio spec limited to the wasm-supported feature set.
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-tokio.workspace = true
-
-# WebAssembly (browser) spike — Config 1. On wasm32-unknown-unknown the default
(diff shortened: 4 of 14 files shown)


crates/utils/Cargo.toml
```diff
@@ -30,5 +27,27 @@ uuid = { workspace = true, features = ["v4", "v5"] }
...
+# WebAssembly (browser) spike — Config 1. On wasm32-unknown-unknown the default
+# getrandom backends are unavailable, so route both versions in our tree through
+# the JS crypto API. These are feature-only shims (the crates are already pulled
+# transitively — getrandom 0.2 via rand 0.8, getrandom 0.4 via uuid); declaring
+# them here lets Cargo's feature unification turn the JS backends on. getrandom
+# 0.4 additionally needs the getrandom_backend="wasm_js" cfg, set in
+# .cargo/config.toml. See docs/spike-wasm-config1.md.
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2", features = ["js"] }
+getrandom_v04 = { package = "getrandom", version = "0.4", features = ["wasm_js"] }
```
GraDKh: `getrandom_v04` (getrandom 0.4) isn't actually in the wasm dependency graph. uuid 1.23 gates its `getrandom`/`rand` deps to non-wasm and switches to `uuid-rng-internal-lib` + `wasm-bindgen` on wasm32, enabled purely by uuid's `js` feature. So this line *forces* an otherwise-absent crate (plus its wasm-bindgen) into the wasm build, and the matching `--cfg getrandom_backend="wasm_js"` in `.cargo/config.toml` is read by nobody (getrandom 0.2 predates that cfg system; 0.4 isn't present).

Verified with `cargo tree --target wasm32-unknown-unknown`: models pulls only `uuid`; utils/chunking pull `uuid` + `rand 0.8 -> getrandom 0.2.17`. No getrandom 0.4.

Suggestion: drop `getrandom_v04` and the `getrandom_backend` rustflag. Keep `getrandom 0.2 { features = ["js"] }` (real — needed by the `rand` jitter). And move uuid's `js` onto the **workspace** uuid dependency rather than declaring it here: it's a no-op on native (the wasm-bindgen/js-sys deps are target-gated inside uuid), and at workspace level every crate inherits it — which also fixes the models gap noted on its Cargo.toml.

crates/utils/Cargo.toml
```diff
@@ -30,5 +27,27 @@ uuid = { workspace = true, features = ["v4", "v5"] }
...
+# uuid 1.23 refuses to build on wasm32 without an explicit randomness source.
+# `js` routes v4 generation through the browser crypto API; feature unification
+# applies this to every crate sharing this uuid in the wasm build.
+uuid = { workspace = true, features = ["v4", "v5", "js"] }
+tokio = { version = "1", default-features = false, features = ["sync", "macros", "io-util", "rt", "time"] }
```
GraDKh: utils' only non-test tokio use is `tokio::time::sleep` in `retry.rs:182`, so this entire wasm tokio spec exists just to compile that one line. But tokio's `time`/`rt` don't actually work on wasm32-unknown-unknown — no thread parking, no clock — so the timer silently won't fire. `retry_with_backoff` is compile-only on the Config-1 surface today (its only runtime caller is the ingestion URL fetcher, which is out of scope), so it's not a live bug yet, but it's a Config-2 landmine.

Suggestion: replace `tokio::time::sleep` with a cross-platform timer — `futures_timer::Delay::new(delay).await` (its `wasm-bindgen` feature is a native no-op, so no in-code cfg needed), or `wasmtimer::tokio::sleep` as an API-identical drop-in. Then utils needs no tokio on wasm at all and this block can be removed; tokio reverts to a native-only dependency.

crates/utils/Cargo.toml
```diff
@@ -30,5 +27,27 @@ uuid = { workspace = true, features = ["v4", "v5"] }
...
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
```
GraDKh, line 53: This dev-dependency wasn't target-split like the others, so it still pulls `rt-multi-thread` on wasm — `cargo test -p cognee-utils --target wasm32-unknown-unknown` would fail to compile. Not triggered by the current build-only scope for utils, but worth gating it the way `crates/chunking/Cargo.toml` splits its dev-deps, for consistency.

crates/models/Cargo.toml
```diff
@@ -20,4 +20,18 @@ schemars.workspace = true
...
+
+# tokio: the workspace dep carries rt-multi-thread + fs, which fail to compile on
+# wasm32-unknown-unknown (the runtime has no OS threads or filesystem). Workspace
+# inheritance can only add features, not drop them, so wasm gets a direct tokio
+# spec limited to the wasm-supported set. See docs/spike-wasm-config1.md.
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio.workspace = true
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+tokio = { version = "1", default-features = false, features = ["sync", "macros", "io-util", "rt", "time"] }
```
GraDKh: Two things on this wasm block:

1. **It's dead.** On wasm, models uses zero tokio — the only uses (`tokio::fs::File` / `AsyncReadExt` in the `FilePath` arm) are already cfg'd off. `process_by_chunks` is `async` but runtime-agnostic. So this `[target.'cfg(target_arch = "wasm32")']` tokio spec can be removed and tokio made non-wasm-only — fewer cfgs.

2. **Missing uuid `js`.** `DataPoint::new` calls `Uuid::new_v4()` (`data_point.rs:101`), which needs uuid's `js` feature on wasm. It's declared only in `cognee-utils`, and models doesn't depend on utils — so models' wasm build only links because the repro command co-compiles both and feature unification leaks the feature in. A standalone wasm build of models wouldn't get it. Cleanly fixed by moving `js` to the workspace uuid dep (see the utils note).

The `chrono` `wasmbind` line here is correct and should stay.

crates/chunking/tests/wasm_smoke/mod.rs
```diff
@@ -0,0 +1,53 @@
+//! Shared smoke-test bodies for the Config-1 wasm spike.
```
GraDKh, line 1: These wasm test files are all `#![cfg(target_arch = "wasm32")]`, so native CI (`scripts/check_all.sh`, `ci.yml`) compiles them to empty crates and never type-checks them. This shared module references concrete `DocumentChunk` fields (`chunk_index`, `chunk_size`, `document_id`, …) and `chunk_text`'s signature — if those drift, native CI stays green and the break only surfaces when someone manually runs the wasm-bindgen runner. Worth wiring at least the Node runner (or a `cargo check --target wasm32-unknown-unknown`) into CI to catch drift, even if the headless-browser runner stays manual.

GraDKh (commented):
Thanks for the spike — it's clearly documented and the cfg-gating genuinely keeps the native build untouched. A few concerns posted inline on the relevant lines. The dependency claims below were checked against the actual `getrandom`/`uuid` manifests and `cargo tree --target wasm32-unknown-unknown`.

Themes: (1) the getrandom-0.4 machinery is unnecessary because uuid 1.23 handles its own wasm randomness; (2) most of the per-crate wasm `tokio` blocks are dead — only `retry.rs`'s `sleep` actually needs a timer, and tokio's timer doesn't work on wasm anyway; (3) a couple of consistency/coverage gaps. Net, several of these point toward *fewer* cfgs than the current PR.

GraDKh (changes requested):

rahulm820:
Thank you for guidance @GraDKh 
I've read through the requirements and have a good understanding of the changes needed.
I'm currently working on this and expect to have an update within the next few hours.

rahulm820:
Hey @GraDKh 
# Requested Changes — prune dead cfgs, fix correctness gaps, wire wasm into CI 
 
This round addresses `@GraDKh`'s review feedback on the Config-1 wasm spike: prune the wasm `cfg`s that turned out unnecessary, fix two latent correctness gaps, and wire wasm into CI so it can't silently drift.
 
**Net direction: fewer cfgs, more coverage.**
 
## Files changed + why
 
| File | Change | Reason |
|---|---|---|
| `Cargo.toml` | Added `futures-timer` (with `wasm-bindgen` feature); added `js` to the workspace `uuid` dep | `js` at workspace level is a native no-op (`uuid` target-gates its `wasm-bindgen`/`js-sys` deps) and makes every crate's `uuid` v4 work on wasm — fixes the `models` gap. `futures-timer` replaces `tokio` for retry backoff. |
| `crates/utils/Cargo.toml` | Removed `getrandom_v04` shim + per-crate `uuid` override; swapped lib `tokio` → `futures-timer`; target-split the `tokio` dev-dep off wasm | `getrandom` 0.4 was never in the real wasm graph (only our shim pulled it). `uuid` `js` now comes from workspace. `tokio` is no longer a library dep at all. dev-dep split stops `rt-multi-thread` from breaking `cargo test --target wasm32`. |
| `.cargo/config.toml` | Removed the `getrandom_backend="wasm_js"` rustflag | Read by nobody — `getrandom` 0.2 predates that cfg system and 0.4 isn't present. |
| `crates/models/Cargo.toml` | Removed the dead `cfg(wasm32)` tokio block (kept `chrono` `wasmbind`) | `models` uses zero tokio on wasm (its only tokio is the cfg'd-off `FilePath` arm). |
| `crates/utils/src/retry.rs` | `tokio::time::sleep` → `futures_timer::Delay::new(delay).await`; gated the `#[tokio::test]` module with `cfg(all(test, not(target_arch = "wasm32")))` | tokio's timer never fires on wasm (no clock/parking) — `futures-timer` works on both and removes the Config-2 landmine. Gating the test module keeps `cargo test --target wasm32` compiling. |
| `.github/workflows/ci.yml` | New wasm job: build logic crates → `--no-run` drift guard → install version-matched `wasm-bindgen-cli` → run Node smoke tests | The `#![cfg(wasm32)]` test files were invisible to native CI; this catches `DocumentChunk`/`chunk_text` drift and exercises the runtime. |
| `scripts/check_all.sh` | Build-only wasm drift guard (`--no-run`, no Node needed) | Same drift protection locally, without requiring `wasm-bindgen-cli`. |

 
## New functions added
 
None. This round is dependency/config/CI changes plus a single one-line implementation swap in `retry.rs` (`tokio::time::sleep` → `futures_timer::Delay`). No new public or private functions.
 
## Verification (all green)
 
- wasm builds (`models`, `utils`, `chunking`+`tiktoken`); native check + fmt clean
- Node tests 2/3; headless Chrome tests 2/3
- standalone `models` wasm build proves `uuid` `js` without `utils` unification
- native `utils` tests 24 passed (incl. `test_exponential_backoff_timing`)
- `cargo test -p cognee-utils --target wasm32` now compiles (was failing)
- drift guard negative-tested: a broken field reference fails CI as intended

.github/workflows/ci.yml
```diff
@@ -222,6 +222,63 @@ jobs:
...
+        run: |
+          cargo build -p cognee-models -p cognee-utils --target wasm32-unknown-unknown
+          cargo build -p cognee-chunking --no-default-features --features tiktoken --target wasm32-unknown-unknown
+
+      # Drift guard: type-check every wasm test target (wasm.rs, wasm_browser.rs
+      # and the shared wasm_smoke module). --no-run needs no runner, so it works
+      # before wasm-bindgen-cli is installed. A renamed DocumentChunk field or a
+      # changed chunk_text signature breaks here instead of slipping through.
+      - name: Type-check wasm test targets (drift guard)
+        run: cargo test -p cognee-chunking --features tiktoken --target wasm32-unknown-unknown --no-run
```
GraDKh: The drift guard only covers `cognee-chunking`, but this PR also fixed `utils`'s test layer for wasm — the `tokio` dev-dep split + the `#[cfg(all(test, not(target_arch = "wasm32")))]` gate on `retry.rs`'s test module — specifically so that `cargo test -p cognee-utils --target wasm32-unknown-unknown` compiles (per the PR's Verification section).

Nothing here exercises that: the build step above only does `cargo build -p cognee-utils`, which compiles the **lib only** — not test targets or dev-deps. So a future PR that re-adds an ungated `tokio` dev-dep to `utils`, or drops the `not(target_arch = "wasm32")` from the retry test module, still passes `cargo build` and CI stays green — the fix silently regresses, invisible until someone manually builds the wasm tests.

Consider widening the guard to the crates whose test layer changed (and mirroring it in `scripts/check_all.sh`):

```
cargo test -p cognee-utils -p cognee-models --target wasm32-unknown-unknown --no-run
```

(`utils`/`models` have no wasm-specific *tests*, so this guards "the wasm test build keeps compiling" rather than any assertion — lower-stakes than the chunking guard, but it's the one place a fix made in this PR isn't protected by the CI this PR adds.)

---

Related, on the same step: the guard only type-checks the `--features tiktoken` build. The default-feature (no-tiktoken) build of `tests/wasm.rs` is never type-checked offline — it's only ever *run* in the live Node step below, which depends on the `wasm-bindgen-cli` install succeeding. A change that compiles under `tiktoken` but breaks the no-feature build of `wasm.rs` slips past this `--no-run` guard. Adding a no-feature `--no-run` pass (or just dropping `--features tiktoken` here, since the drift target is the shared `wasm_smoke` module that doesn't need it) would close that too.

.github/workflows/ci.yml
```diff
@@ -222,6 +222,63 @@ jobs:
...
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: wasm-v1
+
+      # Build first so Cargo.lock exists (it is gitignored) before we pin the
+      # runner to the resolved wasm-bindgen version below.
+      - name: Build logic crates (wasm32)
+        run: |
+          cargo build -p cognee-models -p cognee-utils --target wasm32-unknown-unknown
+          cargo build -p cognee-chunking --no-default-features --features tiktoken --target wasm32-unknown-unknown
```
GraDKh, line 268: Minor consistency nit: `cognee-chunking` is invoked with three different feature spellings across the wasm checks:

- this build step — `--no-default-features --features tiktoken`
- the drift guard just below (L267) — `--features tiktoken`
- `scripts/check_all.sh` — `--no-default-features --features tiktoken`

They're equivalent today only because `cognee-chunking`'s `default = []`. The moment anything is added to that crate's `default` feature list, the three diverge — the local `check_all.sh` check and the CI build would compile a different feature set than the drift guard, so a config could pass locally and fail in CI (or vice-versa). Worth spelling them identically (pick one form) so they can't drift.

docs/spike-wasm-config1.md
```diff
@@ -0,0 +1,323 @@
...
+- **0.4** (post-0.3 backend system) — enable the `wasm_js` cargo feature **and**
+  set `--cfg getrandom_backend="wasm_js"` in `RUSTFLAGS`.
+
+**Fix** (`crates/utils/Cargo.toml`, wasm32 target block + `.cargo/config.toml`):
+
+```toml
+# crates/utils/Cargo.toml
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+getrandom = { version = "0.2", features = ["js"] }
+getrandom_v04 = { package = "getrandom", version = "0.4", features = ["wasm_js"] }
```
GraDKh: This doc still prescribes config that the latest revision of this PR **removed** — Wall 1 contradicts the shipped code:

- `getrandom_v04 = { package = "getrandom", version = "0.4", features = ["wasm_js"] }` (L84) is gone from `crates/utils/Cargo.toml` (only `getrandom 0.2 { js }` remains).
- `rustflags = ['--cfg', 'getrandom_backend="wasm_js"']` (L90) is gone from `.cargo/config.toml` — which now carries only `runner = "wasm-bindgen-test-runner"`, and whose own comment explicitly states the 0.4 backend-cfg is unnecessary because `uuid 1.23` doesn't pull `getrandom` on `wasm32-unknown`.
- The "Files changed" table (L231) likewise still credits `.cargo/config.toml` with "wasm32 rustflags (getrandom_backend)".

As written, Wall 1 tells the next contributor to add a dep and a `RUSTFLAGS` cfg that this PR deliberately dropped — and those rustflags would apply to the plain `cargo build` path too. Please reconcile Wall 1 (and the Files-changed table) with the shipped `.cargo/config.toml` / `utils/Cargo.toml`: drop the `getrandom_v04` dep + `getrandom_backend` rustflag, and keep just `getrandom 0.2 { js }` (with a note that uuid handles its own wasm randomness via the workspace `js` feature).

GraDKh (changes requested):
Thanks for the thorough follow-up round @rahulm820 — this is in great shape. I re-reviewed the full diff from scratch and against the previous round's feedback, and all five of @GraDKh's earlier comments are correctly applied: the dead `getrandom_v04` shim + `getrandom_backend` rustflag are gone, `uuid`'s `js` moved to the workspace dep, the retry timer is now the cross-platform `futures_timer::Delay`, the `models` wasm-tokio block is removed, and the wasm test layer is wired into CI. I independently confirmed the load-bearing dependency claims too — both `futures-timer`'s `wasm-bindgen` feature and `uuid`'s `js` feature are genuine native no-ops (their extra deps are `cfg(target_arch = "wasm32")`-gated), so native builds are unaffected.

The changes are minimal, well-scoped, and consistently `cfg`-gated, and no correctness or runtime issues survived review. I'm marking this **changes requested** only for a few CI-coverage / consistency cleanups (left inline) — none are blockers, and once they're addressed this is good to merge:

- **Widen the wasm drift guard** beyond `cognee-chunking` so the `utils`/`models` test-layer fixes this PR added (the `tokio` dev-dep split + the `retry.rs` test-module gate) can't silently regress — `cargo build` only checks the libs. Same comment notes the guard currently type-checks only the `tiktoken` build, leaving the default-feature build of `tests/wasm.rs` unchecked offline.
- **Spell the `cognee-chunking` feature flags identically** across the CI build step, the drift guard, and `check_all.sh` (they're equivalent today only because `default = []`).
- **Reconcile the spike doc's Wall 1** (and its "Files changed" table) with the shipped config — it still prescribes the `getrandom_v04` dep + `getrandom_backend` rustflag this PR removed.

Really nice spike — the browser-verified acceptance and the shared `wasm_smoke` module are a clean touch. 🚀

rahulm820:
Hey @GraDKh, I've updated the requested changes. I think we're good to go now. If any further changes are needed, I'd be glad to make them.

## C1 — "the drift guard doesn't protect the fixes this PR made"
 
The guard only covered `cognee-chunking`. The `utils`/`models` test-layer changes (tokio dev-dep split, `cfg(not(wasm32))` gates) weren't exercised — `cargo build` only compiles the lib, so a regression would stay green until someone ran the wasm tests by hand. Also, only the `tiktoken` config was type-checked, not the no-feature build of `wasm.rs`.
 
| Change | File | Why |
|---|---|---|
| Gate the one `#[tokio::test]` off wasm with `#[cfg(not(target_arch = "wasm32"))]` (sync `#[test]`s stay) | `crates/models/src/data_input.rs` | Widening the guard to `models` exposed that `cargo test -p cognee-models --target wasm32 --no-run` didn't compile (ungated async test, no tokio on wasm). Surgical fix; runs on native. |
| Add `cargo test -p cognee-utils -p cognee-models … --no-run`; split chunking into no-feature and tiktoken `--no-run` passes | `.github/workflows/ci.yml` | Guards the `utils`/`models` test-build this PR fixed; covers the no-feature `wasm.rs` build previously only run live. |
| Mirror both | `scripts/check_all.sh` | Same protection locally, no Node/`wasm-bindgen-cli`. |
 
## C2 — "chunking is spelled three different ways"
 
`--no-default-features --features tiktoken` (build), `--features tiktoken` (guard), `--no-default-features --features tiktoken` (check_all) — equivalent only because `default = []`. If `default` grows, the checks diverge and a config could pass locally but fail in CI.
 
| Change | File | Why |
|---|---|---|
| Standardize all 7 chunking wasm invocations on `--no-default-features [--features tiktoken]` | `.github/workflows/ci.yml`, `scripts/check_all.sh` | One canonical form; can't diverge. Chose `--no-default-features` because Config-1's intent is the minimal wasm-safe set. |

GraDKh (changes requested):
Thanks for the thorough spike and the detailed write-up — the target-gating is clean and the Node + headless-browser verification is a nice touch. A few things to address before merge: the new `wasm` CI lane is currently failing, plus two smaller items.

## 1. CI failure (blocking) — `wasm32` lane can't find `std`

The `wasm32 (Config-1 logic crates)` job fails at the first build step:

```
error[E0463]: can't find crate for `std`
error: could not compile `once_cell` (lib) due to 1 previous error
```

**Cause:** `rust-toolchain.toml` pins the repo to `channel = "1.91"`, and (as the MSRV lane comment notes) that pin takes precedence over `@stable`. The job installs the wasm32 `std` via the action's `targets:` input — but that lands it on **stable**, while every `cargo build --target wasm32-unknown-unknown` actually runs under the pinned **1.91** toolchain, which has no wasm32 `std`. It passes locally only because `scripts/check_all.sh` runs `rustup target add wasm32-unknown-unknown`, which adds the target to the *active/pinned* toolchain — the CI lane has no equivalent.

**Fix** — add the target to the pinned toolchain explicitly (mirroring `check_all.sh`), instead of relying on the `targets:` input:

```yaml
      - name: Install Rust toolchain
        uses: dtolnay/rust-toolchain@stable

      # rust-toolchain.toml pins the repo to 1.91, and that pin takes precedence
      # over @stable — so `cargo build --target wasm32-unknown-unknown` runs under
      # 1.91, not stable. The wasm32 std component must be on the *pinned*
      # toolchain. `rustup target add` honors the rust-toolchain.toml override
      # (exactly as scripts/check_all.sh does locally); the action's `targets:`
      # input adds it to @stable instead, which is why CI hit
      # `error[E0463]: can't find crate for std`.
      - name: Add wasm32 target to the pinned toolchain
        run: rustup target add wasm32-unknown-unknown
```

(Alternatively, pin the action to `dtolnay/rust-toolchain@1.91.0` with `targets: wasm32-unknown-unknown`, matching the MSRV lane — but the `rustup target add` form auto-tracks the pin if it ever moves and matches the already-working local script, so I'd prefer it.)

## 2. Broken intra-doc link on a wasm doc build — `crates/chunking/src/lib.rs:7`

The module doc links `[`cognify_pipeline`]`, but the PR now gates `pub mod cognify_pipeline` behind `#[cfg(not(target_arch = "wasm32"))]`. On a wasm doc build (`cargo doc --target wasm32-unknown-unknown -p cognee-chunking`) this becomes an `unresolved link` error. It doesn't break current CI (the docs job runs native, `-W missing_docs` only), but it would surface on a docs.rs/wasm build. Simplest fix — drop the link brackets so it's a plain code span:

```rust
//! - [`text_chunker`] / `cognify_pipeline` — the chunking entry points
```

## 3. CI efficiency (non-blocking) — `wasm-bindgen-cli` compiled from source

The `cargo install wasm-bindgen-cli` step compiles from source on every run (the `Swatinem/rust-cache` key covers the project target dir, not `~/.cargo/bin`), adding several minutes per run. Once the lane is green, consider a prebuilt fetch (e.g. `taiki-e/install-action` with `wasm-bindgen-cli@<locked-version>`). Not required for merge — just a nice follow-up.

---

Items 1 and 2 should land before merge (1 is blocking the lane; 2 is a latent doc breakage); 3 is optional. Could you apply these and push? Happy to re-review once the wasm lane is green. Thanks again!

rahulm820:
Hey @GraDKh 
# Updates — Pre-Merge Review Fixes (Pipeline Failure)
  
## Summary of Changes
 
| # | Why requested | What changed | File |
|---|---------------|--------------|------|
| 1 (blocking) | The wasm32 CI lane failed at the first build with `error[E0463]: can't find crate for std`. `rust-toolchain.toml` pins the repo to 1.91, which overrides `@stable` — so `cargo build --target wasm32` runs under 1.91, but the action's `targets:` input had installed wasm32 std onto stable instead. | Dropped the `with: targets:` input; added a `rustup target add wasm32-unknown-unknown` run step, which honors the pin and lands the target on 1.91 (mirrors `check_all.sh`). | `.github/workflows/ci.yml` |
| 2 (latent) | The module doc linked `[cognify_pipeline]`, but that module is now `#[cfg(not(target_arch = "wasm32"))]`. On a `cargo doc --target wasm32` build the link is unresolved → broken-intra-doc-link error (would surface on docs.rs/wasm). | Demoted the link to a plain code span (with a short note explaining why). Native docs keep working. | `crates/chunking/src/lib.rs` |

GraDKh:
Thanks! Looks really good now!

rahulm820:
Thank you, @GraDKh, for this opportunity! It was so much fun, and I learned many new things throughout the process. I'd be happy to work on more issues like these in the future.

Issue #17, opened 2026-06-26
LLM adapters: close parity gap with Python SDK (Anthropic, Azure, Bedrock + compatible-endpoint routing)


GraDKh:
## Summary

The Rust SDK currently ships **one** LLM adapter — `OpenAIAdapter` ([`crates/llm/src/adapters/openai.rs`](https://github.com/topoteretes/cognee-rs/blob/main/crates/llm/src/adapters/openai.rs)). The [`LlmProvider`](https://github.com/topoteretes/cognee-rs/blob/main/crates/llm/src/config.rs) enum *declares* `Anthropic`, `Ollama`, `Gemini`, `Mistral`, `Bedrock`, `Local`, `LiteRt`, `Mock`, but the factory in `crates/lib/src/component_manager.rs` (`init_llm_adapter`) only routes `openai` (+ `mock`); every other value returns *"Unsupported llm_provider"*. (`LiteRt`/`Local` were intentionally extracted to a closed crate and are out of scope here.)

The Python SDK exposes a much broader set via `litellm` + `instructor`: OpenAI, Anthropic, Azure OpenAI, Gemini, Mistral, Ollama, Bedrock, Llama.cpp, and a generic/custom OpenAI-compatible adapter.

The embedding side is **not** affected — `crates/embedding` already implements ONNX/fastembed, OpenAI-compatible, and Ollama engines. This issue is LLM-side only.

## Key finding: most "missing" providers don't need a new adapter

The existing `OpenAIAdapter` accepts a custom `base_url` and has a JSON-mode fallback for non-strict OpenAI-compatible servers. Several providers in the Python SDK are themselves just OpenAI-compatible endpoints, so they only need **factory routing + an alias + docs**, not a new adapter:

| Provider | New adapter needed? | Notes |
|---|---|---|
| **Ollama** | No | OpenAI-compatible endpoint; route `ollama` → `OpenAIAdapter` with endpoint. |
| **Gemini** | Mostly no | Google ships an OpenAI-compat endpoint (`/v1beta/openai/`). Native API only needed for edge multimodal cases. |
| **Mistral** | No | Native OpenAI-compatible API. |
| **Generic / Custom** | Already covered | This *is* `OpenAIAdapter` + endpoint. |
| **Azure OpenAI** | Thin variant | Same wire format; different auth (`api-key` header, `api-version` query param, deployment-based URL). |
[...]

PR #29, opened 2026-06-29
refactor(llm): extract shared OpenAI-compatible adapter builder

RajdeepKushwaha5:
## Summary

Groundwork for the LLM-adapter parity work in #17 (Tier 1). Both wiring paths that build the LLM adapter - the embedded `ComponentManager` (`cognee-lib`) and the standalone HTTP server (`cognee-http-server`) - duplicated the same
construction logic: API-key validation, empty-endpoint → provider-default fallback, and structured-output / network retry wrapping.

As called out on the issue, extracting this into a single shared function is a prerequisite for adding compatible-endpoint provider routing (ollama / mistral / gemini / custom) without re-duplicating it across both sites.

This PR does **only** that extraction. It is a pure refactor with no behavioural change.

## Changes

- Add `cognee_llm::build_openai_compatible_adapter(provider, model, api_key, endpoint, max_retries)`   in a new `crates/llm/src/factory.rs`, exported from the crate root.
- Route both `ComponentManager::init_llm_adapter` and `init_transcriber` (cognee-lib)   through it.
- Route both `wire_llm` and `wire_transcriber` (cognee-http-server) through it.
- Supports the `openai` / OpenAI-compatible path only - the same set both call sites   already handled. Additional providers are layered on top in the Tier 1 follow-up.

## Behaviour

No change. The OpenAI path is equivalent (same key check, same empty-endpoint default, same retry counts). The HTTP server still wires `None` for non-`openai` providers via its existing guard.

## Testing

- `cargo test -p cognee-llm --lib factory`  - 4 new unit tests: builds adapter, strips litellm-style `openai/` model prefix, case-insensitive provider match, missing-key error, unknown-provider error. All pass.
- `cargo fmt --all -- --check`  -  clean.
- `cargo check --all-targets` and `cargo clippy --all-targets -- -D warnings`  - clean.
- `cargo check -p cognee-lib --no-default-features` — clean.

## Scope

Part of #17 ,  **does not close it.** This is the shared-function extraction only. Tier 1 provider routing, Tier 2 (native Anthropic), and Tier 3 (Azure / Bedrock) follow in separate PRs.

### DIFF
--- a/crates/http-server/src/wiring.rs
+++ b/crates/http-server/src/wiring.rs
@@ -12,7 +12,9 @@ use cognee_database::{
 use cognee_delete::DeleteService;
 use cognee_embedding::{EmbeddingConfig, EmbeddingEngine, EmbeddingProvider};
 use cognee_graph::{GraphDBTrait, LadybugAdapter};
-use cognee_llm::{Llm, OpenAIAdapter, OpenAIResponsesClient, ResponsesClient, Transcriber};
+use cognee_llm::{
+    Llm, OpenAIResponsesClient, ResponsesClient, Transcriber, build_openai_compatible_adapter,
+};
 use cognee_ontology::{OntologyManager, OntologyResolver};
 use cognee_search::{
     SeaOrmSessionStore, SearchBuilder, SearchOrchestrator, SessionManager, SessionStore,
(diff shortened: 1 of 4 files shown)


GraDKh:
Thanks for your effort! This one looks good!

PR #30, opened 2026-06-29
feat(llm): route ollama, mistral, gemini, and custom OpenAI-compatible providers

RajdeepKushwaha5:
## Summary

Tier 1 of #17  - close the LLM-adapter parity gap with the Python SDK for the providers that are themselves OpenAI-compatible HTTP endpoints. Builds on #29 (the shared adapter factory).

Routes `ollama`, `mistral`, `gemini`, and `custom` / `openai_compatible` onto the existing `OpenAIAdapter` via the shared factory - no new adapter. The OpenAI-only request quirks stay gated on the `api.openai.com` host, so they never fire against
another endpoint.

## Routing

| Provider | API key | `LLM_ENDPOINT` default | Model prefix stripped |
|---|---|---|---|
| `ollama` | optional (placeholder used) | `http://localhost:11434/v1` | `ollama/` |
| `mistral` | required | `https://api.mistral.ai/v1` | `mistral/` |
| `gemini` | required | `https://generativelanguage.googleapis.com/v1beta/openai/` | `gemini/` |
| `custom` / `openai_compatible` | optional | **required (no default)** | none |

`LLM_ENDPOINT` always overrides the default. litellm-style model prefixes are stripped in the factory (the adapter already strips `openai/`).

## Wiring

- `ComponentManager` (`init_llm_adapter`) and the HTTP server (`wire_llm`) route the   new providers for completion.
- Transcription (`init_transcriber` / `wire_transcriber`) stays on `openai` and  `custom`/`openai_compatible` (which may expose `/audio/transcriptions`); `ollama`/`mistral`/`gemini` get graceful no-audio (`None`) rather than an adapter that 404s at runtime.

## Tests / docs

- 9 factory unit tests: per-provider endpoint defaults, prefix stripping, required-key / required-endpoint errors, unknown-provider error.
- Supported-provider table in `docs/configuration.md`; `.env.example` Ollama block switched to the native `ollama` provider.
- Verified: `fmt`, `clippy --all-targets -D warnings`, `cognee-lib --no-default-features`.

## Scope

Part of #17 - does not close it. Native Anthropic (Tier 2) and Azure/Bedrock (Tier 3) follow in separate PRs.

### DIFF
--- a/crates/lib/src/component_manager.rs
+++ b/crates/lib/src/component_manager.rs
@@ -594,11 +594,12 @@ impl ComponentManager {
             }
         }
 
-        // Build the real adapter exactly as before.
+        // Build the real adapter. OpenAI and the OpenAI-compatible providers all
+        // route through the shared factory (issue #17, Tier 1).
         let adapter: Arc<dyn Llm> = match provider.as_str() {
-            "openai" => {
+            "openai" | "ollama" | "mistral" | "gemini" | "custom" | "openai_compatible" => {
                 let adapter = build_openai_compatible_adapter(
-                    "openai",
+                    &provider,
                     &llm_model,
                     &llm_api_key,
                     &llm_endpoint,
@@ -617,7 +618,8 @@ impl ComponentManager {
             }
             _ => {
                 return Err(ComponentError::Config(format!(
-                    "Unsupported llm_provider '{provider}'. Supported: openai, mock.",
+                    "Unsupported llm_provider '{provider}'. \
+                     Supported: openai, ollama, mistral, gemini, custom, mock.",
                 )));
             }
         };
@@ -659,9 +661,14 @@ impl ComponentManager {
         };
 
         match provider.as_str() {
-            "openai" => {
+            // Whisper-style transcription works against OpenAI and any user-pointed
+            // OpenAI-compatible server that exposes /audio/transcriptions
+            // (Groq, vLLM, a LiteLLM proxy). Ollama/Mistral/Gemini do not expose
+            // that route via the chat path, so they return None (graceful no-audio)
+            // rather than building an adapter that 404s at runtime.
+            "openai" | "custom" | "openai_compatible" => {
                 let adapter = build_openai_compatible_adapter(
-                    "openai",
+                    &provider,
                     &llm_model,
                     &llm_api_key,
                     &llm_endpoint,
@@ -671,7 +678,7 @@ impl ComponentManager {
 
                 Ok(Some(Arc::new(adapter) as Arc<dyn Transcriber>))
             }
-            // litert and any future providers that do not implement Transcriber
+            // litert and providers that do not expose OpenAI-compatible audio
             // return None — audio stays gracefully unsupported (D5).
             _ => Ok(None),
         }
--- a/crates/llm/src/factory.rs
+++ b/crates/llm/src/factory.rs
@@ -24,6 +24,17 @@ const MISTRAL_DEFAULT_ENDPOINT: &str = "https://api.mistral.ai/v1";
 /// Default Gemini OpenAI-compatible base URL.
 const GEMINI_DEFAULT_ENDPOINT: &str = "https://generativelanguage.googleapis.com/v1beta/openai/";
 
+/// Providers this factory routes onto [`OpenAIAdapter`]. Single source of truth so
+/// the "unsupported provider" message can never drift from the match arms below.
+const SUPPORTED_PROVIDERS: &[&str] = &[
+    "openai",
+    "ollama",
+    "mistral",
+    "gemini",
+    "custom",
+    "openai_compatible",
+];
+
 /// Build an [`OpenAIAdapter`] for an OpenAI-compatible `provider`.
 ///
 /// Supported providers (case-insensitive): `openai`, `ollama`, `mistral`,
@@ -55,57 +66,45 @@ pub fn build_openai_compatible_adapter(
     let provider = provider.to_ascii_lowercase();
     let endpoint = endpoint.trim();
 
-    // Per provider: resolved base URL, whether an API key is mandatory, and the
-    // litellm-style model prefix to strip (if any).
-    let (base_url, api_key_required, strip_prefix): (Option<String>, bool, Option<&str>) =
-        match provider.as_str() {
-            // Empty endpoint → None so the adapter uses its OpenAI default.
-            "openai" => (non_empty(endpoint), true, None),
-            "ollama" => (
-                Some(endpoint_or(endpoint, OLLAMA_DEFAULT_ENDPOINT)),
-                false,
-                Some("ollama/"),
-            ),
-            "mistral" => (
-                Some(endpoint_or(endpoint, MISTRAL_DEFAULT_ENDPOINT)),
-                true,
-                Some("mistral/"),
-            ),
-            "gemini" => (
-                Some(endpoint_or(endpoint, GEMINI_DEFAULT_ENDPOINT)),
-                true,
-                Some("gemini/"),
-            ),
-            "custom" | "openai_compatible" => {
-                if endpoint.is_empty() {
-                    return Err(LlmError::ConfigError(format!(
-                        "llm_endpoint must be configured for provider '{provider}'"
-                    )));
-                }
-                (Some(endpoint.to_string()), false, None)
-            }
-            other => {
+    // Per provider: resolved base URL and the litellm-style model prefix to strip.
+    let (base_url, strip_prefix): (Option<String>, Option<&str>) = match provider.as_str() {
+        // Empty endpoint → None so the adapter uses its OpenAI default.
+        "openai" => (non_empty(endpoint), None),
+        "ollama" => (
+            Some(endpoint_or(endpoint, OLLAMA_DEFAULT_ENDPOINT)),
+            Some("ollama/"),
+        ),
+        "mistral" => (
+            Some(endpoint_or(endpoint, MISTRAL_DEFAULT_ENDPOINT)),
+            Some("mistral/"),
+        ),
+        "gemini" => (
+            Some(endpoint_or(endpoint, GEMINI_DEFAULT_ENDPOINT)),
+            Some("gemini/"),
+        ),
+        "custom" | "openai_compatible" => {
+            if endpoint.is_empty() {
                 return Err(LlmError::ConfigError(format!(
-                    "Unsupported llm_provider '{other}'. \
-                     Supported: openai, ollama, mistral, gemini, custom."
+                    "llm_endpoint must be configured for provider '{provider}'"
                 )));
             }
-        };
+            (Some(endpoint.to_string()), None)
+        }
+        other => {
+            return Err(LlmError::ConfigError(format!(
+                "Unsupported llm_provider '{other}'. Supported: {}.",
+                SUPPORTED_PROVIDERS.join(", ")
+            )));
+        }
+    };
 
-    if api_key_required && api_key.is_empty() {
+    // Every supported provider requires an API key, matching the Python SDK's
+    // _API_KEY_REQUIRED_PROVIDERS (openai, ollama, mistral, gemini, custom).
+    if api_key.is_empty() {
         return Err(LlmError::ConfigError(
             "llm_api_key must be configured".to_string(),
         ));
     }
-
-    // Ollama ignores auth but the OpenAI-style client still sends a bearer token;
-    // use a harmless placeholder when none is configured (matches Python cognee's
-    // `LLM_API_KEY="ollama"` convention).
-    let api_key = if provider == "ollama" && api_key.is_empty() {
-        "ollama"
-    } else {
-        api_key
-    };
 
     let model = match strip_prefix {
         Some(prefix) => model.strip_prefix(prefix).unwrap_or(model),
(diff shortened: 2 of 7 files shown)


GraDKh:
@RajdeepKushwaha5 Could you please rebase it on the latest main

crates/llm/src/factory.rs
```diff
@@ -0,0 +1,239 @@
...
+//! trigger any OpenAI-specific behaviour.
+
+use crate::{LlmError, LlmResult, OpenAIAdapter};
+
+/// Default OpenAI-compatible base URL for a local Ollama server.
+const OLLAMA_DEFAULT_ENDPOINT: &str = "http://localhost:11434/v1";
+/// Default Mistral OpenAI-compatible base URL.
+const MISTRAL_DEFAULT_ENDPOINT: &str = "https://api.mistral.ai/v1";
+/// Default Gemini OpenAI-compatible base URL.
+const GEMINI_DEFAULT_ENDPOINT: &str = "https://generativelanguage.googleapis.com/v1beta/openai/";
```
GraDKh, line 25: **Gemini's default endpoint will 404 — the trailing slash produces a double-slash path.**

`GEMINI_DEFAULT_ENDPOINT` ends in `/`, and the adapter builds request URLs as `format!("{}/chat/completions", self.base_url)` (`adapters/openai.rs:196`, plus `/audio/transcriptions` at `:862`) without normalizing the trailing slash. So the default `LLM_PROVIDER=gemini` config (no `LLM_ENDPOINT`) hits `…/v1beta/openai//chat/completions` — the `url` crate keeps the empty segment, and Google's gateway rejects it. The other defaults (openai/ollama/mistral) have no trailing slash, so only gemini is affected, and no unit test catches it (the tests only assert `.model()`, never the resolved `base_url`).

Suggest normalizing once in `OpenAIAdapter::new` rather than special-casing the constant:

```rust
base_url: base_url
    .map(|u| u.trim_end_matches('/').to_string())
    .unwrap_or_else(|| Self::DEFAULT_BASE_URL.to_string()),
```

That fixes the gemini default and also protects any user-supplied `custom`/`gemini` `LLM_ENDPOINT` that happens to end in `/`.

GraDKh:
**Tier 1 acceptance: missing the gracefully-skipped integration test.**

Issue #17 Tier 1 lists "unit + a gracefully-skipped integration test mirroring `crates/llm/tests/integration_openai.rs`." This PR adds the 9 factory unit tests but no integration test for the new providers — nothing under `crates/llm/tests/` references ollama/mistral/gemini/custom. The unit tests assert only `.model()`, so they never build a request URL or touch an endpoint.

That's also why two gemini issues went unnoticed: the double-slash endpoint (see inline comment) and, more subtly, structured output — the adapter requests strict `response_format: {json_schema, strict:true}` with a closed-schema rewrite and falls back to `json_object`, whereas Python uses the native Gemini transport with instructor `json_mode`. Google's `/v1beta/openai/` shim doesn't reliably honor strict json_schema, so gemini cognify may silently run through the lenient fallback or fail.

A gracefully-skipped integration test (skip when the provider's key/endpoint env vars are absent, like `integration_openai.rs`) that runs one real completion per configured provider would cover both — and would let gemini be marked "validated" rather than "wired."

GraDKh:
**Settings dropdown (`routers/settings.rs:36`) is now out of sync with the factory.**

This PR makes the factory accept `custom`/`openai_compatible`, but `llm_providers()` doesn't list them, so they're not selectable in the UI. Conversely the dropdown still offers `anthropic` and `bedrock`, which the factory rejects with `Unsupported llm_provider` (pre-existing, not introduced here). Worth reconciling the list to what the factory actually supports — `openai, ollama, mistral, gemini, custom` — in this PR or a quick follow-up.

GraDKh:
**Minor / parity with Python cognee:** `init_embedding_engine` falls the embedding endpoint back to `llm_endpoint` (`component_manager.rs:461`). The Python cognee SDK — which this project tracks as its behavioral reference — only shares the *api_key* across LLM ↔ embedding, never the endpoint: in `cognee/infrastructure/databases/vector/embeddings/get_embedding_engine.py`, `embedding_endpoint` is sourced solely from `EMBEDDING_ENDPOINT`, and `OllamaEmbeddingEngine` receives it with no LLM fallback (only `OpenAICompatible`/`LiteLLM` engines do `embedding_api_key or llm_api_key`, key-only).

This PR makes `LLM_PROVIDER=ollama` reachable with the OpenAI-compat default `…:11434/v1`, so the combo `LLM_ENDPOINT=…/v1` + `EMBEDDING_PROVIDER=ollama` (no `EMBEDDING_ENDPOINT`) now leaks `/v1` into the ollama embedder, which needs `/api/embed`. The endpoint fallback itself is pre-existing and out of scope here, but it's a behavioral divergence from the Python reference — worth a docs caveat, and possibly revisiting the fallback to match Python's key-only behavior.

crates/llm/src/factory.rs
```diff
@@ -0,0 +1,239 @@
...
+    if api_key_required && api_key.is_empty() {
+        return Err(LlmError::ConfigError(
+            "llm_api_key must be configured".to_string(),
+        ));
+    }
+
+    // Ollama ignores auth but the OpenAI-style client still sends a bearer token;
+    // use a harmless placeholder when none is configured (matches Python cognee's
+    // `LLM_API_KEY="ollama"` convention).
+    let api_key = if provider == "ollama" && api_key.is_empty() {
```
GraDKh: **This should match the Python cognee behavior — please require the API key for `ollama` and `custom`.**

Python's `_API_KEY_REQUIRED_PROVIDERS` (`get_llm_client.py:95`) lists `OPENAI, OLLAMA, CUSTOM, GEMINI, MISTRAL, ANTHROPIC`, so Python *requires* `LLM_API_KEY` for both `ollama` and `custom` and raises `LLMAPIKeyNotSetError` when it's missing — there is no placeholder auto-injection. Since the Python SDK is this project's behavioral reference, the priority here is parity: `ollama` and `custom`/`openai_compatible` should require the key (add them to the `api_key_required` arms) rather than defaulting it.

Concretely:
- Drop the `"ollama"` placeholder block and set `api_key_required = true` for `ollama`.
- Set `api_key_required = true` for `custom` / `openai_compatible`.

Also note the comment's claim that the placeholder "matches Python cognee's `LLM_API_KEY="ollama"` convention" is inaccurate — Python has no such auto-injection — so it should be removed along with the placeholder.

GraDKh:
**Parity follow-up (likely out of scope for this PR):** Python's `LLMConfig.ensure_env_vars_for_ollama` (`config.py:154`) rejects a partial ollama config — when `LLM_PROVIDER=ollama`, it requires `LLM_MODEL`/`LLM_ENDPOINT`/`LLM_API_KEY` to be set together, and likewise the `EMBEDDING_PROVIDER`/`EMBEDDING_MODEL`/`EMBEDDING_DIMENSIONS` triplet. The Rust port has no equivalent validation, so a partial ollama config that Python errors on will silently fall back to defaults here. Since the Python cognee SDK is this project's behavioral reference, worth porting this validator (here or as a follow-up) now that `ollama` is a supported provider.

crates/lib/src/component_manager.rs
```diff
@@ -669,29 +661,24 @@ impl ComponentManager {
...
-                let adapter = OpenAIAdapter::new(llm_model, llm_api_key, endpoint)
-                    .map_err(|e| ComponentError::Llm(format!("initialization failed: {e}")))?
-                    .with_structured_output_retries(retries)
-                    .with_network_retries(retries);
+            // Whisper-style transcription works against OpenAI and any user-pointed
+            // OpenAI-compatible server that exposes /audio/transcriptions
+            // (Groq, vLLM, a LiteLLM proxy). Ollama/Mistral/Gemini do not expose
+            // that route via the chat path, so they return None (graceful no-audio)
+            // rather than building an adapter that 404s at runtime.
+            "openai" | "custom" | "openai_compatible" => {
```
GraDKh, line 669: **Parity note (acceptable for Tier 1, track for later):** Python implements `create_transcript` for all three deferred providers — Mistral via its native SDK (`mistral/adapter.py:165`), Ollama via `whisper-1` on its OpenAI-compat client, and Gemini via `GenericAPIAdapter`. Returning `None` here is a deliberate, safer simplification (avoids a runtime 404), but it's a divergence from the Python cognee reference. Fine to keep for Tier 1; worth tracking as a parity gap for when native Mistral/Gemini adapters land.

crates/llm/src/factory.rs
```diff
@@ -0,0 +1,239 @@
...
+                if endpoint.is_empty() {
+                    return Err(LlmError::ConfigError(format!(
+                        "llm_endpoint must be configured for provider '{provider}'"
+                    )));
+                }
+                (Some(endpoint.to_string()), false, None)
+            }
+            other => {
+                return Err(LlmError::ConfigError(format!(
+                    "Unsupported llm_provider '{other}'. \
```
GraDKh: **Derive the "Supported: …" list from a single source instead of hardcoding it.** The current message omits `openai_compatible` (an accepted alias) and `mock` (handled upstream in the component manager), and any future provider added to the match arms will silently drift out of this string again.

Suggest hoisting the accepted provider names into one constant and formatting the error from it, so the list can't go stale:

```rust
const SUPPORTED_PROVIDERS: &[&str] =
    &["openai", "ollama", "mistral", "gemini", "custom", "openai_compatible"];
// ...
other => Err(LlmError::ConfigError(format!(
    "Unsupported llm_provider '{other}'. Supported: {}.",
    SUPPORTED_PROVIDERS.join(", ")
))),
```

(`mock` is resolved before this function, so either include it in the const or note in the message that it's handled separately — whichever you prefer.)

GraDKh (changes requested):
Thanks for this — the shared-factory approach is clean, the per-provider prefix stripping is correct, and routing the OpenAI-compatible providers through one adapter is exactly the right call for Tier 1. Really nice, well-documented work, and the `api.openai.com` gating of the OpenAI-only quirks is a thoughtful touch.

I've left inline and general comments. Requesting changes mainly on one runtime issue plus a couple of parity items:

- **Blocking:** the `gemini` default endpoint's trailing slash produces a double-slash request path (`…/openai//chat/completions`) that will 404 — see the inline note for the suggested one-line normalization in `OpenAIAdapter::new`.
- **Parity (the Python cognee SDK is our behavioral reference):** please require the API key for `ollama` and `custom` to match Python's `_API_KEY_REQUIRED_PROVIDERS`, and correct the placeholder rationale comment.
- **Tier 1 acceptance:** add the gracefully-skipped integration test mirroring `integration_openai.rs` — that's also what would have caught the gemini issues.
- Smaller follow-ups: reconcile the settings-UI provider list with the factory, and a few tracked parity notes (embedding endpoint fallback, ollama env validation, transcription) that can land here or as follow-ups.

Thanks again — happy to re-review as soon as the gemini fix and key-requirement parity are in.

RajdeepKushwaha5:
Thanks for the thorough review. Rebased onto latest `main` and pushed; addressed the blocking + parity items:

Gemini double-slash (blocking): normalized the trailing slash once in `OpenAIAdapter::new` (`trim_end_matches('/')`), so the `.../v1beta/openai/` default and any user endpoint ending in `/` no longer build `//chat/completions`. Added a unit test asserting the resolved `base_url`.

API-key parity: `ollama` and `custom`/`openai_compatible` now require `LLM_API_KEY`, matching Python's `_API_KEY_REQUIRED_PROVIDERS`. Dropped the placeholder block and the inaccurate "matches Python" comment; added `ollama_requires_api_key` / `custom_requires_api_key` tests; updated `docs/configuration.md` and `.env.example`.

Gracefully-skipped integration test: added `crates/llm/tests/integration_providers.rs` mirroring `integration_openai.rs` - one real completion per provider (ollama/mistral/gemini/custom), early-return skip when that provider's env vars are absent, so CI stays green without secrets. This is the harness that would surface the gemini strict-`json_schema` behavior against Google's shim once a key is configured.

Supported-provider list: hoisted into a `SUPPORTED_PROVIDERS` const so the error message can't drift; it now includes the `openai_compatible` alias.

Embedding-endpoint caveat: documented that `EMBEDDING_PROVIDER=ollama` needs an explicit `EMBEDDING_ENDPOINT` (`/api/embed`), since it does not inherit LLM_ENDPOINT.

Tracked as follow-ups (you flagged these as out of scope here): reconciling the settings-UI provider list with the factory (it intersects the documented Python settings-DTO parity contract, so I'd rather do it as its own change), porting `ensure_env_vars_for_ollama`, and native Mistral/Gemini transcription parity.

Happy to fold the settings-UI list into this PR instead if you'd prefer it here rather than a follow-up.

RajdeepKushwaha5:
The `Test (stable)` failure doesn't reproduce locally. On this branch I've run the full set that doesn't need live credentials and it's green: `cargo test -p cognee-llm` (78 lib + the offline integration tests), `cognee-lib`, `cognee-http-server`, the new factory/provider unit tests, doctests, and `clippy --all-targets -D warnings + fmt`.

The failing step is the live-LLM `Test` lane (`cargo nextest run --workspace` against the real `OPENAI_KEY`). This change doesn't alter the request sent to `api.openai.com/v1` - the `openai` factory arm is unchanged from what's already on `main`, and the only `OpenAIAdapter` tweak (trailing-slash trim) is a no-op for an endpoint without a trailing slash. The new provider/anthropic integration tests gracefully skip when their env vars are absent, so they don't run in CI.

Could you re-run the `Test (stable)` job? It fails fast (~1–2 min) across all three stacked PRs, which looks like a transient/quota issue on the shared key rather than a logic failure. If it's a real test, please point me at the failing test name from the log and I'll fix it right away.

(Separately, `http-parity` is also red on `main` right now, so it doesn't look introduced by these PRs.)

GraDKh:
Thanks @RajdeepKushwaha5 — really clean work. The shared-factory routing is exactly the right call for Tier 1, and you addressed all the blocking and parity feedback with backing tests: the gemini trailing-slash fix, API-key parity with Python's `_API_KEY_REQUIRED_PROVIDERS`, the gracefully-skipped `integration_providers` harness, and the `SUPPORTED_PROVIDERS` const.

Verified locally on the branch: `cognee-llm`, `cognee-lib`, and `cognee-http-server` are all green for everything this PR touches. (The one `post_cognify_blocking` failure is a pre-existing embedding-path 401 from a missing key — it reproduces identically on `main` and is unrelated to this diff.)

Two follow-ups to track separately, neither blocking: (1) gemini structured output is wired but only skip-validated — worth a live run against Google's `/v1beta/openai/` shim or documenting gemini as experimental, since strict `json_schema` may fall back to lenient mode; (2) minor parity nit — the API-key check uses `is_empty()` vs Python's `.strip() == ""`. Plus the already-noted settings-UI dropdown reconciliation and `ensure_env_vars_for_ollama`.

Merging. Thanks again!

PR #31, opened 2026-06-29
ci(test): run workspace tests in parallel under nextest

GraDKh:
## What

Run the workspace test suite **in parallel** under `cargo nextest` instead of serially.

The test job ran `cargo nextest run --no-capture` (which forces `--test-threads=1`), spending ~20 min on the run phase alone. That serial constraint was a holdover from the `cargo test` era, where a shared process + threads let env-var mutation and `once_cell` singletons race. nextest runs each test in its own process, so that isolation is intrinsic and the global serialization is redundant.

## Validation

Validated locally with real credentials: the full suite (**2312 tests**) passes under parallel execution with no races, leaks, port conflicts, or timeouts (1 flaky test passed on retry).

This PR's own CI run is the measurement of the wall-clock effect on a real 4-core runner.

## Changes

- `.config/nextest.toml`: `default` + `ci` profiles (parallel, `fail-fast` off; `ci` adds `retries=1` for transient external flakiness). Concurrency left at the nextest default (= CPU cores).
- `scripts/lib/common.sh`: drop `--no-capture` from the nextest invocation.
- `ci.yml`: test job exports `NEXTEST_PROFILE=ci` (read natively by nextest).
- `crates/cli/src/config_store.rs`: honor `XDG_CONFIG_HOME` on all platforms. `dirs::config_dir()` only consults it on Linux; on macOS it returned `~/Library/Application Support`, so parallel CLI integration tests (each setting its own `XDG_CONFIG_HOME` temp dir) collided on one shared config file and raced on the atomic `config.json[.tmp]` replace. No-op on Linux (where CI runs); restores cross-platform per-test isolation.
- `docs/build/ci-test-parallelism.md`: rationale + isolation gotcha.

### DIFF
--- a/.github/workflows/record-cassettes.yml
+++ b/.github/workflows/record-cassettes.yml
@@ -1,0 +1,119 @@
+name: Record LLM cassettes
+
+# Manually-triggered job that re-records the committed LLM cassettes used by the
+# cassette-replay integration tests (Approach E — see
+# docs/build/ci-test-parallelism.md). The normal CI test job runs these tests
+# offline from the committed cassettes (COGNEE_TEST_REPLAY=1); when a prompt,
+# schema, or model changes, the cassettes drift and replay fails with
+# MissPolicy::Error. Run this workflow to regenerate them against the live API,
+# then it commits the refreshed cassettes back to the branch.
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Branch to record on and commit the refreshed cassettes to"
+        required: true
+        default: main
+
+concurrency:
+  group: record-cassettes-${{ github.event.inputs.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_PROFILE_DEV_DEBUG: "0"
+  CARGO_INCREMENTAL: "0"
+  ORT_CACHE_DIR: ${{ github.workspace }}/target/ort-cache
+  CCACHE_COMPILERCHECK: content
+  # Record mode: wrap the real adapter and write/merge each cassette on drop.
+  COGNEE_RECORD_LLM: "1"
+  OPENAI_URL: https://api.openai.com/v1
+  OPENAI_TOKEN: ${{ secrets.OPENAI_KEY }}
+  OPENAI_MODEL: gpt-4o-mini
+
+jobs:
+  record:
+    name: Record cassettes
+    runs-on: ubuntu-latest
+    # contents:write lets the default GITHUB_TOKEN commit the refreshed cassettes
+    # (they are not workflow files, so no `workflow` scope is needed).
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.ref }}
+
+      - name: Install Rust (stable)
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install mold linker
+        uses: rui314/setup-mold@v1
+        with:
+          make-default: true
+
+      - name: Install build deps
+        run: sudo apt-get update && sudo apt-get install -y cmake protobuf-compiler
+
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: true
+          swap-storage: true
+          docker-images: true
+
+      - name: ccache (lbug bundled C++, restore-only)
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: lbug-cxx
+          max-size: 1G
+          save: false
+
+      - name: Cache cargo/target (test)
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: workspace-test-v6
+
+      - name: Cache ORT binary
+        uses: actions/cache@v4
+        with:
+          path: target/ort-cache
+          key: ort-v2.0.0-rc.12-cpu-linux-x86_64
+
+      # Record each cassette-backed test against the live API. --test-threads=1
+      # so tests in the same file merge into one cassette without racing the
+      # final atomic write. The deterministic embedding engine is chosen in-code
+      # (no MOCK_EMBEDDING env needed), so embeddings match replay exactly.
+      - name: Record cognify cassettes
+        run: |
+          cargo test -p cognee-cognify \
+            --test integration_fact_extraction \
+            --test integration_summarization \
+            --test e2e_triplet_vector_cleanup \
+            --test e2e_delete_preview_accuracy \
+            --test e2e_shared_entity_graph_delete \
+            --test e2e_lifecycle_loop \
+            -- --test-threads=1
+
+      - name: Record search cassettes
+        run: cargo test -p cognee-search --test last_accessed_update -- --test-threads=1
+
+      - name: Record delete cassettes
+        run: cargo test -p cognee-delete --test hard_mode_orphan_sweep -- --test-threads=1
+
+      - name: Commit refreshed cassettes
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          if git diff --quiet -- '**/tests/fixtures/cassettes/*.json'; then
+            echo "Cassettes unchanged — nothing to commit."
+            exit 0
+          fi
+          git add '**/tests/fixtures/cassettes/*.json'
+          git commit -m "chore(test): re-record LLM cassettes [skip ci]"
+          git push origin HEAD:${{ github.event.inputs.ref }}
--- a/.github/workflows/record-cassettes.yml
+++ b/.github/workflows/record-cassettes.yml
@@ -121,10 +121,17 @@ jobs:
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          if git diff --quiet -- '**/tests/fixtures/cassettes/*.json'; then
+          # Stage first, then gate on the *staged* diff. `git diff --quiet`
+          # (working tree vs HEAD) ignores untracked files, so the very first
+          # recording of a NEW cassette — an untracked <name>.json — would read
+          # as "no changes" and be silently dropped, leaving the next CI run to
+          # fail replay with MissPolicy::Error. `git add` picks up new + modified
+          # files; `git diff --cached --quiet` then reflects exactly what would
+          # be committed.
+          git add '**/tests/fixtures/cassettes/*.json'
+          if git diff --cached --quiet; then
             echo "Cassettes unchanged — nothing to commit."
             exit 0
           fi
-          git add '**/tests/fixtures/cassettes/*.json'
           git commit -m "chore(test): re-record LLM cassettes [skip ci]"
           git push origin "HEAD:$REF"
--- a/.github/workflows/record-cassettes.yml
+++ b/.github/workflows/record-cassettes.yml
@@ -74,10 +74,15 @@ jobs:
           max-size: 1G
           save: false
 
-      - name: Cache cargo/target (test)
+      # Restore-only: this manual job builds with the `mock` feature (a divergent
+      # target dir) and must NOT write back to workspace-test-v6 — that key has a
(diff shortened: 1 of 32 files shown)


crates/cognify/tests/e2e_triplet_vector_cleanup.rs
```diff
@@ -81,14 +75,7 @@ async fn test_triplet_vector_cleanup_after_data_delete() {
...
 
-    let llm: Arc<dyn Llm> = Arc::new(
-        OpenAIAdapter::new(
-            require_env("OPENAI_MODEL"),
-            require_env("OPENAI_TOKEN"),
-            Some(require_env("OPENAI_URL")),
-        )
-        .expect("OpenAIAdapter::new"),
-    );
+    let llm: Arc<dyn Llm> = create_llm_from_env("triplet_vector_cleanup");
```
GraDKh, line 78: **Finding #1 (highest priority) — pre-existing `Err => return` skip blocks silently swallow `MissPolicy::Error` cassette misses, defeating the replay-fail-loudly design.**

This test (and its siblings) wrap `cognify(...)` in `match { Ok(r) => r, Err(e) => { eprintln!("Skipping…"); return; } }`. In replay mode a missing/stale cassette entry makes `ReplayLlm` return `LlmError::InvalidResponse` (`replay.rs:89`), which `cognify` propagates as `Err`, so the test prints to stderr and **returns green with zero assertions**. This directly negates the guarantee the PR's own `test_utils` doc states ("fails loudly instead of silently returning an empty graph").

*Failure scenario:* a developer legitimately edits a cognify prompt template → input hash changes → every cassette entry misses → CI stays green while testing nothing, so a real cognify regression ships undetected.

Same pattern at:
- `e2e_triplet_vector_cleanup.rs:151,181` (skip block)
- `e2e_delete_preview_accuracy.rs:137`
- `e2e_lifecycle_loop.rs:193`
- `e2e_shared_entity_graph_delete.rs:161,191`
- `hard_mode_orphan_sweep.rs:238,417`

**Fix:** in replay mode (`COGNEE_TEST_REPLAY` set), `panic!`/`.expect()` on the cognify error instead of `return`, so a cassette miss actually fails CI.

GraDKh (commented):
Automated review (medium effort). The parallelization and CI restructuring are well-reasoned; the inline comments flag issues in the replay test design and the `config_store.rs` change. Ranked actionable items: **#1** silent test neutering (highest), **#2/#3** config relocation + relative-path regression, **#7** un-gated cache writer, **#4** dispatch injection hardening, with **#5/#6** as latent/maintenance items.

Two findings reference lines outside the diff hunks and are anchored to the nearest changed line in the same file (the real location is named in the comment).

crates/cli/src/config_store.rs
```diff
@@ -28,9 +28,20 @@ impl Default for ConfigDocument {
...
+    // Honor `XDG_CONFIG_HOME` on every platform when it is set. `dirs::config_dir()`
+    // only consults it on Linux — on macOS it unconditionally returns
+    // `~/Library/Application Support`, ignoring the override. That platform gap let
+    // concurrent CLI invocations (e.g. parallel integration tests, each setting its
+    // own `XDG_CONFIG_HOME` to an isolated temp dir) collide on the single shared
+    // macOS config file and race on the atomic `config.json[.tmp]` replace. Checking
+    // the variable first makes config isolation work identically across platforms;
+    // it is a no-op on Linux where `dirs` already resolves it.
+    let base_dir = match std::env::var_os("XDG_CONFIG_HOME") {
+        Some(xdg) if !xdg.is_empty() => PathBuf::from(xdg),
```
GraDKh: **Finding #2 — honoring `XDG_CONFIG_HOME` on macOS silently relocates durable user config, orphaning existing secrets.**

`config.json` holds persistent credentials (`llm_api_key`, `db_password`, `graph_database_password`, …) with no regeneration path, and `load_config()` returns `ConfigDocument::default()` when the file is absent (lines 63–64). On macOS a dev who has `XDG_CONFIG_HOME` set (common — nvim/git/etc.) and previously ran `cognee-cli config set llm_api_key …` (written to `~/Library/Application Support/cognee-rust/`) will, after this build, read `$XDG_CONFIG_HOME/cognee-rust/config.json`, find nothing, and silently run with empty credentials. No migration, no warning. Consider gating on a test-only var, or falling back to the legacy macOS path when the XDG one is absent.

**Finding #3 — a relative `XDG_CONFIG_HOME` is now accepted, contradicting the comment's "no-op on Linux" claim and the XDG spec.**

The guard is only `!xdg.is_empty()`. `dirs::config_dir()` (the prior Linux path) ignores a *relative* `XDG_CONFIG_HOME` per spec; this branch uses it verbatim. With e.g. `XDG_CONFIG_HOME=.config`, config resolves relative to the current working directory, so `config set` from `~/projA` is invisible when run from `~/projB` — config silently becomes per-CWD. This is a behavior change on Linux, not the claimed no-op. Add a `Path::is_absolute()` check before accepting the override.

.github/workflows/record-cassettes.yml
```diff
@@ -0,0 +1,119 @@
...
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          if git diff --quiet -- '**/tests/fixtures/cassettes/*.json'; then
+            echo "Cassettes unchanged — nothing to commit."
+            exit 0
+          fi
+          git add '**/tests/fixtures/cassettes/*.json'
+          git commit -m "chore(test): re-record LLM cassettes [skip ci]"
+          git push origin HEAD:${{ github.event.inputs.ref }}
```
GraDKh: **Finding #4 — `${{ github.event.inputs.ref }}` interpolated directly into a `run:` shell line (GitHub Actions script-injection sink).**

`${{ … }}` is template-substituted into the script as raw text *before* bash parses it — the documented injection pattern, not a separate argv token. A `ref` value like `main$(…)` would execute on a runner that has `contents: write` and the `OPENAI_KEY` secret. Low severity (requires `workflow_dispatch` access), but trivially hardened: pass the input via an `env:` var and reference `"$REF"`.

.github/workflows/record-cassettes.yml
```diff
@@ -0,0 +1,119 @@
...
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: lbug-cxx
+          max-size: 1G
+          save: false
+
+      - name: Cache cargo/target (test)
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: workspace-test-v6
```
GraDKh, line 84: **Finding #7 — this record job is an un-gated writer of `workspace-test-v6`, contradicting this PR's own cache-collision fix.**

Every other lane touching this key (`docs-and-extras`, pgvector-spans) was deliberately set to `save-if: "false"` and the binding jobs were moved to their own keys, specifically to "stop multi-job cache-key collisions on workspace-test-v6" (commit eb9f5b5). This step still saves it. *Failure scenario:* a manually-dispatched record run racing a `main` CI run both push `workspace-test-v6`; the later writer wins with a target dir built against a different crate subset / the `mock`-feature runtime path, triggering a mass rebuild on the next `test` run. Low probability (dispatch-only) but the asymmetry is clearly unintended — add `save: false` or give it a dedicated key.

crates/llm/src/mock/cassette.rs
```diff
@@ -81,6 +81,17 @@ impl LlmCassette {
         let contents = serde_json::to_string_pretty(self).map_err(|e| {
             LlmError::SerializationError(format!("failed to serialize cassette: {e}"))
         })?;
+        // Create the parent directory if it does not exist yet, so recording to a
+        // fresh `tests/fixtures/cassettes/<name>.json` path succeeds on the first
+        // run instead of silently failing in `RecordingLlm`'s Drop flush.
+        if let Some(parent) = path.parent().filter(|p| !p.as_os_str().is_empty()) {
+            std::fs::create_dir_all(parent).map_err(|e| {
```
GraDKh, line 88: **Finding #5 (latent) — re: `canonicalize` at line 142, not this added block.** The replay key canonicalizes object keys but deliberately preserves *array* order (`canonicalize`, line 142; test at 253). The hashed schema includes arrays like `"required": [...]`; a schemars version bump or a field reorder in `KnowledgeGraph`/`SummarizedContent` changes that order → hash changes → all entries miss. Combined with Finding #1, the result is silent: a routine dependency bump quietly neuters the cassette tests instead of failing CI. Deterministic within a build (not a flake) — a latent maintenance trap worth a note.

crates/search/tests/last_accessed_update.rs
```diff
@@ -118,7 +68,7 @@ async fn test_search_updates_last_accessed_timestamp() {
     // In-memory mock vector DB (qdrant extracted to closed cognee-vector-qdrant).
     let vector_db: Arc<dyn VectorDB> = Arc::new(MockVectorDB::new());
 
-    let llm: Arc<dyn Llm> = create_adapter_from_env();
+    let llm: Arc<dyn Llm> = create_llm_from_env("last_accessed_update");
```
GraDKh, line 79: **Finding #6 (minor) — coverage erosion + stale header.** Switching to `MockEmbeddingEngine::deterministic` is safe here only because `MockVectorDB::search_similar` applies no similarity threshold and there is a single chunk, so retrieval always returns it. The test now verifies the timestamp-update plumbing but no longer exercises semantic retrieval at all. Also, the file header (lines 11–13) still claims it requires `COGNEE_E2E_EMBED_MODEL_PATH`/`COGNEE_E2E_TOKENIZER_PATH`, which is now misleading. Update the comment and add an explicit note that ranking is no longer tested.

crates/cognify/tests/e2e_triplet_vector_cleanup.rs
GraDKh: Fixed in 8bda728. Added \`fail_loudly_on_replay_miss(what, err)\` (cognify & search \`test_utils\`, inlined in the delete test) and call it in every skip-on-Err block. In \`COGNEE_TEST_REPLAY\` mode a cassette miss now panics ("cognify failed in replay mode — re-record cassettes") instead of returning green with zero assertions. Verified empirically: emptying a cassette's `entries` makes the test fail loudly.

crates/cli/src/config_store.rs
GraDKh: Fixed in 8bda728. Dropped \`XDG_CONFIG_HOME\` keying entirely — production always resolves via \`dirs::config_dir()\` now, unchanged on every platform (no macOS config relocation / orphaned secrets). Tests use a dedicated \`COGNEE_CONFIG_HOME\`, honored only when \`Path::is_absolute()\` (so a relative value can no longer make config CWD-dependent). CLI tests updated to set it; verified parallel \`cli_memify\` isolation with no real-config-dir pollution.

.github/workflows/record-cassettes.yml
GraDKh: Fixed in 8bda728. The \`ref\` input now flows through an \`env: REF:\` var and the push uses \`"HEAD:$REF"\` — no \`${{ }}\` interpolation into the shell line.

.github/workflows/record-cassettes.yml
GraDKh: Fixed in 8bda728. The record job's rust-cache is now \`save-if: "false"\` (restore-only), so it no longer writes \`workspace-test-v6\` — the single-writer invariant from eb9f5b5 is preserved.

crates/llm/src/mock/cassette.rs
GraDKh: Addressed in 8bda728 with a doc note at \`canonicalize()\`: JSON array order (schema \`required\` lists, property order) is part of the hash, so a schemars bump / field reorder misses all entries. With Finding #1 fixed, that now surfaces loudly (MissPolicy::Error + the skip-block guard) → re-record, rather than silently. Keying itself unchanged by design.

crates/search/tests/last_accessed_update.rs
GraDKh: Fixed in 8bda728. Header no longer claims it requires \`COGNEE_E2E_EMBED_MODEL_PATH\`/tokenizer; added a scope note that it uses deterministic mock embeddings and no longer exercises semantic retrieval/ranking (that stays in \`integration_search_matrix\`, on a real engine).

PR #34, opened 2026-06-29
Perf: Optimize embeddings generation & engines

### LINKED ISSUE #20: Perf: optimize embedding generation (eliminate double-embedding pass, batch tuning)
## Summary

Investigate and optimize **embedding generation** (remote OpenAI-compatible API, local ONNX, Ollama) during ingestion, cognify and search. The trait is already batch-aware and most ingestion call sites batch correctly — but there is a **confirmed redundant double-embedding pass** that roughly doubles embedding work per `cognify`, plus a couple of smaller wins.

LLM slice is tracked separately; DB slice separately. Complements #18.

## Current state (verified)

**Trait & engines**
- `crates/embedding/src/engine.rs:30` — `EmbeddingEngine::embed(&self, texts: &[&str]) -> Vec<Vec<f32>>`. Batch-aware by design; engines sub-batch internally and L2-normalize.
- `crates/embedding/src/openai_compatible.rs:186` — sends multiple texts per HTTP request (`input: Vec<&str>`), sub-batched by `batch_size` (default 36, `EMBEDDING_BATCH_SIZE`). Retry w/ jitter. ✅ batched, sub-batches sent **sequentially**.
- `crates/embedding/src/onnx.rs:303` — sub-batched (default 32) via `spawn_blocking`. ✅
- `crates/embedding/src/ollama.rs:180` — ⚠️ **one HTTP request per text** (`input: &str`), concurrency via `join_all`. Does not exploit batch input arrays.

**Ingestion call sites** (`crates/cognify/src/tasks.rs`) — all use single batched `embed(&[...])` calls per type: chunks, entity names, summaries, entity types, triplets. ✅

## ⚠️ Primary finding: chunks/entities/summaries are embedded twice per `cognify`

Within a single `add_data_points` run, **both** of these are called back-to-back on the same inputs:
- `generate_embeddings(...)` — `crates/cognify/src/tasks.rs:1080` → embeds chunks (`:2593`), entities (`:2610`), summaries (`:2627`) into `Vec<Embedding>`, returned as `CognifyResult.embeddings`.
- `index_data_points(...)` — `crates/cognify/src/tasks.rs:1088` → **re-embeds the same** chunks (`:2679`), entities (`:2736`), summaries (`:2855`) to push into the vector DB.

The output of `generate_embeddings` (`CognifyResult.embeddings`) is **never persisted** — it is consumed only for a `.len()` count in reporting and tests:
- `crates/cli/src/commands/cognify.rs:211`, `:220`
- `crates/cli/src/commands/add_and_cognify.rs:217`
- `crates/bindings-common/src/wire.rs:36`
- `crates/cognify/tests/integration_embeddings.rs` (assertions only)

So the first full embedding pass over chunks + entities + summaries exists essentially to produce a count, and the *real* indexing embeds everything again. On the remote API this doubles embedding requests/latency for those three types; on local ONNX it doubles CPU.

**Likely quick win:** have `index_data_points` reuse the vectors produced by `generate_embeddings` (pass them in, or fold the two functions together), so each text is embedded once. Needs care: `index_data_points` also embeds types/triplets that `generate_embeddings` does not, and the `CognifyResult.embeddings` count contract must be preserved.

## Secondary opportunities

1. **Ollama batch input** — `crates/embedding/src/ollama.rs:180`: check whether `/api/embed` accepts an array `input` (it does in recent Ollama) and batch like the OpenAI adapter, instead of one request per text.
2. **OpenAI sub-batch size & parallelism** — `crates/embedding/src/config.rs` default 36 is conservative (API allows up to ~2048 inputs for `text-embedding-3-*`). Investigate a higher default and/or sending sub-batches concurrently with a bounded cap, balanced against payload size and rate limits.
3. **ONNX adaptive batch sizing** — fixed 32 can OOM on large tensors / under-utilize otherwise; consider memory-aware tuning.

## Acceptance criteria

- [ ] Eliminate the double embedding of chunks/entities/summaries per `cognify` (embed once, reuse for indexing); preserve the reported embedding count and all `integration_embeddings` assertions.
- [ ] Benchmark embedding call count + wall-clock before/after on a large document (remote API and local ONNX).
- [ ] Decide on Ollama batch-input support and implement if available.
- [ ] Evaluate raising the OpenAI sub-batch default and/or concurrent sub-batches; document chosen value.

_Investigation issue — verify file/line anchors against `main` before implementing._

---
Part of #22 (external-API/I/O performance umbrella).

hassaansaleem28:
Closes #20.

Every cognify run was embedding the same chunks, entities and summaries twice: once
in generate_embeddings (only used for a count) and again in index_data_points before
writing to the vector database. This doubles embedding requests and latency on remote
APIs and doubles CPU on local ONNX.

What changed:

* generate_embeddings now runs once and index_data_points reuses those vectors. It
  only embeds what generate_embeddings never produced (entity types, triplets, edge
  types), with a fallback so nothing goes unembedded. Reported counts and existing
  assertions are unchanged.
* Ollama now sends a batched array input instead of one request per text, and falls
  back to the old per text path on older servers that reject arrays. Both the batch
  path and the fallback are covered by mock server tests.
* The OpenAI compatible engine now sends its sub batches concurrently with a bounded
  cap instead of one after another. The request batch default stays at 36 (Python's
  value); raise it with EMBEDDING_BATCH_SIZE for providers that allow larger batches.
* The ONNX inference batch is decoupled from the request batch and keeps its own
  default of 32, so raising EMBEDDING_BATCH_SIZE no longer changes tensor sizes on
  constrained edge devices.

A small test prints the saving on a fixed input:

  cargo test -p cognee-cognify --lib report_embedding_reuse_savings -- --nocapture

  Embedding work per cognify (24 chunks / 16 entities / 10 summaries):
    BEFORE (double pass): 7 embed() calls, 104 texts
    AFTER  (reuse)      : 4 embed() calls, 54 texts
    Saved: 50 texts (48% fewer)

Verified locally with cargo check on the workspace, clippy with warnings denied, fmt,
and the embedding and cognify unit tests. Provider backed integration tests and the
cross SDK parity suite are left to CI. No changes to IDs, schema, content hashing or
vector collection names, so Python parity is unaffected.

### DIFF
--- a/crates/cognify/src/tasks.rs
+++ b/crates/cognify/src/tasks.rs
@@ -2652,6 +2652,7 @@ async fn reuse_or_embed(
     ids: &[Uuid],
     texts: &[&str],
 ) -> Result<Vec<Vec<f32>>, CognifyError> {
+    debug_assert_eq!(ids.len(), texts.len(), "ids and texts must be parallel");
     let missing_texts: Vec<&str> = ids
         .iter()
         .zip(texts)
--- a/crates/cognify/src/tasks.rs
+++ b/crates/cognify/src/tasks.rs
@@ -1098,6 +1098,7 @@ pub async fn add_data_points(
         embedding_engine,
         vector_db,
         config,
+        &embeddings,
     )
     .await?;
 
@@ -2642,6 +2643,42 @@ async fn generate_embeddings(
     Ok(embeddings)
 }
 
+/// Return one vector per `texts[i]`, reusing `precomputed[ids[i]]` when present
+/// and embedding only the texts whose id is missing. `ids` and `texts` must be
+/// parallel slices.
+async fn reuse_or_embed(
+    engine: &Arc<dyn EmbeddingEngine>,
+    precomputed: &std::collections::HashMap<Uuid, Vec<f32>>,
+    ids: &[Uuid],
+    texts: &[&str],
+) -> Result<Vec<Vec<f32>>, CognifyError> {
+    let missing_texts: Vec<&str> = ids
+        .iter()
+        .zip(texts)
+        .filter(|(id, _)| !precomputed.contains_key(*id))
+        .map(|(_, text)| *text)
+        .collect();
+
+    let fresh = if missing_texts.is_empty() {
+        Vec::new()
+    } else {
+        engine
+            .embed(&missing_texts)
+            .await
+            .map_err(|e| CognifyError::EmbeddingError(e.to_string()))?
+    };
+
+    let mut fresh = fresh.into_iter();
+    ids.iter()
+        .map(|id| match precomputed.get(id) {
+            Some(vector) => Ok(vector.clone()),
+            None => fresh
+                .next()
+                .ok_or_else(|| CognifyError::EmbeddingError("missing fresh embedding".into())),
+        })
+        .collect()
+}
+
 /// Index data points in vector database.
 #[allow(clippy::too_many_arguments)]
 async fn index_data_points(
@@ -2657,9 +2694,18 @@ async fn index_data_points(
     engine: Arc<dyn EmbeddingEngine>,
     vector_db: Arc<dyn VectorDB>,
     config: &CognifyConfig,
+    precomputed_embeddings: &[Embedding],
 ) -> Result<IndexedFieldsStats, CognifyError> {
     let mut stats = IndexedFieldsStats::default();
     let dimension = engine.dimension();
+
+    // Vectors already produced by `generate_embeddings`, keyed by data point id,
+    // so the chunk/entity/summary collections below reuse them rather than
+    // re-embedding the same text.
+    let precomputed: std::collections::HashMap<Uuid, Vec<f32>> = precomputed_embeddings
+        .iter()
+        .map(|e| (e.data_point_id, e.vector.clone()))
+        .collect();
 
     // 1. Index DocumentChunk.text field
     if !chunks.is_empty() {
@@ -2674,11 +2720,9 @@ async fn index_data_points(
                 .map_err(|e| CognifyError::VectorDBError(e.to_string()))?;
         }
 
+        let ids: Vec<Uuid> = chunks.iter().map(|c| c.base.id).collect();
         let texts: Vec<_> = chunks.iter().map(|c| c.text.as_str()).collect();
-        let vectors = engine
-            .embed(&texts)
-            .await
-            .map_err(|e| CognifyError::EmbeddingError(e.to_string()))?;
+        let vectors = reuse_or_embed(&engine, &precomputed, &ids, &texts).await?;
 
         let points: Vec<VectorPoint> = chunks
             .iter()
@@ -2731,11 +2775,9 @@ async fn index_data_points(
                 .map_err(|e| CognifyError::VectorDBError(e.to_string()))?;
         }
 
+        let ids: Vec<Uuid> = entities.iter().map(|e| e.entity.base.id).collect();
         let names: Vec<_> = entities.iter().map(|e| e.entity.name.as_str()).collect();
-        let vectors = engine
-            .embed(&names)
-            .await
-            .map_err(|e| CognifyError::EmbeddingError(e.to_string()))?;
+        let vectors = reuse_or_embed(&engine, &precomputed, &ids, &names).await?;
 
         let points: Vec<VectorPoint> = entities
             .iter()
@@ -2850,11 +2892,9 @@ async fn index_data_points(
                 .map_err(|e| CognifyError::VectorDBError(e.to_string()))?;
         }
 
+        let ids: Vec<Uuid> = summaries.iter().map(|s| s.base.id).collect();
         let texts: Vec<_> = summaries.iter().map(|s| s.text.as_str()).collect();
-        let vectors = engine
-            .embed(&texts)
-            .await
-            .map_err(|e| CognifyError::EmbeddingError(e.to_string()))?;
+        let vectors = reuse_or_embed(&engine, &precomputed, &ids, &texts).await?;
 
         let points: Vec<VectorPoint> = summaries
             .iter()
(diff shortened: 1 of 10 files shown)


GraDKh:
**ONNX inference batch is silently raised 32→100, and `EMBEDDING_ONNX_BATCH_SIZE` doesn't take effect on the normal path**

`crates/lib/src/component_manager.rs:509` wires `onnx.batch_size = settings.embedding_batch_size`, and this PR bumps that default 32→100. So through the CLI/library path (the one the comment at `component_manager.rs:433` calls canonical), the ONNX *inference* batch — the leading dimension of the activation tensors, not just the OpenAI request size — goes from 32 to 100. On the constrained edge/Android devices this project targets, that's exactly the OOM/slowdown risk #20 flagged. The PR description's "ONNX default stays 32" isn't true on this path.

The new `EMBEDDING_ONNX_BATCH_SIZE` is only read by `EmbeddingConfig::from_env()`, which this path doesn't call, so it can't be used to pull the value back down — the knob is effectively dead where it ships.

Since only one embedding provider is live per run, the two batch sizes never coexist. I'd suggest dropping the separate knob and instead decoupling the mapping so ONNX keeps its own default:

```rust
// component_manager.rs
batch_size: OnnxEmbeddingConfig::default().batch_size, // stays 32; API batch stays 100
```

That fixes the regression with the least surface. If runtime provider-switching (local offline / remote online) is a real use case, a single `Option<u32>` resolved per provider would be the next step up — but a standalone ONNX env var that only works via `from_env` is the one option I'd avoid. (For reference, Python cognee uses a single `embedding_batch_size` for all providers.)

GraDKh:
**Why raise the default batch size 36→100?**

36 (Python's value; ~32 here) looks like it was tuned experimentally to stay within real request limits, and the increase's stated rationale — "well within the OpenAI `/v1/embeddings` input cap (~2048 for `text-embedding-3-*`)" — is about the *input-count* cap, which wasn't the binding constraint. The limits that actually matter here are total tokens per request and the **server-side client-batch limit**, and this engine targets self-hosted servers (`vLLM, llama.cpp, TEI, LocalAI`) that default small — e.g. TEI's max client batch size is 32. At 100, a default TEI/llama.cpp server rejects the request as a non-retryable 4xx (→ `ApiError`, not a retryable 429), which is a hard regression for exactly the setups this adapter exists to serve.

The token-per-request total cuts the same way: 100 × per-chunk tokens roughly triples the per-request token count vs. 36, pushing toward the cap that the smaller value was tuned to stay under.

Suggest keeping the default at the validated value (36, for Python parity) and making 100 opt-in via `EMBEDDING_BATCH_SIZE`, unless there's a before/after benchmark on a real provider justifying the bump.

GraDKh:
**Thought (no action needed for this PR):** the within-call batching here is a fine incremental step. The more general win for a concurrent workload would be a coalescer that batches embed requests *across* independent callers (e.g. many concurrent search queries each embedding one text), flushing on a full batch or a short timeout — which would let the per-engine concurrency logic be simplified back down later.

Worth noting it's *not* a parity feature: I checked, and the Python SDK doesn't coalesce either — it does per-call `batch_size` chunking (default 36) plus an `AsyncLimiter` rate limiter, with Ollama still one request per text. So this PR's Ollama array-batching is actually ahead of Python.

Captured as a follow-up (with the honest "where it helps / where it doesn't" and a cheaper rate-limiter alternative): #39. Nothing to change here.

GraDKh:
**The Ollama array→per-text fallback path has no test.**

The new unit tests cover `extract_all_embeddings_from_value` parsing shapes, but the actual compatibility behavior in `embed_all` (`ollama.rs:276`) — array request fails with a non-transient `ApiError` (e.g. an older server that rejects array `input`) → fall back to one concurrent request per text — is the whole point of the change and is untested. The existing tests in this file are all pure/parse-only (no HTTP mock), so nothing exercises that branch end to end.

Worth adding a mock-server test (e.g. wiremock) that returns a 4xx / count-mismatch on the batch endpoint and asserts the per-text fallback still yields correct, correctly-ordered embeddings. If a full HTTP mock is too heavy, factoring the "batch failed → fall back" decision into a small testable unit would at least cover the dispatch logic. As-is, a regression that silently breaks old-Ollama compatibility wouldn't be caught.

GraDKh (changes requested):
Thanks for taking this on, @hassaansaleem28 — the headline change is genuinely valuable. The double-embedding reuse (`reuse_or_embed`) is correct and well-tested, and the embedding-call-count test is exactly the right way to demonstrate the win. The Ollama array-batching (with fallback) is also a nice step that actually puts us ahead of the Python SDK here.

I'm marking this as changes-requested mainly for the bundled config-default changes, which I think need another look before merge:

- **ONNX batch silently raised 32→100** through `component_manager.rs:509`, and the new `EMBEDDING_ONNX_BATCH_SIZE` doesn't take effect on the CLI/library path — the OOM risk #20 flagged on edge devices. (Suggested decoupling fix in the inline comment.)
- **Default batch size 36→100** is justified against the input-count cap, but the binding limit for the self-hosted servers this adapter targets (TEI/llama.cpp, default ~32) is smaller — 100 can cause hard 4xx rejections. Suggest keeping the validated 36 and making 100 opt-in.
- **Ollama fallback path** (array→per-text) has no end-to-end test.

None of this touches the core reuse logic, which is solid — it's about not coupling the perf win to riskier default changes. Happy to talk through any of it. The broader cross-caller batching idea is captured separately in #39 (no action needed here). Thanks again!

hassaansaleem28:
Hi @GraDKh, changes are addressed now. LMK if any further changes are required.

crates/embedding/src/ollama.rs
```diff
@@ -176,23 +190,113 @@ impl OllamaEmbeddingEngine {
...
-        let futures: Vec<_> = sanitized_refs
-            .iter()
-            .map(|&text| self.embed_single_with_retry(text))
-            .collect();
-
-        let results = future::join_all(futures).await;
-
-        let embeddings: EmbeddingResult<Vec<Vec<f32>>> = results.into_iter().collect();
+        let mut embeddings: Vec<Vec<f32>> = Vec::with_capacity(texts.len());
+        for batch in sanitized_refs.chunks(self.batch_size) {
```
GraDKh: **Crash on `EMBEDDING_BATCH_SIZE=0`.** The old `embed_all` used `join_all` over all texts and never chunked, so `batch_size` couldn't crash it. `EMBEDDING_BATCH_SIZE` is parsed as a bare `usize` with no lower bound (`config.rs:422-425`), so `EMBEDDING_BATCH_SIZE=0` yields `batch_size == 0` and `slice::chunks(0)` panics (`"chunk size must be non-zero"`) — every `embed()` call aborts the process instead of returning an error.

```suggestion
        for batch in sanitized_refs.chunks(self.batch_size.max(1)) {
```

crates/lib/src/component_manager.rs
```diff
@@ -506,7 +506,11 @@ impl ComponentManager {
                     model_name: settings.embedding_model_name.clone(),
                     dimensions: settings.embedding_dimensions as usize,
                     max_sequence_length: settings.embedding_max_sequence_length as usize,
-                    batch_size: settings.embedding_batch_size as usize,
+                    // ONNX inference batch is the leading tensor dimension, not an
+                    // HTTP request size, so it keeps its own (memory-tuned) default
+                    // rather than inheriting embedding_batch_size, which can OOM on
+                    // constrained edge/Android devices.
+                    batch_size: cognee_embedding::OnnxEmbeddingConfig::default().batch_size,
```
GraDKh: **ONNX inference batch is now unconfigurable at runtime.** Pinning to `OnnxEmbeddingConfig::default().batch_size` (32) means `EMBEDDING_BATCH_SIZE` no longer reaches the ONNX path — the *only* path where batch size is a memory-critical tensor dimension, and exactly the constrained edge/Android target this PR cites. A user who needs to *lower* it to 8 under memory pressure (or raise it for throughput) now has no knob at all.

Consider a dedicated override (e.g. `ONNX_BATCH_SIZE` / `EMBEDDING_ONNX_BATCH_SIZE`) that falls back to the ONNX default, rather than hardcoding it — so the decoupling from the HTTP request size is kept but ONNX stays tunable.

crates/embedding/src/ollama.rs
```diff
@@ -176,23 +190,113 @@ impl OllamaEmbeddingEngine {
...
+            match self.embed_batch_with_retry(batch).await {
+                Ok(batch_embeddings) => embeddings.extend(batch_embeddings),
+                Err(EmbeddingError::ApiError(_)) => {
```
GraDKh: **Fallback fires on *any* `ApiError` → `1 + N` request storm.** `embed_batch_once` classifies every non-429/non-5xx status (e.g. 404 model-not-found, 400 bad request) as `ApiError`, and this arm treats all `ApiError` as "server rejects array input," re-issuing one request per text. For a real 404/400, all N per-text requests also fail — so the caller does `1 + N` requests and surfaces the *per-text* error instead of the more informative batch error.

The same mechanism makes legacy servers that return the single `{"embedding":[...]}` shape (→ count mismatch → `ApiError`) fall back on *every* batch, so the fast path is silently never used for that server class.

Consider distinguishing "array unsupported" from other API errors — e.g. only fall back on the count-mismatch/unrecognised-shape signal, not on arbitrary HTTP 4xx.

crates/embedding/src/openai_compatible.rs
```diff
@@ -189,11 +195,21 @@ impl EmbeddingEngine for OpenAICompatibleEmbeddingEngine {
...
 
-        let mut results: Vec<Vec<f32>> = Vec::with_capacity(texts.len());
+        // Dispatch sub-batches concurrently (bounded by MAX_CONCURRENT_BATCHES).
+        // `buffered` preserves input order, so the flattened result still lines
+        // up one-to-one with `texts`.
+        let batch_futures: Vec<_> = texts
+            .chunks(self.batch_size)
+            .map(|batch| self.embed_batch_with_retry(batch))
+            .collect();
+        let batch_results: Vec<EmbeddingResult<Vec<Vec<f32>>>> = stream::iter(batch_futures)
```
GraDKh: **Blocks on all in-flight retries before surfacing an early error.** The old sequential loop returned on the first failed batch and never issued later ones. Now every sub-batch is dispatched; if one fails fast (e.g. 400) while others are mid-retry against a 429/5xx with the 8s→128s backoff, `embed()` blocks up to ~128s before returning that fast error. Order and first-error selection are preserved, so output is correct — this is a wasted-latency/request regression under partial failure.

Since this is an optimization PR, worth considering short-circuiting (e.g. `buffer_unordered` + early-abort on the first non-retryable error, or a shared cancellation flag) so a fast failure doesn't wait out the slow retriers.

crates/cognify/src/tasks.rs
```diff
@@ -2642,6 +2643,42 @@ async fn generate_embeddings(
     Ok(embeddings)
 }
 
+/// Return one vector per `texts[i]`, reusing `precomputed[ids[i]]` when present
+/// and embedding only the texts whose id is missing. `ids` and `texts` must be
+/// parallel slices.
+async fn reuse_or_embed(
```
GraDKh, line 2649: **No length-equality guard on the parallel slices.** All three current call sites build `ids` and `texts` from the same collection, so they're always aligned — but the "parallel slices" precondition is only a doc comment. If a future caller passes mismatched lengths, `zip` silently truncates `missing_texts`, `engine.embed` produces too few vectors, and they misalign to ids with no guard (either a `"missing fresh embedding"` error or silently-wrong vectors).

A cheap `debug_assert_eq!(ids.len(), texts.len());` at the top makes the contract enforceable.

GraDKh (changes requested):
Thanks — the core optimization is sound. The reuse keying (`chunk.base.id`→text, `entity.base.id`→name, `summary.base.id`→text) is identical and same-ordered in both `generate_embeddings` and `index_data_points`, `buffered(8)` preserves order, and EntityType/Triplet/EdgeType are correctly left to embed fresh. No data-corruption bugs.

A few things to address before merge — one is a genuine crash path, and since this is a performance PR, the latency/wasted-work regressions under partial failure are in scope too:

1. **(bug)** Ollama `chunks(batch_size)` panics when `EMBEDDING_BATCH_SIZE=0` — new crash path.
2. **(regression)** ONNX inference batch is no longer tunable at runtime.
3. **(perf)** Ollama fallback fires on *any* `ApiError` → `1 + N` request storm + wrong surfaced error.
4. **(perf)** OpenAI `buffered().collect()` blocks on all in-flight retries before surfacing an early error.
5. **(defensive)** `reuse_or_embed` has no length-equality guard on its parallel slices.

Details inline.

hassaansaleem28:
Hey @GraDKh, thanks for review -- changes addressed! 
Verified locally with cargo check on the workspace, clippy with warnings denied, fmt, and
the embedding and cognify unit tests. Provider backed integration tests and the binding
checks are left to CI.

crates/cognify/src/tasks.rs
```diff
@@ -2642,6 +2643,43 @@ async fn generate_embeddings(
...
+    ids: &[Uuid],
+    texts: &[&str],
+) -> Result<Vec<Vec<f32>>, CognifyError> {
+    debug_assert_eq!(ids.len(), texts.len(), "ids and texts must be parallel");
+    let missing_texts: Vec<&str> = ids
+        .iter()
+        .zip(texts)
+        .filter(|(id, _)| !precomputed.contains_key(*id))
+        .map(|(_, text)| *text)
+        .collect();
```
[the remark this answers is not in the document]
GraDKh: Not a compile error — this is a false positive. The **Lint (fmt + check + clippy)** and **MSRV (1.91)** CI checks both pass on this commit, which run `cargo check` and `cargo clippy -D warnings`.

The types work out via match ergonomics: `ids.iter().zip(texts)` yields `(&Uuid, &&str)`, so `filter`'s closure binds `id: &&Uuid`, and `*id` is therefore `&Uuid` — exactly what `HashMap<Uuid, _>::contains_key(&Q)` expects (`Q = Uuid`). No change needed.

crates/cognify/src/tasks.rs
```diff
@@ -2657,10 +2695,19 @@ async fn index_data_points(
...
     let mut stats = IndexedFieldsStats::default();
     let dimension = engine.dimension();
 
+    // Vectors already produced by `generate_embeddings`, keyed by data point id,
+    // so the chunk/entity/summary collections below reuse them rather than
+    // re-embedding the same text.
+    let precomputed: std::collections::HashMap<Uuid, Vec<f32>> = precomputed_embeddings
+        .iter()
+        .map(|e| (e.data_point_id, e.vector.clone()))
+        .collect();
```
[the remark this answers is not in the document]
GraDKh: Accurate observation, but the impact is negligible and not worth changing here. The double `Vec<f32>` clone is a memcpy of a few KB per vector; the cost this PR eliminates is a redundant **embedding call** (network round-trip / ONNX inference), which is orders of magnitude more expensive. So it does not meaningfully offset the perf win.

The suggested `HashMap<Uuid, &[f32]>` is valid (it can borrow from `precomputed_embeddings`, which outlives the function) but removes only the map-build clone — the reused vector still needs an owned `Vec` for `VectorPoint`. Marginal gain for added lifetime complexity; leaving as-is. Can revisit as a follow-up if profiling ever flags it.

GraDKh (approved):
Thanks for your efforts! Now this one looks good, merging it.

Issue #21, opened 2026-06-26
Perf: optimize database batching & eliminate N+1 round-trips (graph/relational/vector)


GraDKh:
## Summary

Investigate and optimize **database access** (relational via SeaORM/SQLite/Postgres, graph DB, vector DB) for batching and N+1 elimination. Bulk node/edge/vector upserts are already chunked and correct, but there are two confirmed N+1 round-trip loops on hot paths, no transactional grouping of related writes, and default (unconfigured) connection pooling.

This mirrors the relational-grouping work done on the Python side (Linear COG-4160). LLM and embedding slices are tracked separately. Complements #18.

## Current state (verified)

**Backends**
- Relational (SeaORM) — `crates/database/src/traits/ingest_db.rs`, ops under `crates/database/src/ops/`.
- Vector — `crates/vector/` (`pgvector_adapter.rs`, `lancedb_adapter.rs`).
- Graph — `crates/graph/` (`pg_graph_adapter.rs`, `ladybug.rs`).

**Already good (batched)**
- `crates/database/src/ops/graph_storage.rs:24` — `upsert_nodes` chunks at 500 with `insert_many` + `on_conflict`.
- `crates/vector/src/pgvector_adapter.rs` — `index_points` batches at 100 with cross-dataset metadata union.
- `crates/vector/src/lancedb_adapter.rs:298` — single `RecordBatch` add per collection.

## Findings / opportunities

1. **N+1: `has_edges` — one query per edge** — `crates/graph/src/pg_graph_adapter.rs:540`. Loops `has_edge` per edge (there's already a `TODO` noting it). For E edges → E round-trips. Fix: single statement using `unnest($1::text[], $2::text[], $3::text[])` to check all candidates at once. Highest-impact (called on bulk edge ingestion).
2. **N+1: `update_last_accessed` — fetch + update per id** — `crates/database/src/ops/data.rs:152`. Loops `find_by_id` then `update` per id → 2N round-trips. Fix: single `UPDATE ... SET last_accessed = ? WHERE id IN (...)` via `update_many`.
3. **No transactions around related bulk writes** — `upsert_nodes` and `upsert_edges` (and provenance) run as separate statements; a crash mid-way leaves an inconsistent graph. Wrap logically-atomic groups in `begin/commit`. (Reliability + lets the driver pipeline within one round-trip group.)
[...]

PR #35, opened 2026-06-29
perf(db): wrap bulk provenance writes in a transaction + configure connection pool

Taimoo-r:
PR 2 of #21: transactions + connection pool.

**Transactions.** `upsert_provenance` writes the node and edge batches through `upsert_provenance_graph`, which wraps both in a single `db.begin()`/`commit()`, so a failure partway through rolls the whole group back rather than leaving provenance half-written. The public `upsert_nodes`/`upsert_edges` signatures stay concrete (`&DatabaseConnection`) for semver safety; a private connection-generic impl is what runs inside the transaction. On SQLite the group holds the single writer lock for its duration, which is a deliberate trade for atomicity: under WAL readers are never blocked, and a competing writer waits out `busy_timeout` rather than failing.

**Connection pool.** `connect()` now configures the pool explicitly (`max_connections`, `min_connections`, `acquire_timeout`, `idle_timeout`). The sizing keeps sqlx's own defaults (max 10, min 0) rather than inventing numbers: an embedded SQLite database has one writer regardless, so the pool only ever buys concurrent readers under WAL, and a file database is free to fall back to zero connections when idle.

SQLite is built directly on the sqlx pool because sea-orm's `ConnectOptions` cannot express what this path needs:

- **File-backed, writable:** WAL + `synchronous=NORMAL`, plus `busy_timeout=120s` matching Python's `SqlAlchemyAdapter` (the #2717 "database is locked" fix).
- **Read-only:** WAL is gated on real filesystem writability rather than the URL alone, so a read-only mount, a read-only file, a read-only parent directory, or an explicit `?mode=ro` opens read-only and issues no journal pragma. The probe uses the driver's decoded filename (so percent-encoded paths are handled) and checks that the parent directory can hold the `-wal`/`-shm` sidecars.
- **In-memory:** at least one connection is kept and both sqlx reapers are disabled, since the database lives only as long as its connections. A non-shared `:memory:` URL is additionally capped at one connection.

**Behavior change worth flagging (WAL + `synchronous=NORMAL`).** These pragmas match Python's adapter, but against the previous Rust behavior (no journal/synchronous pragma, so SQLite's DELETE + FULL defaults) they are a real change:

- `NORMAL` under WAL can lose the last committed transactions on power loss (it stays safe on process crash). This was effectively FULL before.
- WAL is persisted in the database header and needs shared memory plus proper file locking, so it can break on some NFS/SMB and container bind-mounts.
- WAL creates `-wal`/`-shm` sidecars, so a single-file `cp` backup taken while the database is running is incomplete.

Gating WAL behind a config opt-out for network-filesystem deployments looks like a sensible follow-up shared by both SDKs.

**Tests.** `provenance_transaction.rs` asserts that a mid-group failure rolls back both the node and the edge writes, and that the happy path commits atomically; a cognify-level test covers the `upsert_provenance` wrapper and its guard deterministically, with no LLM. `connection_pool.rs` pins the file vs in-memory pool behavior, WAL mode, and the read-only paths (URL, file permissions, and percent-encoded paths). Everything runs on SQLite with no external services.

Scoped to PR 2; PR 3 (audit + minor items) to follow.


crates/database/src/connection.rs
```diff
@@ -1,12 +1,39 @@
...
-    Database::connect(url)
+    let mut opt = ConnectOptions::new(url.to_owned());
+    opt.acquire_timeout(POOL_ACQUIRE_TIMEOUT);
+
+    // SQLite (the default/embedded backend, including `sqlite::memory:`) must use
+    // a single connection: writes serialise on one file lock anyway, and every
+    // extra pooled connection to `:memory:` would get its own separate empty
+    // database. A server backend (Postgres) gets a real pool shared with the
+    // graph adapter.
+    if url.starts_with("sqlite") {
```
GraDKh, line 96: Two different concerns are fused into this one prefix check, which is what makes it feel like a leaky abstraction:

- **Correctness invariant** — a pooled connection to a *non-shared* `:memory:` SQLite DB gets its own empty database, so exactly-one-connection is the only thing that makes `sqlite::memory:` work at all.
- **Perf policy** — `max_connections = 10` plus the timeouts for a server backend.

Separately, `starts_with("sqlite")` also pins **file-backed** SQLite to a single connection, which needlessly serializes concurrent reads (e.g. the http-server on a file DB) where the old `Database::connect(url)` default allowed a pool.

Suggested split:

**1. Keep only the irreducible invariant here, keyed on *in-memory* rather than *sqlite*** (and skip it for `?cache=shared`, which genuinely does share one DB across connections):

```rust
// sea-orm/sqlx gives every pooled connection to a non-shared `:memory:`
// database its own empty DB, so in-memory SQLite MUST be a single
// connection. This is a backend invariant, not tuning.
fn is_exclusive_in_memory_sqlite(url: &str) -> bool {
    url.starts_with("sqlite")
        && url.contains(":memory:")
        && !url.contains("cache=shared")
}
```

**2. Move the pool-sizing *policy* (max/min connections, idle timeout) to the config layer** that already selects the URL (`lib/config.rs`, `http-server/config.rs`). Have it build a `ConnectOptions` (or a small `PoolConfig`) and pass it into `connect()`; then `connect()` just applies what it's given and enforces the in-memory invariant on top — no backend guessing left in generic plumbing.

Note there is no relational "SQLite adapter" object to host this (unlike the vector/graph adapters — the relational path is raw sea-orm), so the URL is the only signal at connect time. That's fine: detecting *in-memory* is detecting an intrinsic property of the URL, whereas pool *sizing* is policy that belongs with whoever chose the backend.

crates/database/src/connection.rs
```diff
@@ -1,12 +1,39 @@
...
+        opt.max_connections(1).min_connections(1);
+    } else {
+        opt.max_connections(POOL_MAX_CONNECTIONS)
+            .min_connections(POOL_MIN_CONNECTIONS)
+            .idle_timeout(POOL_IDLE_TIMEOUT);
+    }
+
+    Database::connect(opt)
```
GraDKh, line 106: Please add a regression test for the pool-sizing behaviour — this is the part the new tests don't cover, and it's exactly where the file-vs-memory distinction can silently regress. sea-orm 1.1 exposes the underlying sqlx pool, so it's directly assertable:

```rust
use cognee_database::connect;
use sea_orm::ConnectionTrait; // for get_sqlite_connection_pool via DatabaseConnection

#[tokio::test]
async fn in_memory_sqlite_is_single_connection() {
    let db = connect("sqlite::memory:").await.expect("connect");
    assert_eq!(
        db.get_sqlite_connection_pool().options().get_max_connections(),
        1,
        "in-memory SQLite must be pinned to one connection",
    );
}

#[tokio::test]
async fn file_sqlite_allows_a_pool() {
    let dir = tempfile::tempdir().unwrap();
    let url = format!("sqlite://{}?mode=rwc", dir.path().join("t.db").display());
    let db = connect(&url).await.expect("connect");
    assert!(
        db.get_sqlite_connection_pool().options().get_max_connections() > 1,
        "file-backed SQLite should not be pinned to a single connection",
    );
}
```

(`get_sqlite_connection_pool()` is a direct method on `DatabaseConnection`; `.options().get_max_connections()` is the sqlx getter.) The second test will fail against the current `starts_with("sqlite")` logic — which is the point: it pins down the desired post-fix behaviour.

crates/database/tests/provenance_transaction.rs
```diff
@@ -0,0 +1,104 @@
...
+            indexed_fields: json!({ "index_fields": ["name"] }),
+            attributes: None,
+            created_at: Utc::now(),
+        })
+        .collect()
+}
+
+/// A failure partway through the node+edge group rolls the node write back too.
+#[tokio::test]
+async fn provenance_group_rolls_back_on_midway_failure() {
```
GraDKh: These two tests re-implement the `begin` → upsert → `commit`/`rollback` sequence by hand against `upsert_nodes`/`upsert_edges`; neither calls `upsert_provenance`, which is the function this PR actually changed. So a regression in `upsert_provenance`'s own wiring — an upsert moved back outside the `if`, or a dropped `commit()` — would still pass green here. What's verified is `ConnectionTrait` genericity + SQLite rollback semantics, not the call-site change.

Since `upsert_provenance` is private with a heavy signature, two pragmatic ways to close the gap: (a) expose a thin testable seam over the node+edge write, or (b) add a cognify-level test that injects a failure between the node and edge writes and asserts no provenance row persisted. Not blocking on its own, but worth a follow-up; at minimum please land the pool-sizing test from the `connection.rs` comment, since that guards the other half of this PR.

GraDKh (changes requested):
Thanks for splitting this out of #21 and for adding the rollback tests — the transaction wrapping itself is clean, and making `upsert_nodes`/`upsert_edges`/`database_system_label` generic over `ConnectionTrait` reads well and stays source-compatible with every existing caller. 🙏

Requesting changes on one design point in the pool config: `connect()` decides pool sizing from a `starts_with("sqlite")` string check, which (1) fuses a hard backend invariant with tunable perf policy inside generic plumbing, and (2) over-pins *file*-backed SQLite to a single connection — regressing concurrent reads versus the previous `Database::connect(url)` default. Suggested split + a regression test are inline. Happy to discuss how far to take the config-layer boundary if option 2 is more churn than you want in this PR.

crates/database/src/connection.rs
```diff
@@ -1,12 +1,78 @@
...
+///
+/// The pool sizing comes entirely from `pool`; the only thing decided here is the
+/// exclusive-in-memory-SQLite invariant, which forces a single connection on top
+/// of whatever sizing was requested.
+pub async fn connect_with_pool(
+    url: &str,
+    pool: PoolConfig,
+) -> Result<DatabaseConnection, DatabaseError> {
+    let mut opt = ConnectOptions::new(url.to_owned());
+    opt.max_connections(pool.max_connections)
```
GraDKh: **File-backed SQLite goes from 1 → 10 connections here, but stays in rollback-journal mode — which is the worst of both worlds.**

The old `connect()` left `max_connections` unset, so sea-orm's SQLite driver auto-pinned it to 1 (`driver/sqlx_sqlite.rs:85-86`). `connect_with_pool` now always sets `.max_connections(10)`, defeating that pin for the default file-backed `sqlite:./cognee.db?mode=rwc`, and `file_sqlite_allows_a_pool` locks the new behavior in.

Grounded concurrency facts:
- SQLite writers serialize regardless of journal mode, so extra connections give **zero** write-throughput gain (a known benchmark shows a large pool ~20× *slower* than a single writer connection for writes).
- WAL's only win is reader/writer concurrency. In the current rollback journal a writer takes an EXCLUSIVE lock that blocks readers too — so 10 connections buys no read concurrency either, just more contention for the single write lock.
- The 5s `busy_timeout` default masks this until a writer holds the lock >5s — e.g. the new provenance transaction batching thousands of rows on a large graph — at which point a concurrent writer fails with `SQLITE_BUSY "database is locked"`. Under the old single-connection pool this could never happen.

Two coherent options (the current state is neither):
- **A — enable WAL for file-backed URLs**: `journal_mode=WAL`, `synchronous=NORMAL`, keep `busy_timeout`, `wal_autocheckpoint=1000`. Gains real read concurrency; must be gated to file URLs (WAL is silently ignored for `:memory:`).
- **B — re-pin file SQLite to 1 connection**: matches prior behavior; removes the BUSY risk now, WAL as a follow-up.

Minor related caveat: `db.begin()` issues a deferred `BEGIN`; write transactions are safest with `BEGIN IMMEDIATE` so `busy_timeout` can serialize writers rather than risk a non-recoverable lock on a read→write upgrade.

crates/database/src/connection.rs
```diff
@@ -1,12 +1,78 @@
...
+
+/// True only for an *exclusive* in-memory SQLite database.
+///
+/// sea-orm/sqlx give every pooled connection to a non-shared `:memory:` database
+/// its own empty DB, so in-memory SQLite must be a single connection. `?cache=shared`
+/// genuinely shares one DB across connections, so it is excluded. This is a backend
+/// invariant, not tuning, which is why it is keyed on *in-memory* and not on the
+/// `sqlite` scheme (file-backed SQLite can and should use a pool).
+fn is_exclusive_in_memory_sqlite(url: &str) -> bool {
+    url.starts_with("sqlite") && url.contains(":memory:") && !url.contains("cache=shared")
```
GraDKh: This only matches the `:memory:` spelling. An exclusive in-memory DB written as `sqlite:file:foo?mode=memory` (no `cache=shared`) won't be detected, so it keeps `max_connections=10` — and each pooled connection then gets its own separate empty DB, the exact failure this guard exists to prevent (migrations run on one connection, a query on another hits "no such table"). No current caller uses that spelling, so this is latent, but worth hardening the predicate to also catch `mode=memory`.

crates/database/src/connection.rs
```diff
@@ -1,12 +1,78 @@
...
+    pool: PoolConfig,
+) -> Result<DatabaseConnection, DatabaseError> {
+    let mut opt = ConnectOptions::new(url.to_owned());
+    opt.max_connections(pool.max_connections)
+        .min_connections(pool.min_connections)
+        .acquire_timeout(pool.acquire_timeout)
+        .idle_timeout(pool.idle_timeout);
+
+    if is_exclusive_in_memory_sqlite(url) {
+        opt.max_connections(1).min_connections(1);
```
GraDKh: For an exclusive in-memory DB the whole database lives inside this one pooled connection — so it must never be reaped, or an in-process demo / one-shot run silently loses all its data mid-flight. Pinning `max_connections(1).min_connections(1)` isn't enough: sqlx's pool defaults still apply `idle_timeout=600s` **and** `max_lifetime=30min` (the latter isn't set anywhere here), and the reaper closes the idle/expired connection then reconnects to satisfy `min_connections` — reconnecting to `:memory:` yields a fresh empty DB.

Watch out: sea-orm's `idle_timeout`/`max_lifetime` setters take `Duration` (no `None`), and sea-orm leaves sqlx's defaults in place for any field you don't set — so *omitting* the call doesn't disable reaping. For the in-memory branch, neutralize **both** timeouts (a large sentinel `Duration`, or build the pool with sqlx `PoolOptions.idle_timeout(None).max_lifetime(None)` + `SqlxSqliteConnector::from_sqlx_sqlite_pool`) so the single connection lives for the process lifetime.

crates/http-server/src/wiring.rs
```diff
@@ -134,7 +134,9 @@ async fn wire_database(cfg: &HttpServerConfig) -> Result<Arc<DatabaseConnection>
         }
     }
 
-    let db = connect(&url)
+    // Pool sizing is chosen here, in the layer that selects the URL; tune via
+    // `PoolConfig` rather than pushing backend guesses into `connect`.
+    let db = connect_with_pool(&url, PoolConfig::default())
```
GraDKh: [nit] `connect_with_pool(&url, PoolConfig::default())` is exactly what `connect()` already does, so this (and the same line in `crates/lib/src/component_manager.rs:168`) is churn for no behavioral change unless a non-default `PoolConfig` is passed. Fine to keep if the intent is "this layer owns pool sizing," but it's currently equivalent to `connect(&url)`.

GraDKh (changes requested):
Nice work — the provenance transaction (atomic node+edge rollback) and the `ConnectionTrait` generalization are clean, correctly threaded through every call site, and well covered by the new tests. My only substantive concerns are on the connection-pool sizing (file-backed SQLite going to a 10-connection pool without WAL, and the idle-reaping of the pinned in-memory connection). If those are addressed along with the smaller notes below, this is good to merge. Requesting changes just to track them.

Taimoo-r:
@GraDKh all changes has been addressed, kindly review it as soon you get a chance.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,175 @@
...
+            .pragma("wal_autocheckpoint", "1000");
+    }
+
+    let mut pool_opts = SqlitePoolOptions::new()
+        .max_connections(pool.max_connections)
+        .min_connections(pool.min_connections)
+        .acquire_timeout(pool.acquire_timeout)
+        .idle_timeout(pool.idle_timeout);
+
+    if exclusive_in_memory {
```
GraDKh: **Silent data loss for `cache=shared` in-memory DBs.** Disabling both reapers only happens in this `exclusive_in_memory` branch. A `sqlite::memory:?cache=shared` URL is deliberately excluded from `is_exclusive_in_memory_sqlite`, so it takes the pooled branch above where `max_lifetime` is never set — sqlx 0.8.6 defaults it to 30 min, and `idle_timeout` is 600s here. sqlx's reaper closes the last connection *before* opening its replacement (`min_connections=1` doesn't keep one continuously alive), so the count hits zero, SQLite frees the shared in-memory DB, and subsequent queries hit an empty database — within ~10 min of idleness. Shared-cache in-memory needs the same `idle_timeout(None).max_lifetime(None)` (+ `min_connections>=1`) treatment as the exclusive case (or just treat it as exclusive).

crates/database/src/connection.rs
```diff
@@ -1,19 +1,175 @@
...
+    let exclusive_in_memory = is_exclusive_in_memory_sqlite(url);
+
+    let mut conn_opts = SqliteConnectOptions::from_str(url)
+        .map_err(|e| DatabaseError::ConnectionError(e.to_string()))?
+        .busy_timeout(Duration::from_secs(5));
+
+    // WAL is silently ignored for :memory:, so only enable it for file-backed DBs.
+    if !exclusive_in_memory {
+        conn_opts = conn_opts
+            .journal_mode(SqliteJournalMode::Wal)
```
GraDKh: **Unconditional WAL is a behavior change that breaks read-only opens.** Pre-PR, `Database::connect` left journal mode untouched — sqlx 0.8.6 deliberately does *not* default to WAL. Now every file-backed connect issues `PRAGMA journal_mode=WAL`, which errors with 'attempt to write a readonly database' for a `?mode=ro` URL or a DB on a read-only mount — both connected fine before (sqlx runs the pragmas at connect and propagates the failure). It also persistently converts existing DBs: committed rows live in the `-wal` sidecar until checkpoint, so a `.db`-only backup/copy (or the Python SDK pointed at a copied file) sees a stale snapshot; and `synchronous=NORMAL` (line 120) trades away the prior FULL durability. WAL+NORMAL is a fine default, but please gate the pragmas so read-only opens don't fail, and call the migration/backup implications out in the PR description.

crates/database/src/ops/graph_storage.rs
```diff
@@ -30,8 +30,8 @@ const PROVENANCE_INSERT_BATCH: usize = 500;
     fields(cognee.db.system = tracing::field::Empty),
     err,
 )]
-pub async fn upsert_nodes(
-    db: &DatabaseConnection,
+pub async fn upsert_nodes<C: ConnectionTrait>(
```
GraDKh: **Breaking API change on a published crate.** Generalizing `&DatabaseConnection` to `<C: ConnectionTrait>` (here and on `upsert_edges` / `database_system_label`) drops deref-coercion — callers passing `&Arc<DatabaseConnection>` no longer compile, which is exactly why the test files in this PR had to switch to `db.as_ref()`. `cognee-database` is published on crates.io (0.1.3) and this PR bumps no version, so the break would ship as a semver-compatible patch and fail external callers on `cargo update`. Please bump the minor line (0.2.0 under 0.x semantics) or otherwise flag the break.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,175 @@
...
+#[derive(Clone, Copy, Debug)]
+pub struct PoolConfig {
+    pub max_connections: u32,
+    pub min_connections: u32,
+    pub acquire_timeout: Duration,
+    pub idle_timeout: Duration,
+}
+
+impl Default for PoolConfig {
+    /// Sized for the relational crate and the Postgres graph adapter sharing one
```
GraDKh: **This rationale isn't accurate.** `PgGraphAdapter::new` (`crates/graph/src/pg_graph_adapter.rs:143`) and `PgVectorAdapter::new` (`crates/vector/src/pgvector_adapter.rs:83`) each call `Database::connect(url)` directly with driver defaults, and their `from_connection` sharing hook is never called anywhere — so nothing actually shares this pool. Under the ingestion workload described, you get three uncoordinated pools hitting one Postgres server, none of them covered by this `PoolConfig`. Either route the adapters through `connect_with_pool` (real sharing), or fix the comment to describe what actually happens.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,175 @@
...
+    use sea_orm::SqlxSqliteConnector;
+    use sea_orm::sqlx::sqlite::{
+        SqliteConnectOptions, SqliteJournalMode, SqlitePoolOptions, SqliteSynchronous,
+    };
+
+    let exclusive_in_memory = is_exclusive_in_memory_sqlite(url);
+
+    let mut conn_opts = SqliteConnectOptions::from_str(url)
+        .map_err(|e| DatabaseError::ConnectionError(e.to_string()))?
+        .busy_timeout(Duration::from_secs(5));
```
GraDKh: **Most of these just restate the defaults.** Verified against sqlx 0.8.6: `busy_timeout` already defaults to 5s, and SQLite's compiled-in `wal_autocheckpoint` is 1000 (line 121) — both no-ops. Same for the server branch (line 80): `max_connections=10` / `acquire_timeout=30s` / `idle_timeout=600s` all equal sqlx's `PoolOptions` defaults; only `min_connections=1` (vs 0) actually differs. The doc comments present these as deliberate tuning, which is misleading — either drop the no-op settings or add a line noting they intentionally pin the upstream defaults.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,175 @@
...
+            .idle_timeout(None)
+            .max_lifetime(None);
+    }
+
+    let sqlx_pool = pool_opts
+        .connect_with(conn_opts)
+        .await
+        .map_err(|e| DatabaseError::ConnectionError(e.to_string()))?;
+
+    Ok(SqlxSqliteConnector::from_sqlx_sqlite_pool(sqlx_pool))
```
GraDKh, line 202: **SQLite silently loses sea-orm's statement logging.** `from_sqlx_sqlite_pool` bypasses sea-orm's `ConnectOptions`, which defaults to `log_statements(Info)`; the raw sqlx default is `Debug`. So SQLite now logs statements at DEBUG while the Postgres path (still via `ConnectOptions`) keeps INFO — a backend-dependent inconsistency under `RUST_LOG=sqlx=info`. Low impact given the repo's default `sqlx=warn` filter, but a `.log_statements(log::LevelFilter::Info)` on `conn_opts` restores parity.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,175 @@
...
+/// sea-orm/sqlx give every pooled connection to a non-shared in-memory database
+/// its own empty DB, so in-memory SQLite must be a single connection. `?cache=shared`
+/// genuinely shares one DB across connections, so it is excluded. This is a backend
+/// invariant, not tuning, which is why it is keyed on *in-memory* and not on the
+/// `sqlite` scheme (file-backed SQLite can and should use a pool).
+///
+/// Both in-memory spellings are matched: the `:memory:` shorthand and the URI form
+/// `file:name?mode=memory` — the latter is exclusive too unless `cache=shared` is set.
+#[cfg(feature = "sqlite")]
+fn is_exclusive_in_memory_sqlite(url: &str) -> bool {
```
GraDKh: **Fragile in-memory detection.** This matches substrings anywhere in the URL, including the path, and DB URLs are user-supplied via `DATABASE_URL` — so a file DB whose path contains `mode=memory` (or `cache=shared`) is misclassified and mis-pooled. `connect_sqlite` already parses the URL with `SqliteConnectOptions::from_str` a few lines down, so prefer deriving the decision from the parsed form, or at minimum split on `?` and match query parameters rather than the whole string. Also note this diverges from the `!url.contains(":memory:")` check in `component_manager.rs:131` (which misses `mode=memory`) — worth exporting one shared predicate so the two layers agree.

crates/http-server/src/wiring.rs
```diff
@@ -136,7 +136,9 @@ async fn wire_database(cfg: &HttpServerConfig) -> Result<Arc<DatabaseConnection>
...
```
GraDKh: **Still equivalent to `connect(&url)`** (same at `crates/lib/src/component_manager.rs:168`). `connect()` is defined as `connect_with_pool(url, PoolConfig::default())`, and nothing reads pool sizing from config/env, so despite the comment this layer doesn't actually 'own pool sizing' yet — this is the open half of last round's thread. Please either wire `PoolConfig` into the `Settings`/`ConfigManager` env-var system (and document it in `docs/configuration.md`) so it's genuinely tunable, or revert both call sites to `connect(&url)` and drop the duplicated comment.

crates/database/tests/provenance_transaction.rs
```diff
@@ -0,0 +1,106 @@
...
```
GraDKh: **Still doesn't exercise `upsert_provenance`** (carried over from the previous review). These tests hand-roll `begin → upsert_nodes/upsert_edges → rollback/commit`, so they verify SQLite rollback semantics and `ConnectionTrait` genericity — but the function this PR actually changed is never called, so a regression in its own wiring (an upsert moved back outside the `if`, a dropped `commit()`) would still pass green. A cognify-level test that injects a failure between the node and edge writes and asserts no provenance row persisted would close the gap.

GraDKh (changes requested):
Thanks for the effort here — the provenance transaction and the `ConnectionTrait` generalization are clean, the in-memory single-connection invariant is now correctly enforced with **both** reapers disabled, and the new `connection_pool.rs` tests pin the file-vs-memory behavior nicely. Most of the substantive pool-correctness points from the last round are genuinely resolved: WAL for file-backed SQLite, `mode=memory` detection, and never-reaping the pinned in-memory connection.

That said, there's still some work left before this is mergeable — a mix of a couple of newly-introduced issues and a thread or two from the previous review that aren't fully closed. Requesting changes to track them. Details are inline; summarized here, most-severe first:

**New issues**
1. **Silent data loss for `cache=shared` in-memory DBs** (`connection.rs`) — excluding `cache=shared` from pinning (correct on its own) routes it to the pooled branch, which never disables `max_lifetime`; sqlx reaps the last connection after ~10 min idle and the shared in-memory DB is freed. This is the *interaction* of the two fixes from the last round, so it slipped through.
2. **Unconditional WAL breaks read-only opens** (`connection.rs`) — `PRAGMA journal_mode=WAL` fails at connect for a `?mode=ro` URL or a read-only filesystem, which worked pre-PR; it also converts existing DBs (`.db`-only copies become stale) and `synchronous=NORMAL` weakens durability vs the prior FULL default.
3. **Breaking API change on a published crate** (`graph_storage.rs`) — making the fns generic drops deref-coercion for `&Arc<DatabaseConnection>`; `cognee-database` is on crates.io and no version bump ships this, so it would land as a semver-compatible patch.
4. **`PoolConfig` doc rationale is false** — the Postgres graph/vector adapters open their own pools; they don't share this one.

**Cleanups / still open from last round**
5. `connect_with_pool(&url, PoolConfig::default())` is still identical to `connect()`, and pool sizing was never wired into config (last round's thread + the nit).
6. Several "tuning" values just restate the underlying defaults.
7. SQLite loses sea-orm's INFO statement logging (bypassed by `from_sqlx_sqlite_pool`) while Postgres keeps it.
8. In-memory detection by whole-string substring match is fragile for user-supplied URLs.
9. The provenance rollback tests still don't call `upsert_provenance` (previous round's thread), so a regression in that function's own wiring would pass green.

For completeness, two things I checked that turned out **fine** and are *not* being requested: the "long transaction → `SQLITE_BUSY`" concern (the txn body is pre-built inserts with no awaits, well under the 5s busy timeout) and in-memory connection starvation (not a regression — sea-orm already pinned SQLite to a single connection). The non-atomic provenance *deletes* in `crates/delete` are pre-existing and fit the announced PR 3, so no action needed here.

crates/database/src/connection.rs
Taimoo-r: Good catch, this was the interaction between the two round 2 fixes. I moved the reaper-disabling out of the exclusive-only branch, so now any in-memory URL, shared cache or not, gets `idle_timeout(None)` and `max_lifetime(None)` with `min_connections` kept at 1. Only the pin to a single connection stays exclusive-only. Added `shared_cache_in_memory_disables_reapers` to lock it in.

crates/database/src/connection.rs
Taimoo-r: Fixed. `classify_sqlite_url` now detects `mode=ro` and `immutable`, and the WAL plus `synchronous=NORMAL` pragmas are gated behind `!in_memory && !read_only`, so a read-only URL or a read-only mount connects without trying to write the journal. `read_only_file_sqlite_connects_and_keeps_journal_mode` seeds a DELETE-mode file with raw sqlx and asserts the read-only open succeeds and leaves the journal mode untouched.

Two operational notes on the WAL default, since you asked me to call them out. Recent commits live in the `-wal` sidecar until checkpoint, so a `.db`-only copy or backup can see a stale snapshot until then. And `synchronous=NORMAL` trades the previous FULL durability, so the last transactions can be lost on power loss but not on a process crash. WAL is gated to file-backed read-write URLs, so read-only and in-memory opens are unaffected. I documented the same trade in the `connect_sqlite` doc comment.

crates/database/src/ops/graph_storage.rs
Taimoo-r: I preserved the API rather than bumping the version. `upsert_nodes` and `upsert_edges` keep their concrete `&DatabaseConnection` signatures as the public surface, and the generic bodies moved into private `_on` helpers. `database_system_label` stays concrete too. The transactional path is a new additive `upsert_provenance_graph`. So deref coercion from `&Arc<DatabaseConnection>` still works and there is no semver break, which is why `delete/src/lib.rs` could go back to `&db` instead of `db.as_ref()`. Happy to bump anyway if you prefer, but this way nothing breaks on `cargo update`.

crates/database/src/connection.rs
Taimoo-r: You are right, the adapters open their own pools. I rewrote the `PoolConfig::default` doc to say this pool serves only the relational database and that `PgGraphAdapter` and `PgVectorAdapter` open their own separate pools. I kept them separate rather than routing them through `connect_with_pool` to keep this PR scoped, but the comment now describes what actually happens.

crates/database/src/connection.rs
Taimoo-r: Dropped the no-op settings. `busy_timeout` and `wal_autocheckpoint` are gone since they only restated the sqlite defaults, and the `PoolConfig` doc now says the values deliberately pin sqlx's own pool defaults except `min_connections=1`, which is the only one that actually differs.

crates/database/src/connection.rs
Taimoo-r: Added `.log_statements(log::LevelFilter::Info)` to the sqlite `ConnectOptions` so it matches the Info level the Postgres path gets through sea-orm, instead of dropping to the raw sqlx Debug default.

crates/database/src/connection.rs
Taimoo-r: Reworked. `classify_sqlite_url` strips the scheme and `file:` prefix, splits path from query on `?`, and matches query parameters exactly, so a file path that contains `mode=memory` no longer gets misclassified. There is a `file_paths_are_never_misclassified_by_substring` test for that. I also exported `sqlite_url_is_in_memory` as the one shared predicate and switched `component_manager` over to it so the two layers agree.

crates/http-server/src/wiring.rs
Taimoo-r: Wired for real now. Added `db_pool_max_connections` and the three siblings to `Settings` and `HttpServerConfig`, reading `DB_POOL_MAX_CONNECTIONS` and friends, and both call sites build a `PoolConfig` from that config instead of `PoolConfig::default()`. Documented under a new Connection pooling section in `docs/configuration.md`. So the config layer genuinely owns pool sizing now.

crates/database/tests/provenance_transaction.rs
Taimoo-r: The provenance write is now a real function, `upsert_provenance_graph`, that both cognify and the tests call, so we exercise the actual wiring instead of hand-rolling begin and commit. `provenance_graph_rolls_back_when_edge_upsert_fails` injects a foreign-key violation between the node and edge writes through the public API and asserts no node provenance persisted, and `provenance_graph_commits_atomically` checks the happy path. A dropped commit or an upsert moved outside the batch would now fail these.

crates/http-server/src/wiring.rs
Taimoo-r: Update after rebasing on main: the cognee-components refactor (#56) landed while this was in review and moved the relational DB connect out of both this wiring layer and lib's component_manager into the shared `cognee_components::build_database`, which calls `connect()`.

Given that, I went with the "revert to `connect(&url)`" option you offered here rather than re-threading pool sizing through the new `BackendBuildContext`. So the `DB_POOL_*` env vars and the `Settings`/`HttpServerConfig` fields are removed, and both call sites now delegate to `build_database`. Pool behavior is unchanged since `connect()` still applies `PoolConfig::default()`: file-backed SQLite keeps its WAL multi-connection pool, in-memory stays pinned with the reapers off, and Postgres gets the sized pool. The only thing dropped is the env-var override. `PoolConfig` and `connect_with_pool` stay in the database crate for any caller that wants explicit sizing later.

Taimoo-r:
Hi @GraDKh, Hackathon is going to end today, so it would be amazing if you review it at your earliest convenience 😋

crates/database/src/ops/graph_storage.rs
```diff
@@ -152,6 +177,38 @@ pub async fn upsert_edges(
...
+/// Upsert a provenance node+edge group atomically in one transaction.
+///
+/// A failure partway through rolls the whole group back (the transaction is
+/// dropped uncommitted, which sea-orm turns into a rollback), so the
+/// provenance graph never ends up half-written.
+///
+/// `begin()` issues a deferred `BEGIN`, but this transaction is write-first:
+/// the first statement is an upsert, which takes SQLite's write lock
+/// immediately, so there is no read-to-write lock upgrade to deadlock on.
+#[instrument(
```
GraDKh, line 243: **[nit / telemetry] Nested `#[instrument(..., err)]` doubles the ERROR events for one provenance failure.**

`upsert_provenance_graph` carries `err`, and it calls `upsert_nodes_on` / `upsert_edges_on` which each also carry `err`. So when an inner upsert returns `Err` (FK violation, `SQLITE_BUSY`, disk-full), the inner span records an ERROR event and the `?` propagation makes this outer span record a *second* ERROR event for the same underlying failure.

Pre-PR, cognify called the two public `upsert_nodes` / `upsert_edges` as separate top-level calls — one `err` span each — so a single failure produced exactly one ERROR event. Now every provenance failure is counted twice on any log-based error-rate dashboard / alert keyed on `cognee.db.relational.graph_storage.*`. Consider dropping `err` from the outer `upsert_provenance_graph` instrument (the inner spans already record it) or from the inner ones.

crates/database/tests/provenance_transaction.rs
```diff
@@ -0,0 +1,134 @@
...
+    let missing_dataset = Uuid::new_v4();
+    let poison_edge = make_edge(user, missing_dataset, data, nodes[0].id, nodes[1].id);
+
+    let result = upsert_provenance_graph(&db, &nodes, &[poison_edge]).await;
+    assert!(
+        result.is_err(),
+        "edge upsert against a missing dataset must fail"
+    );
+
+    let persisted = get_nodes_by_dataset(&db, dataset).await.expect("query");
```
GraDKh: **[test] Rollback test never asserts the edge side rolled back.**

This is the one test that guards transactional atomicity, but it only queries `get_nodes_by_dataset` and asserts nodes didn't persist — it never checks edges. Because the poison edge is the statement that *fails*, the test proves "nodes rolled back when a later edge fails" but says nothing about edge-side atomicity.

A regression that reintroduced per-batch autocommit (committing already-inserted edges before a later edge in the same batch fails) would still pass this test green. Cheap to close: seed one *valid* edge ahead of the poison edge and assert `get_edges_by_dataset(&db, dataset)` is empty after the rollback, so both sides of the group are covered.

crates/database/src/connection.rs
GraDKh: Re-checked on the rebased head (`a054d3a`): the `?mode=ro` / `immutable` URL case is fixed and gated correctly — thanks. But the **read-only mount / read-only file-permission** half of this comment is still open.

`read_only` is derived only from the URL query params in `classify_sqlite_url` (connection.rs:73), never from actual filesystem writability. So a plain `sqlite:///data/app.db` on a read-only mount (or an `r--` file) still classifies as `read_only: false`, enters the WAL branch at connection.rs:161, issues `PRAGMA journal_mode=WAL`, and fails the connect with `attempt to write a readonly database`.

Verified against vendored sqlx-sqlite 0.8.6: the pragma runs eagerly at connect (`options/connect.rs`) and the error propagates — there's no fallback — while pre-PR (`5af3034~1`) set no journal pragma at all, so the same URL opened read-only and served reads. Genuine regression for that case. Either fall back to no-WAL when the file isn't writable, or document that read-only opens require an explicit `mode=ro`.

crates/database/src/connection.rs
GraDKh: The URL-parsing hardening landed and `sqlite_url_is_in_memory` was exported as promised — but the *shared-predicate* half regressed on the #56 rebase, so this isn't fully closed.

`component_manager` no longer opens the DB; `cognee_components::build_database` does, and its `sqlite_fs_path` (crates/components/src/builtins/database.rs:29) still classifies in-memory with `url.contains(":memory:")`, which misses the `mode=memory` spelling. Net: `sqlite_url_is_in_memory` now has **zero non-test callers** (definition + unit tests + the lib.rs re-export only), so the two layers diverge again on exactly the case the helper was written to unify — and its doc comment still claims it "prevents the layers from diverging."

Reachable effect is mild (a spurious `create_dir_all` on the parent of an in-memory `mode=memory` URL, non-fatal), so low severity — but either route `sqlite_fs_path` through the shared predicate, or drop the helper and its doc claim.

crates/database/src/connection.rs
GraDKh: You went with option A (WAL), which does buy real reader/writer concurrency — good. One consequence of pairing that with the 10-connection pool is still open, though.

`upsert_provenance_graph` holds a single transaction (SQLite's exclusive writer lock) across *all* node batches plus *all* edge batches (graph_storage.rs:205-208), and file-backed SQLite is now a 10-connection pool where pre-PR sea-orm pinned it to 1 (confirmed: sea-orm 1.1.20 `driver/sqlx_sqlite.rs:85-86` → `if get_max_connections().is_none() { max_connections(1) }`). So two datasets cognifying the same file DB concurrently can have the second writer wait out `busy_timeout` (~5s sqlx default, not overridden) and fail with `SQLITE_BUSY` — the window this thread predicted.

Bounded in practice — the transaction body is only local batched inserts over pre-materialized `Vec`s, so realistic hold is sub-second — so it's acceptable if intended, but worth a conscious call (keep file SQLite at fewer connections, or accept the small BUSY window as the price of atomic provenance).

Also confirming the deferred-`BEGIN` sub-point from this thread is now correctly handled by the write-first argument in the doc at graph_storage.rs:186-188 — that part is resolved, and `BEGIN IMMEDIATE` wouldn't add anything since the first statement already takes the write lock.

crates/http-server/src/wiring.rs
GraDKh: Confirming the revert landed as we discussed — both call sites now go through `build_database` → `connect()` and the `DB_POOL_*` env wiring is gone. One loose end remains from this thread's resolution:

`PoolConfig` + `connect_with_pool` stayed `pub` but now have **zero callers that pass a non-default config** — only `connect()` (which passes `PoolConfig::default()`) and the lib.rs re-export reference them, and there are no `DB_POOL_*` env vars anywhere in the tree. So it's published-API surface with no exercised configurability. Either fold it back into `connect()`, or leave a `// TODO: not yet wired to config` so a future maintainer doesn't read it as a live knob.

(Minor precision on my earlier "only min_connections 0→1 changed": that holds for the Postgres path; for SQLite, `connect()` also newly applies WAL + `synchronous=NORMAL` and the reaper changes — but that's the intended perf work, independent of the unused `PoolConfig` seam.)

crates/database/tests/provenance_transaction.rs
GraDKh: The extraction to `upsert_provenance_graph` plus the FK-violation rollback test is a real improvement — a dropped `commit()` or an upsert moved outside the batch *inside that function* would now be caught. But the specific gap this thread raised (twice) was the call-site wiring in `upsert_provenance`, and that's still uncovered.

The new tests call `upsert_provenance_graph(&db, ...)` directly (provenance_transaction.rs:87, 122); they never go through cognify's `upsert_provenance` (tasks.rs:2305) or its guard at tasks.rs:2575 (`if !prov_nodes.is_empty() || !prov_edges.is_empty()`). The only thing that exercises the wrapper is `crates/cognify/tests/provenance_e2e.rs`, which is LLM-gated **and** treats any `cognify()` error as a skip (`Err(e) => { eprintln!(...); return; }`), so it can't fail on a wrapper regression either.

Low severity — the callee re-guards emptiness at graph_storage.rs:202, so a wrapper-guard mistake is a benign no-op, not corruption — but a deterministic cognify-level test that asserts on the relational `nodes`/`edges` tables would finally close this.

GraDKh (changes requested):
## Re-review after the rebase onto `main` (@a054d3a)

Thanks for the thorough round-3 work — most of the earlier threads are genuinely resolved (shared-cache reaper fix, statement-logging parity, no-op pragma cleanup, semver-safe concrete signatures, adapter-pool doc correction, WAL gated off read-only *URLs*, durability trade-offs documented). I re-ran the review against the rebased head and independently validated each remaining item against the source. Requesting one more iteration for the items below.

### Blocking

- **WAL still fails on a physically read-only DB** (connection.rs) — the `?mode=ro` case is fixed, but the read-only *mount / file-permission* case from that same thread isn't: `read_only` is inferred only from URL query params, not filesystem writability, so a plain URL on a read-only mount now fails to connect (`PRAGMA journal_mode=WAL` → `attempt to write a readonly database`) where it opened read-only before. Verified in sqlx-sqlite 0.8.6 (pragma runs eagerly, no fallback).
- **1→10 connection pool + long provenance transaction → `SQLITE_BUSY` window** (graph_storage.rs) — pre-PR sea-orm pinned file SQLite to 1 connection (`driver/sqlx_sqlite.rs:85-86`); now it's 10, and `upsert_provenance_graph` holds the writer lock across all node+edge batches, so concurrent same-file cognify can hit BUSY after `busy_timeout`. Bounded (local inserts, sub-second) — fine if it's a conscious call, but let's decide explicitly. (The deferred-`BEGIN` sub-point *is* resolved by the write-first doc — good.)

### Non-blocking (worth closing this iteration)

- **`sqlite_url_is_in_memory` regressed to a dead helper** — after the #56 move to `build_database`, its `sqlite_fs_path` still uses `.contains(":memory:")` (misses `mode=memory`); the exported predicate now has zero non-test callers, so the "one shared predicate" goal is undone. Route the components layer through it or drop the helper + doc claim.
- **`PoolConfig` / `connect_with_pool` are unused configurable API** — after the revert to `connect()`, nothing passes a non-default config and no `DB_POOL_*` env vars remain. Fold back into `connect()` or mark as not-yet-wired.
- **Provenance tests don't reach the cognify wrapper** — the new tests hit the DB seam (`upsert_provenance_graph`) directly, never `upsert_provenance`/its guard at tasks.rs:2575; the only wrapper coverage (`provenance_e2e.rs`) is LLM-gated and swallows `cognify()` errors. Low severity, but a deterministic cognify-level assertion would finally close the twice-raised thread.
- **Nested `#[instrument(err)]` double-counts ERROR events** (new) — `upsert_provenance_graph` and the inner `_on` fns all carry `err`, so one provenance failure emits two ERROR events; drop `err` from the inner fns (the outer one is the only coverage for begin/commit failures). Telemetry-only.
- **Rollback test never asserts the edge side** (new) — `provenance_graph_rolls_back_when_edge_upsert_fails` only checks nodes; seed a valid edge ahead of the poison edge and assert `get_edges_by_dataset` is empty. Low severity.

Details are in the inline comments and the reopened threads above. (One item I raised mid-review — extra `BEGIN`/`COMMIT` cost for one-sided batches — I retracted: under WAL+`synchronous=NORMAL` there's no extra fsync and it's a net win for >500-row batches.)

Nothing here is a correctness-of-data bug; the two blocking items are behavior regressions on the default file-backed path that are worth a deliberate decision before merge.

Taimoo-r:
Thanks for the round-3 re-review. All items from that pass are addressed in bdbc88f. Summary below, blocking first.

**Blocking**

1. **WAL on a physically read-only DB** (`connection.rs`). WAL is now gated on real filesystem writability, not the URL alone. `connect_sqlite` probes whether the target file is writable; when it is not (a read-only mount or `r--` permissions), the connection opens read-only and issues no journal-mode pragma. So a plain `sqlite:///data/app.db` on a read-only mount connects and serves reads as it did pre-PR. Covered by a new `read_only_file_permission_connects_without_wal` regression test, which skips where DAC cannot make the file unwritable (for example, root).

2. **Pool plus long provenance transaction reaching `SQLITE_BUSY`.** This is now a deliberate, documented choice. `busy_timeout` had been dropped in the rebase, so I restored `.busy_timeout(5s)`: a concurrent writer on the same file waits for the lock instead of failing immediately. The single transaction across all node and edge batches is intentionally kept, since it is the atomicity guarantee this PR adds, and it is documented on `upsert_provenance_graph`: under WAL readers are never blocked, writers serialize under the 5s timeout, and the batches are pre-built local inserts that commit well within that window.

**Non-blocking**

3. **`sqlite_url_is_in_memory` dead helper.** `build_database`'s `sqlite_fs_path` now routes through the shared `sqlite_url_is_in_memory` predicate instead of its own `.contains(":memory:")`, so the `mode=memory` spelling is handled and the one-predicate goal holds again.

4. **`PoolConfig` / `connect_with_pool` unused API.** Folded back into `connect()`. Both are removed and the pool sizing lives in internal constants, so no unused configurable surface remains.

5. **Tests not reaching the cognify wrapper.** Added a deterministic test that drives `upsert_provenance` and its guard on in-memory SQLite: one document produces exactly one provenance node, with no LLM, so a regression in the wrapper's own wiring now fails a green test.

6. **Nested `#[instrument(err)]` double-counting.** The `err`-recording span moved off the generic `_on` impls onto the public `upsert_nodes` and `upsert_edges`, and stays on `upsert_provenance_graph`. One failure now records exactly one ERROR event on both the direct-upsert and provenance paths, and begin/commit failures stay covered by the outer span.

7. **Rollback test not asserting the edge side.** `provenance_graph_rolls_back_when_edge_upsert_fails` now seeds a valid edge ahead of the poison edge and asserts `get_edges_by_dataset` is empty after the rollback, so both sides of the group are covered.

Verified with `cargo clippy -D warnings` and the test suites for `cognee-database`, `cognee-components`, and `cognee-cognify` on Linux.

Taimoo-r:
Hi @GraDKh, everything is addressed now. Please review it at your earliest convenience.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,293 @@
...
+        .await
+        .map_err(|e| DatabaseError::ConnectionError(e.to_string()))?;
+
+    Ok(SqlxSqliteConnector::from_sqlx_sqlite_pool(sqlx_pool))
+}
+
+/// Best-effort filesystem path for a file-backed SQLite URL; `None` for
+/// in-memory URLs or when no path is discernible.
+#[cfg(feature = "sqlite")]
+fn sqlite_fs_path(url: &str) -> Option<String> {
```
GraDKh: **Correctness bug: the writability probe doesn't percent-decode, but sqlx does.**

`sqlite_fs_path` splits the URL by hand and never percent-decodes, whereas sqlx's `SqliteConnectOptions::from_str` *does* decode the filename (sqlx-sqlite `options/parse.rs`). So for `sqlite:///data/my%20app.db` pointing at an existing **read-only** file, this probe checks the literal path `.../my%20app.db` — which doesn't exist — so `sqlite_path_is_writable` returns `true`, the `journal_mode=WAL` pragma is issued on a read-only file, and the connect fails with *"attempt to write a readonly database"*. That's precisely the failure the probe exists to prevent, and it only bites on paths with URL-escaped characters (spaces, etc.).

The fix **deletes code**: `sqlite_path_is_writable` already runs *after* `SqliteConnectOptions::from_str(url)` has succeeded a few lines above, so pass the already-decoded path from `conn_opts.get_filename()` instead of re-parsing the raw URL, and drop `sqlite_fs_path` entirely. (This also collapses the divergent-duplication concern with the sibling `components::builtins::database::sqlite_fs_path`.)

crates/database/src/connection.rs
```diff
@@ -1,19 +1,293 @@
...
+    let kind = classify_sqlite_url(url);
+
+    // Statement logging at INFO matches sea-orm's `ConnectOptions` default,
+    // which the Postgres path still goes through; raw sqlx defaults to DEBUG.
+    // `busy_timeout` lets a writer wait for the lock (WAL still serializes
+    // writers) instead of erroring out immediately with `SQLITE_BUSY`.
+    let mut conn_opts = SqliteConnectOptions::from_str(url)
+        .map_err(|e| DatabaseError::ConnectionError(e.to_string()))?
+        .log_statements(log::LevelFilter::Info)
+        .busy_timeout(Duration::from_secs(5));
```
GraDKh: Two points on `busy_timeout(5s)`:

1. **5s just restates sqlx's default** (`SqliteConnectOptions` defaults `busy_timeout` to 5s), so this line currently changes nothing.
2. The Python reference (`SqlAlchemyAdapter`, the #2717 fix) uses **`busy_timeout=120000` (120s)**. It matters here because `upsert_provenance_graph` now holds SQLite's single writer lock across the *entire* node+edge batch group; on a slow/edge device a large group can exceed 5s while a second writer/process is waiting, surfacing as `SQLITE_BUSY`. Bumping to 120s matches Python and closes that window (or, if you'd rather rely on the default, drop the call and add a comment saying so).

crates/database/src/connection.rs
```diff
@@ -1,19 +1,293 @@
...
+/// Relational connection-pool sizing, applied by [`connect`].
+///
+/// These pin sqlx's own pool defaults, except `POOL_MIN_CONNECTIONS = 1` (sqlx
+/// defaults to 0) so one warm connection survives idle periods. The pool serves
+/// only the relational database: the Postgres graph and vector adapters
+/// (`PgGraphAdapter`, `PgVectorAdapter`) open their own separate pools.
+/// `POOL_ACQUIRE_TIMEOUT` surfaces pool exhaustion as a prompt error instead of
+/// a silent hang. In-memory SQLite overrides parts of this at connect time
+/// because correctness requires it (see [`connect_sqlite`]).
+const POOL_MAX_CONNECTIONS: u32 = 10;
```
GraDKh, line 24: `POOL_MAX_CONNECTIONS = 10` is arbitrary — no benchmark motivates it, and for an embedded single-writer SQLite DB the idiomatic shape is *1 writer + N readers in separate pools* (which neither sqlx nor sea-orm's single `DatabaseConnection` can express here, so a modest shared pool is the practical ceiling — that's fine). Not a blocker, but please either document why 10 or reduce it to a considered number.

Related (line 19): `POOL_MIN_CONNECTIONS = 1` on **file-backed** DBs keeps a warm connection — and the `-wal`/`-shm` sidecars — alive for the whole process lifetime with no correctness need. The "never drop to zero connections" invariant only matters for **in-memory** DBs (already handled on that branch). Consider setting min=0 for file DBs and keeping min=1 only on the in-memory path.

crates/database/src/connection.rs
```diff
@@ -1,19 +1,293 @@
...
+        // In-memory: no file to journal; sqlx's default WAL is a no-op here.
+    } else if kind.read_only || !sqlite_path_is_writable(url) {
+        // A read-only URL, or a file we cannot write (read-only mount or
+        // permissions): open read-only so sqlx issues no journal-mode pragma.
+        // `PRAGMA journal_mode=WAL` writes to the file and would otherwise fail
+        // the connect, where before it served reads.
+        conn_opts = conn_opts.read_only(true);
+    } else {
+        conn_opts = conn_opts
+            .journal_mode(SqliteJournalMode::Wal)
```
GraDKh: No change required here — but this WAL + `synchronous=NORMAL` default deserves an explicit note in the PR body, because vs. the *previous Rust* behavior it's a real change (prior code issued no journal/synchronous pragma → SQLite defaults DELETE + FULL):

- `NORMAL` under WAL can lose the last committed transactions on **power loss** (safe on process crash) — was effectively FULL before.
- WAL is persisted in the DB header, requires shared-memory + proper file locking (**breaks on some NFS/SMB / container bind-mounts**), and creates `-wal`/`-shm` sidecars so a single-file `cp`/backup taken while running is incomplete.

These match Python's behavior, so keeping them is the right call for parity. Ideally gate WAL behind a config opt-out for network-filesystem deployments — a good shared follow-up for both SDKs.

crates/database/src/lib.rs
```diff
@@ -16,28 +16,36 @@ pub use ops::tutorial_seeder::{
...
+pub fn database_system_label(db: &DatabaseConnection) -> &'static str {
+    connection_system_label(db)
 }
 pub use ops::checkpoint::{CheckpointStore, SeaOrmCheckpointStore};
 pub use pipelines::sea_orm_impl::SeaOrmPipelineRunRepository;
 pub use pipelines::{
     NoopPipelineRunRepository, PipelineRunRepository, PipelineRunWithAttributionRow,
 };
-pub use sea_orm::DatabaseConnection;
+pub use sea_orm::{DatabaseConnection, TransactionTrait};
```
GraDKh: Minor / API-surface: re-exporting `TransactionTrait` at the crate root widens the published `cognee-database` API, and no internal caller needs it there — `graph_storage.rs` imports `TransactionTrait` directly from `sea_orm`. It's defensible *if* you intend external callers to do `db.begin()` without pulling a version-matched `sea-orm` dep; if that's not a goal, drop it, since it can't be removed later without a semver-breaking release.

GraDKh (changes requested):
Thanks — the core of this PR is solid and I verified most of it against the vendored sea-orm 1.1.20 / sqlx 0.8.6 sources and the Python reference. A few things worth stating up front so the review reads fairly:

- **The architecture is right.** Building the pool via `SqliteConnectOptions` + `SqlxSqliteConnector::from_sqlx_sqlite_pool` is the correct approach — sea-orm 1.1.20's `ConnectOptions` genuinely cannot disable sqlx's pool reapers (`idle_timeout`/`max_lifetime`), so the direct-sqlx path is *forced*, not gratuitous. The in-memory reaper fix is a real latent-bug fix (an idle in-memory connection was being reaped → silently reconnecting to a fresh empty DB).
- **WAL + `synchronous=NORMAL` is a faithful port** of Python's `SqlAlchemyAdapter` (same pragmas, added for their "database is locked" issue #2717), so I'm not asking to remove it.
- The transaction wrapper (write-first deferred BEGIN) and the `connection_pool.rs` / `provenance_transaction.rs` regression tests are good and pin exactly the right invariants.

Requesting changes for **one real correctness bug** (the percent-decoding gap in the writability probe) plus a few hygiene items called out inline. Two non-inline items:

1. **Stale PR description.** The body still says "SQLite is pinned to a single connection on purpose" — that was round-1 behavior; the head now uses a pool of 10 for file DBs. Please refresh so reviewers/history aren't misled.
2. **Test churn.** The `&db` → `db.as_ref()` edits in `memify_sync_graph_session.rs` and `document_node_hard_delete.rs` look unnecessary now that the public signatures stayed concrete (`&DatabaseConnection`) — revert to keep the diff minimal, unless there's a reason they're needed.

Taimoo-r:
Heyyy @GraDKh, I've settled all the things you've mentioned. Review it at your earliest convenience!!

crates/database/src/connection.rs
GraDKh: The bump to 120s landed correctly ✅ — one leftover from this thread: the `upsert_provenance_graph` doc (`graph_storage.rs:200`) still says a concurrent writer "waits out the **5s** `busy_timeout`". Flagged inline; it just needs to read 120s to match `SQLITE_BUSY_TIMEOUT`.

crates/database/src/connection.rs
GraDKh: The percent-decode fix (`get_filename`) resolves the escaped-path case ✅. One gap remains in the same probe: `sqlite_path_is_writable` checks only the file, not its parent directory. An *existing, writable* DB file inside a read-only directory (`chmod 555`) makes the probe return `true`, so WAL is force-enabled and `PRAGMA journal_mode=WAL` fails the connect creating the `-wal`/`-shm` sidecars (reproduced on SQLite 3.44). Pre-PR that file opened read-only and served reads. The doc at lines 210-213 ("if its parent is unwritable the connect fails the same way with or without WAL") holds for a *missing* file via `mode=rwc`, but not for an existing readable one. Either probe parent-dir writability too, or narrow the doc claim.

crates/database/src/ops/graph_storage.rs
```diff
@@ -152,6 +184,45 @@ pub async fn upsert_edges(
...
+/// provenance graph never ends up half-written.
+///
+/// `begin()` issues a deferred `BEGIN`, but this transaction is write-first:
+/// the first statement is an upsert, which takes SQLite's write lock
+/// immediately, so there is no read-to-write lock upgrade to deadlock on.
+///
+/// On SQLite this holds the single writer lock for the whole group (all node
+/// batches, then all edge batches) — a deliberate trade for atomicity. Under
+/// WAL, readers are never blocked; a concurrent writer on the same file waits
+/// out the 5s `busy_timeout` (see `connect_sqlite`) rather than failing with
```
GraDKh: Stale after the round-5 `busy_timeout` bump: this says the writer "waits out the **5s** `busy_timeout`", but `connect_sqlite` now sets `SQLITE_BUSY_TIMEOUT = 120s` (`connection.rs:36`). Please update to 120s — an operator sizing an upstream timeout off "5s" would abort a provenance write mid-flight under real contention.

GraDKh (changes requested):
Thanks — the round-5 items (decoded-path probe, 120s `busy_timeout`, file-pool `min=0`, dropped `TransactionTrait`) all landed cleanly; requesting one more short pass on the two leftovers noted inline and in the reopened threads.

GraDKh:
Two non-code follow-ups before merge:
1. The PR description still says "SQLite is pinned to a single connection on purpose" — that was round-1 behavior; the head now pools file DBs (WAL, max 10). Worth refreshing so the description/history isn't misleading.
2. The branch is currently conflicting with `main` and will need a rebase before merge.

Taimoo-r:
Hi @GraDKh, thanks for the round-5 pass. Both items are addressed, and the branch is caught up with `main`.

**Round-5 items** (d946b2c)

1. **Parent-directory writability.** `sqlite_path_is_writable` now probes the parent directory as well, by attempting to create a uniquely named temp file next to the database: WAL has to create the `-wal`/`-shm` sidecars there, so an existing writable file inside a read-only directory (`chmod 555`) now opens read-only instead of failing the connect. Covered by a new `cfg(unix)` regression test (`writable_file_in_read_only_dir_connects_without_wal`), which restores and skips where DAC cannot enforce the read-only directory (for example, root). The probe doc no longer overclaims the missing-file case.
2. **Stale doc.** The `upsert_provenance_graph` doc now reads 120s, matching `SQLITE_BUSY_TIMEOUT`.

**Merge with `main`** (88392f7). Resolved the two conflicts from the base advancing:

- `graph_storage.rs`: kept your `dedup_keeping_last` (the ON CONFLICT affects-row-twice fix) alongside the public-entry-point `err` spans; the connection-generic `_on` impls carry the dedup calls, so the transactional provenance path gets it too.
- `integration_text_summary_payload.rs`: both sides had converged on the same `{"nodes": [], "edges": []}` payload; kept `main`'s comment explaining the required `edges` field.

The PR description now also lists the read-only parent-directory case. Lint (fmt + check + clippy) is green on the merge head.

GraDKh:
Thanks for this, @Taimoo-r — really solid work. 🙏

Wrapping the bulk provenance writes in a single transaction plus the explicit connection-pool configuration is exactly the right direction, and the accompanying `busy_timeout` / write-first-transaction handling and the new `provenance_transaction`/`connection_pool` tests show real care. We ran a deep multi-agent review over the diff and the core change held up well under scrutiny.

Merging now. A few follow-up items surfaced around the new `connect_sqlite` path (WAL on network filesystems, the silent read-only downgrade, and a couple of smaller cleanups) plus some design nuances worth revisiting — I'll open a separate follow-up PR for those so they don't hold this one up. Thanks again!

Taimoo-r:
Thanks @GraDKh, this was a great experience working with you. I learned a lot along the way, especially how changes take shape at runtime as requirements keep evolving, which is really at the core of open source. Appreciate the thorough and patient reviews throughout.

PR #38, opened 2026-06-29
feat(ios): Swift async/await SDK + xcframework packaging (fixes #3512)

yonkosan:
## What this adds

Cross-compiles `cognee-capi` for `aarch64-apple-ios` and `aarch64-apple-ios-sim`, packages the output as `CogneeSDK.xcframework`, and wraps the C API in an idiomatic Swift package.

### Files changed

| Path | What |
|---|---|
| `capi/cognee-capi/build.rs` | Emit iOS linker flags + SDKROOT via xcrun |
| `capi/cognee-capi/Cargo.toml` | Gate `crate-type` to `staticlib` on iOS; add `mock-llm` feature |
| `capi/scripts/build_xcframework.sh` | Build script — compiles both slices, runs `xcodebuild -create-xcframework` |
| `ios/Package.swift` | Swift package manifest with binary + wrapper + test targets |
| `ios/Sources/CogneeSDK/Cognee.swift` | Public API — `warm`, `add`, `cognify`, `search`, `recall`, `configure*` |
| `ios/Sources/CogneeSDK/CogneeError.swift` | Maps `CgErrorCode` → Swift `Error` |
| `ios/Sources/CogneeSDK/ContinuationBox.swift` | Heap wrapper bridging continuations across the C boundary |
| `ios/Tests/CogneeSDKTests/PipelineTests.swift` | XCTest: full offline pipeline using cassette |
| `ios/Tests/CogneeSDKTests/Fixtures/demo_cassette.json` | LLM cassette recorded against llama3.2:3b via Ollama |
| `.github/workflows/ios.yml` | CI: `cargo check` for both iOS targets + Swift parse |
| `ios/README.md` | Build instructions, usage, test run, architecture |

### Test results (iPhone 17 Simulator)

```
Test Case 'testConfigureMockMode'     passed (0.031 seconds)
Test Case 'testSDKInit'               passed (0.001 seconds)
Test Case 'testWarmAddCognifySearch'  passed (0.172 seconds)
Test Suite 'All tests' passed
```

The pipeline runs fully offline — no network, no API key, no running model. The cassette replays real `llama3.2:3b` responses recorded once via Ollama.

### CI note

The xcframework (6.6 GB, gitignored) is not rebuilt in CI — GitHub's macOS runners have ~14 GB disk. The workflow uses `cargo check` which type-checks both iOS targets without linking and catches broken C API signatures on every PR.

### How to verify locally

```bash
./capi/scripts/build_xcframework.sh
cd ios
xcodebuild test \
  -scheme CogneeSDK \
  -destination 'platform=iOS Simulator,name=iPhone 17' \
  2>&1 | grep -E 'Test Suite|Test Case|PASS|FAIL'
```


.github/workflows/ios.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Validate Package.swift
+        run: |
+          cd ios
+          swift package dump-package > /dev/null
+          echo "Package.swift is valid"
+
+      # Parse-check Swift source files for syntax errors without linking.
+      # The xcframework binary is not present in CI (6.6 GB, gitignored), so
+      # we use `swiftc -parse` rather than `swift build`.
+      - name: Parse-check Swift sources
```
GraDKh: **CI doesn't type-check the Swift↔C boundary, and the tests never run in CI**

`swiftc -parse` only checks *syntax* — it never resolves `import CogneeSDKCore`, so the failure mode that actually matters for a hand-written C bridge slips through: a renamed `cg_sdk_*` function, a changed argument count, or a changed `CgErrorCode` representation all pass CI and only fail at local link time. (The README's claim that CI "catches broken C API signatures" isn't true with `-parse`.)

We can close this cheaply without the 6.6 GB xcframework. `cognee_sdk.h` is committed and self-contained (it `#include`s `cognee.h`, which pulls only standard headers), so we can synthesise a Clang module from `capi/include` and `swiftc -typecheck` the wrapper against the *real* C API. Type-checking a library's sources needs only the headers, not the linked symbols:

```diff
-      # Parse-check Swift source files for syntax errors without linking.
-      # The xcframework binary is not present in CI (6.6 GB, gitignored), so
-      # we use `swiftc -parse` rather than `swift build`.
-      - name: Parse-check Swift sources
+      - name: Type-check Swift wrapper against the C header
         run: |
-          swiftc -parse \
+          cat > capi/include/module.modulemap <<'MODULE'
+          module CogneeSDKCore {
+              umbrella header "cognee_sdk.h"
+              export *
+          }
+          MODULE
+          swiftc -typecheck \
+            -sdk "$(xcrun --sdk iphonesimulator --show-sdk-path)" \
+            -target arm64-apple-ios13.0-simulator \
+            -Xcc -fmodule-map-file="$PWD/capi/include/module.modulemap" \
+            -Xcc -I"$PWD/capi/include" \
             ios/Sources/CogneeSDK/ContinuationBox.swift \
             ios/Sources/CogneeSDK/CogneeError.swift \
-            ios/Sources/CogneeSDK/Cognee.swift \
+            ios/Sources/CogneeSDK/Cognee.swift
+
+      - name: Parse-check Swift tests
+        run: |
+          swiftc -parse \
             ios/Tests/CogneeSDKTests/PipelineTests.swift
```

The test target stays on `-parse` because `@testable import CogneeSDK` can't be type-checked without first building the module (which needs the xcframework).

Two things worth a quick decision:
1. **`swift build` is genuinely not an option here** — it would resolve the `CogneeSDKCore` binaryTarget, but the xcframework isn't present on the runner (and the package is iOS-only, so a host build has no matching slice). The `-typecheck` route above is the cheap equivalent.
2. **The XCTest suite never runs in any automated job**, so behavioral regressions in the bridge aren't caught anywhere. Is manual `xcodebuild test` acceptable, or do we want a nightly job that builds the xcframework and runs the suite?

I have the workflow + README change ready to push if we agree on the approach.

ios/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+# Record (from repo root)
+cargo run -p cognee-cli --no-default-features \
+  --features sqlite,testing,mock-llm -- \
+  --llm-provider openai \
+  --llm-base-url http://localhost:11434/v1 \
+  --llm-model llama3.2:3b \
+  --llm-api-key ollama \
+  --cassette-out ios/Resources/demo_cassette.json \
+  bench add ios/Resources/memories.json
```
GraDKh: **"Record a new cassette" writes to a path the tests never read**

This command records to `ios/Resources/demo_cassette.json`, but the fixture the test target actually loads (declared in `Package.swift` as a resource and read via `Bundle.module` in `PipelineTests`) is `ios/Tests/CogneeSDKTests/Fixtures/demo_cassette.json` — which the README itself documents a few lines up ("The cassette is committed at `Tests/CogneeSDKTests/Fixtures/demo_cassette.json`").

So following these instructions regenerates the wrong file: the new cassette lands in `Resources/`, while the test keeps replaying the stale `Fixtures/` copy. The failure is silent — the suite still "passes" against old data.

Suggested fix:
```diff
-  --cassette-out ios/Resources/demo_cassette.json \
+  --cassette-out ios/Tests/CogneeSDKTests/Fixtures/demo_cassette.json \
```

This ties into #3 (the cassette is currently committed in both locations) — if we drop the `ios/Resources/` copy, this command becomes the single source of truth that writes straight to the fixture the tests use.

ios/Resources/demo_cassette.json
```diff
@@ -0,0 +1,92 @@
+{
```
GraDKh: **Duplicate cassette + orphaned `ios/Resources/` directory**

This file is byte-identical (blob `c8b9db9`) to `ios/Tests/CogneeSDKTests/Fixtures/demo_cassette.json`, so the same 92-line cassette is committed twice. Neither `ios/Resources/demo_cassette.json` nor `ios/Resources/memories.json` is referenced by `Package.swift` — only the README's "Record a new cassette" snippet points at this directory (see #2). That leaves `ios/Resources/` as a docs-only copy that can silently drift from the real fixture the tests load.

Suggested cleanup:
- Delete `ios/Resources/` (both `demo_cassette.json` and `memories.json`).
- Point the recording command at `ios/Tests/CogneeSDKTests/Fixtures/demo_cassette.json` (the #2 fix), making the fixture the single source of truth.

If `memories.json` is still useful as the documented recording input, it could move next to the fixture (e.g. `Tests/CogneeSDKTests/Fixtures/`) so everything cassette-related lives in one place, rather than in a separate top-level folder that nothing builds against.

capi/cognee-capi/build.rs
```diff
@@ -2,4 +2,31 @@ fn main() {
...
+            } else {
+                "iphoneos"
+            };
+            if let Ok(output) = std::process::Command::new("xcrun")
+                .args(["--show-sdk-path", "--sdk", sdk])
+                .output()
+            {
+                if output.status.success() {
+                    let path = String::from_utf8_lossy(&output.stdout).trim().to_string();
+                    println!("cargo:rustc-env=SDKROOT={path}");
```
GraDKh: **SDKROOT fallback is ineffective — `cargo:rustc-env` doesn't reach cc-rs**

The intent here (comment above: "Help cc-rs find the iOS SDK") doesn't work. `cargo:rustc-env=SDKROOT=…` only sets a compile-time variable for *this crate's* `rustc` invocation (available to `env!`/`option_env!`); it does not enter the OS process environment. The build scripts that actually call cc-rs/clang — `lz4-sys`, `zstd-sys`, `blake3` — run as separate processes and read `SDKROOT` from their own `std::env`, which this `println!` never touches.

So this branch is effectively dead: the iOS build only links today because clang/xcrun auto-detects the SDK on its own. The code reads as if it's doing something load-bearing when it isn't.

Two options:
- **Drop the SDKROOT block** and rely on clang's auto-detection (simplest — matches what actually happens now).
- **Set it in the process environment** instead, from `build_xcframework.sh` (e.g. `export SDKROOT=$(xcrun --sdk … --show-sdk-path)` before `cargo build`), so every `-sys` build script in the graph inherits it. That's the only place a single SDKROOT export reliably reaches cc-rs.

(The `rustc-link-lib=framework=Foundation` line just above has a related issue — I'll note that separately.)

capi/cognee-capi/build.rs
```diff
@@ -2,4 +2,31 @@ fn main() {
     // Header is maintained manually at capi/include/cognee.h
     // cbindgen was useful for bootstrapping but the manual header
     // gives better control over the C API surface.
-}
+
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    if target_os == "ios" {
+        // iOS requires the Foundation framework at link time.
+        println!("cargo:rustc-link-lib=framework=Foundation");
```
GraDKh: **`rustc-link-lib=framework=Foundation` is a no-op for the iOS staticlib — propose removing it**

On iOS the crate builds as `staticlib` only (per the new `[target.'cfg(target_os = "ios")'.lib]` override in `Cargo.toml`). A static library has no link step, so `cargo:rustc-link-lib=framework=Foundation` doesn't actually link anything here — the framework gets resolved when Xcode links the final app, and Xcode links Foundation by default regardless.

Combined with the SDKROOT issue in #4 (the other half of this `if target_os == "ios"` block is also ineffective), the entire iOS branch in `build.rs` is dead. I'd propose removing it:

```diff
 fn main() {
     // Header is maintained manually at capi/include/cognee.h
     // cbindgen was useful for bootstrapping but the manual header
     // gives better control over the C API surface.
-
-    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
-    if target_os == "ios" {
-        // iOS requires the Foundation framework at link time.
-        println!("cargo:rustc-link-lib=framework=Foundation");
-
-        // Help cc-rs find the iOS SDK if SDKROOT is not already set.
-        if std::env::var("SDKROOT").is_err() {
-            ...
-        }
-    }
 }
```

If we'd rather keep an explicit Foundation dependency for non-static consumers, the better home is the linker flags in `build_xcframework.sh` / the Swift package, not a `build.rs` directive that the staticlib build silently discards. Either way the current block gives a false impression that it's doing load-bearing work.

crates/lib/src/component_manager.rs
```diff
@@ -195,6 +195,9 @@ impl ComponentManager {
                 "graph_database_provider=postgres requires the `pggraph` crate feature".into(),
             )),
 
+            #[cfg(feature = "testing")]
+            "mock" => Ok(Arc::new(cognee_graph::MockGraphDB::new())),
+
             other => Err(ComponentError::Config(format!(
                 "Unsupported graph_database_provider '{other}'. Supported: ladybug, kuzu, postgres.",
```
GraDKh: **Error message omits the new `mock` provider**

The new `#[cfg(feature = "testing")] "mock" => …` arm just above adds `mock` as a valid `graph_database_provider`, but this fallback message still lists only `ladybug, kuzu, postgres`. It's also inconsistent with the vector-store message a bit further down, which already advertises `mock (testing feature only)`. A user who fat-fingers the provider under a `testing` build won't be told `mock` exists.

```suggestion
                "Unsupported graph_database_provider '{other}'. Supported: ladybug, kuzu, postgres, mock (testing feature only).",
```

GraDKh (changes requested):
Thanks for this — really solid piece of work. The async/await bridge over the C SDK is clean (the `passRetained`/`takeRetainedValue` single-shot pattern is exactly right), the offline cassette test is a great touch, and the disk-budget reasoning behind the CI design is well thought through. The Rust-side changes (CLI `default-features = false`, the `mock-llm` feature, the graph `mock` provider) are sound and behavior-preserving for the default build.

I've left a handful of inline comments. The main one worth resolving before merge:

- **CI doesn't type-check the Swift↔C boundary** — `swiftc -parse` is syntax-only and never resolves `import CogneeSDKCore`, so C-API drift (renamed `cg_sdk_*`, changed arg counts, `CgErrorCode` representation) passes CI. There's a cheap `swiftc -typecheck` fix that needs no xcframework; I have it ready to push. Also worth a decision: the XCTest suite currently runs in no automated job.

Smaller items:
- "Record a new cassette" writes to `ios/Resources/` but the tests load `Tests/.../Fixtures/` — regenerating silently misses the fixture.
- That cassette is committed twice; `ios/Resources/` is referenced by nothing but the README.
- The `if target_os == "ios"` block in `build.rs` is effectively dead (the `SDKROOT` `rustc-env` never reaches cc-rs, and the `Foundation` link is discarded in a staticlib build) — proposed removal inline.
- Graph-provider error message doesn't list the new `mock` option (one-line suggestion inline).

Happy to push the CI/README fixes once we agree on the approach. None of this is structural — mostly tightening up CI coverage and docs/dead-code cleanup.

yonkosan:
Good catches, all valid. Will fix everything in one follow-up commit.

Two quick decisions before I push:

**`-typecheck`**: Switching to your YAML. On the XCTest question — running it manually is fine for now. The typecheck catches the structural drift (wrong symbol name, arg count, enum representation) which is what CI actually needs to enforce; behavioral correctness is covered by the cassette tests locally. Full XCTest automation needs a self-hosted runner with ~15 GB spare to build the xcframework, which is a separate conversation.

**`build.rs`**: Removing the iOS block from `build.rs` and adding the `SDKROOT` export to `build_xcframework.sh` instead. Clang auto-detects fine today without it, but the script is the right place — env set there propagates to every cc-rs build script in the dep tree. A dead `rustc-env` that never reaches those subprocesses isn't worth keeping around.

`ios/Resources/` cleanup, cassette dedup, and the error string — fixing those directly.

yonkosan:
@GraDKh all five points addressed in a86acc7 (rebased onto main to resolve the Cargo.toml  conflict). Ready for re-review when you get a chance.

capi/cognee-capi/Cargo.toml
```diff
@@ -18,6 +18,11 @@ publish = false
 [lib]
 crate-type = ["cdylib", "staticlib"]
 
+# iOS forbids dynamic libraries in App Store submissions.
+# Override to staticlib-only when cross-compiling for iOS.
+[target.'cfg(target_os = "ios")'.lib]
```
GraDKh: **Blocking:** `[target.'cfg(target_os = "ios")'.lib]` is not a supported Cargo manifest table — target-specific tables only accept dependency sections, so Cargo silently ignores this and emits `warning: unused manifest key: target.cfg(target_os = "ios").lib` on every build, on all platforms. The crate-type stays `["cdylib", "staticlib"]` for iOS, so the staticlib-only invariant the comment claims (App Store compliance) is never established, and `build_xcframework.sh --release` still builds and links the multi-GB cdylib slice for `aarch64-apple-ios`.

Options: build with `cargo rustc --crate-type staticlib` in the script, or accept the cdylib byproduct and drop this section + fix the comment.

ios/README.md
```diff
@@ -0,0 +1,145 @@
...
+The committed cassette was recorded against `llama3.2:3b` via Ollama. To record a fresh one with different content:
+
+```bash
+# Start Ollama
+ollama serve &
+ollama pull llama3.2:3b
+
+# Record (from repo root)
+cargo run -p cognee-cli --no-default-features \
+  --features sqlite,testing,mock-llm -- \
```
GraDKh: **Blocking:** this recording command cannot run:

- `cognee-cli` has no `--cassette-out`, `--llm-base-url`, or `--llm-api-key` flags (recording is driven by the `llm_record_path` setting / `COGNEE_RECORD_LLM` env var);
- `bench` takes `--memories <file>` / `--mock-memories`, not a positional `add <file>`;
- `--no-default-features` disables the CLI's `bench` feature, so the subcommand doesn't exist in this build at all.

As written, `demo_cassette.json` — the fixture the whole iOS test suite depends on — has no working regeneration path, so it will rot unreproducibly on the first prompt/schema change. Please replace this with a recipe that actually runs (and consider wiring the iOS cassette into the existing `record-cassettes.yml` auto-refresh workflow so it's regenerated alongside the crate cassettes).

crates/cli/Cargo.toml
```diff
@@ -86,7 +87,7 @@ android-default = ["cognee-lib/android-default"]
 
 [dependencies]
 clap = { workspace = true, features = ["derive"] }
-cognee-lib = { path = "../lib", version = "0.1.3" }
+cognee-lib = { path = "../lib", version = "0.1.3", default-features = false }
```
GraDKh: The iOS SDK links `cognee-capi`, not the CLI — as far as I can tell the only reason `crates/cli` is touched is to support the README's cassette-recording recipe, which currently doesn't run (see the README comment). Meanwhile this line has a real side effect: with `default-features = false` on the dependency edge, every non-default CLI build changes. Notably `scripts/android-build-and-deploy.sh` (`--no-default-features --features cognee-cli/android-default`) previously still compiled cognee-lib with its full default set (pdfium, postgres, pgvector, tiktoken, …) and now gets only what `android-default` forwards — a feature combination that has never been compiled in CI.

Please drop the `crates/cli` changes from this PR (both this line and the new `testing` feature), or split them into a separate PR together with a working recording recipe and a `cargo check -p cognee-cli --no-default-features --features android-default` verification.

.github/workflows/ios.yml
```diff
@@ -0,0 +1,101 @@
...
+          swift package dump-package > /dev/null
+          echo "Package.swift is valid"
+
+      # Type-check the Swift wrapper against the real C header — resolves
+      # `import CogneeSDKCore` via a synthesised Clang module so that renamed
+      # or removed `cg_sdk_*` symbols, changed argument counts, and changed
+      # `CgErrorCode` values are caught without the xcframework binary.
+      - name: Type-check Swift wrapper against the C header
+        run: |
+          cat > capi/include/module.modulemap <<'MODULE'
```
GraDKh: The `CogneeSDKCore` module.modulemap is generated by two independent heredocs — this one and the copy in `capi/scripts/build_xcframework.sh`. If the module name or umbrella header ever changes in one but not the other, CI type-checks the Swift wrapper against a different Clang module than the one shipped inside the xcframework: CI stays green while `import CogneeSDKCore` breaks for real consumers (or vice versa). This step also writes into the tracked `capi/include/` directory at CI time, which the capi header-sync tooling watches.

Suggestion: commit `capi/include/module.modulemap` once as a real file; this step then needs no `cat`, and the build script replaces its heredoc with a `cp`.

ios/Tests/CogneeSDKTests/PipelineTests.swift
```diff
@@ -0,0 +1,109 @@
...
+        // ── 1. Create SDK and configure for offline mock mode ─────────────
+        let cognee = try Cognee()
+        try cognee.configureMockMode(cassettePath: path)
+
+        // ── 2. Warm (builds in-memory stores using mock config) ────────────
+        try await cognee.warm()
+
+        // ── 3. Add — text must match what was used during cassette recording
+        //     (memory_to_text format: "Title: …\n\n<content>\n\nReferences: none")
+        let memoryText = """
```
GraDKh: This hardcoded string must match the cassette's `sha256(user_input + schema)` keys byte-for-byte, and the same prose now lives in three places: this literal, `Fixtures/memories.json`, and the cassette's recorded input (including the CLI-internal `memory_to_text` shaping `Title: …\n\n<content>\n\nReferences: none`). Anyone who edits `memories.json` and re-records — or introduces an invisible one-space divergence via the `\` line continuations — gets a cassette miss that surfaces as an unrelated-looking mock-LLM failure.

Suggestion: load the text from the bundled `memories.json` and apply the same shaping in the test, so there's one source of truth.

ios/Sources/CogneeSDK/Cognee.swift
```diff
@@ -0,0 +1,310 @@
...
+                }
+                box.continuation.resume(returning: ())
+            }, ptr)
+        }
+    }
+
+    /// Return the owner UUID as a quoted JSON string (e.g. `"\"abc-…\""`).
+    ///
+    /// Warms the handle lazily if services have not yet been built.
+    public func ownerId() async throws -> String {
```
GraDKh, line 111: Six methods (`ownerId`, `add`, `cognify`, `addAndCognify`, `search`, `recall`) repeat a byte-identical ~15-line continuation/callback block; only the `cg_sdk_*` invocation differs. Since the `@convention(c)` callback is non-capturing and identical everywhere, it can be a single shared file-private callback constant plus one generic helper:

```swift
private func invoke(_ start: (CgSdkResultCallback?, UnsafeMutableRawPointer) -> Void) async throws -> String
```

so each method becomes `try await invoke { cb, ud in cg_sdk_recall(handle, query, optsJSON, cb, ud) }` (the capturing closure is the Swift argument, not the C callback, so the C-convention constraint isn't violated). Collapses ~120 lines to ~30, and any future bridging fix (e.g. the `userData!` force-unwrap) lands in one place instead of seven. A `Void` wrapper over the same helper covers `warm()`.

ios/Sources/CogneeSDK/Cognee.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// (defaults < environment variables < `settingsJSON`).  Network / disk
+    /// access does NOT happen here — it happens on `warm()`.
+    ///
+    /// - Parameter settingsJSON: Optional JSON object whose keys override the
+    ///   environment-loaded settings (e.g. `{"llm_api_key":"sk-…"}`).
+    ///   Pass `nil` to use environment defaults.
+    /// - Throws: `CogneeError` with code 3 (CG_ERR_RUNTIME) if `cg_sdk_new`
+    ///   returns `nil`.
+    public init(settingsJSON: String? = nil) throws {
+        let raw: OpaquePointer? = settingsJSON.map { cg_sdk_new($0) } ?? cg_sdk_new(nil)
```
GraDKh: Nit: `settingsJSON.map { cg_sdk_new($0) } ?? cg_sdk_new(nil)` can just be `cg_sdk_new(settingsJSON)` — Swift bridges `String?` to a nullable C pointer automatically, which this file already relies on when passing `optsJSON: String?` straight into `cg_sdk_add`/`cg_sdk_search`/etc.

.github/workflows/ios.yml
```diff
@@ -0,0 +1,101 @@
+name: iOS bindings
+
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+
+concurrency:
+  group: ios-${{ github.ref }}-${{ github.sha }}
```
GraDKh: The concurrency group includes `${{ github.sha }}`, so every push gets its own group and `cancel-in-progress: true` can never cancel a superseded run — rapid pushes to a PR run multiple full `macos-14` jobs (10×-billed minutes) to completion. Use `group: ios-${{ github.ref }}` like the repo's other workflows.

capi/cognee-capi/Cargo.toml
```diff
@@ -49,6 +54,9 @@ sqlite        = ["cognee-lib/sqlite",                "cognee-bindings-common/sql
 # testing is in defaults because cg_task_context_mock requires mock
 # graph/vector backends — same reason cognee-ts-neon keeps it.
 testing       = ["cognee-lib/testing",               "cognee-bindings-common/testing"]
+# Record/replay cassette-based mock LLM. Required for the iOS demo and offline
+# testing. Enables `llm_mock` / `llm_mock_cassette` settings keys at runtime.
+mock-llm      = ["cognee-lib/mock-llm"]
```
GraDKh, line 60: The comment documents a settings key `llm_mock_cassette` that doesn't exist — the runtime key is `llm_cassette` (`crates/lib/src/config.rs`; your Swift `configureMockMode` uses the correct one). A C-API consumer following this comment gets `CG_ERR_UNKNOWN_CONFIG_KEY`, or worse assumes cassette replay is configured when it silently isn't.

ios/Sources/CogneeSDK/Cognee.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// Enables:
+    /// - Mock LLM that replays responses from `cassettePath`
+    /// - Deterministic mock embeddings (no embedding model needed)
+    /// - In-memory graph store (no database required)
+    /// - In-memory brute-force vector store (no vector database required)
+    ///
+    /// Call this **before** `warm()`.
+    ///
+    /// - Parameter cassettePath: Filesystem path to a `LlmCassette` JSON file.
+    public func configureMockMode(cassettePath: String) throws {
```
GraDKh, line 235: Minor / non-blocking: `configureMockMode` re-encodes the "fully offline mock" recipe that already exists in other consumers, and the copies have started to drift — `crates/cli/src/commands/bench.rs` also sets mock `llm_api_key`/`embedding_api_key`, while this helper adds `graph_database_provider=mock` that the others lack. If offline mode is worth a one-call API (it is), consider defining it once at the lib/capi level (e.g. a single `mock_mode` config key or a shared settings-JSON constant) so Swift, C, TS, and Python bindings can't diverge when the recipe grows a sixth required key.

GraDKh (changes requested):
Thank you for this contribution — an iOS binding is something we've wanted for a while, and it's clear a lot of care went into this PR. The Swift bridging itself is solid: the `ContinuationBox` + `Unmanaged.passRetained`/`takeRetainedValue` pattern, the config keys in `configureMockMode`, and all `cg_sdk_*` signatures check out against `cognee_sdk.h` and the Rust config layer.

A few issues need addressing before merge; the two blocking ones are the invalid Cargo `[target].lib` override (it's silently ignored) and the unrunnable cassette-recording recipe in the README (the committed test fixture has no working regeneration path). Details inline.

Also, the branch has fallen behind — could you rebase onto the current `main`? The capi workspace and CI setup have moved recently (workspace extraction, ORT mirror, release fixes), so a fresh rebase will make sure the checks run against the current layout.

yonkosan:
@GraDKh i have rebased onto current main (`c7d137d` — ORT mirror, release fixes, CRLF header fix).

All points from the second review addressed in `0e5703a`:

**Blockers**

- Removed the invalid `[target.'cfg(target_os = "ios")'.lib]` section. `build_xcframework.sh` now uses `cargo rustc --crate-type staticlib` so only the staticlib slice is built — no cdylib byproduct.
- README cassette recipe was entirely wrong (invented flags that don't exist on `cognee-cli`). Replaced with `COGNEE_RECORD_LLM=<path>` + `bench --memories <file>`, which is what the CLI actually takes.
- Reverted the `crates/cli` changes (`default-features = false` on cognee-lib + the `testing` feature). Those were only there to support the broken recipe and the Android build side-effect was untested.

**Suggestions**

- Dropped `${{ github.sha }}` from the concurrency group so `cancel-in-progress` can actually fire.
- Fixed the comment: `llm_mock_cassette` → `llm_cassette`.
- Committed `capi/include/module.modulemap` as a tracked file. Both the CI step and the build script now `cp` from it — no more independent heredocs that can drift.
- Test now loads the memory text from `memories.json` and applies `memory_to_text` shaping rather than hardcoding it in three places.
- Collapsed the six identical continuation/callback blocks into a shared `invoke` helper. `cg_sdk_new(settingsJSON)` nit fixed too.

On `configureMockMode` drift — agree it belongs at the lib/capi level, leaving that for a follow-up PR to avoid expanding scope here.

yonkosan:
> @GraDKh i have rebased onto current main (`c7d137d` — ORT mirror, release fixes, CRLF header fix).
> 
> All points from the second review addressed in `0e5703a`:
> 
> **Blockers**
> 
> * Removed the invalid `[target.'cfg(target_os = "ios")'.lib]` section. `build_xcframework.sh` now uses `cargo rustc --crate-type staticlib` so only the staticlib slice is built — no cdylib byproduct.
> * README cassette recipe was entirely wrong (invented flags that don't exist on `cognee-cli`). Replaced with `COGNEE_RECORD_LLM=<path>` + `bench --memories <file>`, which is what the CLI actually takes.
> * Reverted the `crates/cli` changes (`default-features = false` on cognee-lib + the `testing` feature). Those were only there to support the broken recipe and the Android build side-effect was untested.
> 
> **Suggestions**
> 
> * Dropped `${{ github.sha }}` from the concurrency group so `cancel-in-progress` can actually fire.
> * Fixed the comment: `llm_mock_cassette` → `llm_cassette`.
> * Committed `capi/include/module.modulemap` as a tracked file. Both the CI step and the build script now `cp` from it — no more independent heredocs that can drift.
> * Test now loads the memory text from `memories.json` and applies `memory_to_text` shaping rather than hardcoding it in three places.
> * Collapsed the six identical continuation/callback blocks into a shared `invoke` helper. `cg_sdk_new(settingsJSON)` nit fixed too.
> 
> On `configureMockMode` drift — agree it belongs at the lib/capi level, leaving that for a follow-up PR to avoid expanding scope here.

One more fix in `032f381` — `memories.json` wasn't declared as a resource in `Package.swift`, so `testWarmAddCognifySearch` was silently skipping instead of running. Added the `.copy("Fixtures/memories.json")` entry; all three tests pass now.

capi/cognee-capi/Cargo.toml
GraDKh: Leaving this open: the manifest key was removed (good, no more `unused manifest key` warning), but the build-script remedy does not actually suppress the cdylib. In `capi/scripts/build_xcframework.sh`:

```
cargo rustc --target … $CARGO_FLAGS -- --crate-type staticlib
```

`--crate-type` is placed **after `--`**, so it is forwarded to `rustc` and *appended* to the crate types Cargo already derives from `[lib] crate-type = ["cdylib","staticlib"]`. rustc then receives `--crate-type cdylib --crate-type staticlib --crate-type staticlib` and still emits the multi-GB cdylib — the exact byproduct this section set out to avoid (and the inline comment claims is gone).

To actually override, pass it as a **Cargo** option, before `--`:

```
cargo rustc --crate-type staticlib --target … $CARGO_FLAGS
```

(This is the form the comment already describes; only the placement is wrong.)

yonkosan:
Rebased onto `f026818` (registry refactor). One non-trivial conflict in `component_manager.rs` — the old inline `init_graph_db` match was replaced by `self.registry.build_graph(&ctx)`, which meant the `#[cfg(feature = "testing")] "mock"` arm had nowhere to live. Moved it into `cognee-components` as `MockGraphFactory`, mirroring the existing `MockVectorFactory` pattern. Also added `cognee-graph/testing` to the `testing` feature, a `"mock"` hint to `unsupported_msg` for graph providers, and extended the drift-guard test to assert `mock` appears in `graph_providers()` when `testing` is on.

Also addressed your follow-up on the `--crate-type` placement — it was after `--`, so rustc was receiving it as an extra flag appended to the existing crate types rather than as an override. Moved it before `--` in both `cargo rustc` invocations.

All three XCTests still pass locally (`testConfigureMockMode`, `testSDKInit`, `testWarmAddCognifySearch`). `cargo check` clean for both `aarch64-apple-ios` and `aarch64-apple-ios-sim` with `--features sqlite,testing,mock-llm`.

yonkosan:
@GraDKh , awaiting for your approval/review for any further changes and requests.

ios/README.md
```diff
@@ -0,0 +1,152 @@
...
+# Record (from repo root).
+#
+# COGNEE_RECORD_LLM   – cassette output path (wraps the real LLM adapter;
+#                        writes the file on Drop).
+# MOCK_EMBEDDING      – use deterministic mock embeddings so only LLM
+#                        responses are recorded (no embedding API calls).
+# LLM_API_BASE / LLM_API_KEY – Ollama's OpenAI-compatible endpoint.
+COGNEE_RECORD_LLM="$(pwd)/ios/Tests/CogneeSDKTests/Fixtures/demo_cassette.json" \
+MOCK_EMBEDDING=deterministic \
+LLM_API_BASE=http://localhost:11434/v1 \
```
GraDKh: `LLM_API_BASE` is not read anywhere in the codebase — I grepped `crates/` and `capi/` and it appears only in this README. The LLM endpoint is loaded from `LLM_ENDPOINT` (with `OPENAI_URL` as the Rust alias) in `crates/lib/src/config.rs`. So following this recipe, the endpoint override is silently discarded: with `--llm-provider openai` the base falls back to the default OpenAI endpoint and tries to authenticate with the key `ollama`, rather than recording against local Ollama.

This means the cassette-regeneration path — the whole point of the earlier blocking comment — still doesn't run. Fix:

```diff
-LLM_API_BASE=http://localhost:11434/v1 \
+LLM_ENDPOINT=http://localhost:11434/v1 \
```

(`COGNEE_RECORD_LLM`, `MOCK_EMBEDDING=deterministic`, and `LLM_API_KEY` are all read correctly — only the endpoint variable name is wrong.)

capi/cognee-capi/build.rs
```diff
@@ -2,4 +2,4 @@ fn main() {
     // Header is maintained manually at capi/include/cognee.h
     // cbindgen was useful for bootstrapping but the manual header
     // gives better control over the C API surface.
-}
+}
```
GraDKh: This hunk's only change is deleting the trailing newline (`\ No newline at end of file`). `rustfmt --check` flags it (wants the file to end in `\n`). It doesn't break CI today — the separate `capi/` workspace isn't covered by any `cargo fmt --check` (root `--all` excludes it, and `capi/scripts/check.sh` runs no fmt) — but it's a purposeless, non-rustfmt-clean edit on a file the PR otherwise doesn't touch. Please restore the trailing newline.

GraDKh (commented):
Thanks for the careful iteration on this — the whole of my first and second review rounds is now resolved. I re-verified each point against the current head: the CI `swiftc -typecheck` boundary, the `cargo rustc --crate-type staticlib` override (correctly before `--` now), the committed `module.modulemap`, the reverted `crates/cli` edits, the `mock` graph provider + error hint, the shared `invoke` helper, and the cassette-driven test loading its text from `memories.json` (I confirmed the `memory_to_text` shaping matches the Rust side byte-for-byte, including empty-refs → `none`). Really solid work.

Two remaining items before merge — one real, the rest nits:

1. **Blocking-ish (regeneration recipe still can't run):** the README record command uses `LLM_API_BASE`, which nothing in the codebase reads — the endpoint is loaded from `LLM_ENDPOINT` (alias `OPENAI_URL`) only. So the recipe silently drops the Ollama endpoint and won't actually record. Details inline. This is the last thread of my earlier "the fixture has no working regeneration path" comment.

2. **Nits:** the `build.rs` trailing-newline removal (inline), plus the two open Copilot nits worth a quick touch-up — `PipelineTests.swift` says `iPhone 16` while the README/PR say `iPhone 17`, and the PR-description file table still claims `build.rs` "emits iOS linker flags + SDKROOT" when it's now a no-op.

The `configureMockMode` cross-binding drift is fine to leave for the follow-up PR as we agreed. Once #1 and the nits are sorted this is good to merge — thank you again for the contribution.

yonkosan:
@GraDKh — addressed all outstanding items in `60f5845` + `f75ee69`:

- `ios/README.md`: `LLM_API_BASE` → `LLM_ENDPOINT` in the recording recipe
- `capi/cognee-capi/build.rs`: trailing newline restored at EOF
- `ios.yml`: added `brew install protobuf` before `cargo check` — `lance-encoding v6.0.0` (pulled in via the rebase onto `f026818`) has a build script that calls `protoc`, which isn't pre-installed on `macos-14` runners; this was the root cause of the iOS CI failure
- `textInputsJSON`: `String(data:encoding:)!` → `try XCTUnwrap(...)` (throws instead of crashing)
- Simulator comment: `iPhone 16` → `iPhone 17`

GraDKh:
Merged — clean iOS bindings with correct C-boundary memory management and feature wiring; please follow up on strengthening the pipeline test assertions and adding a real link/build step to CI.

yonkosan:
Hi @GraDKh, i have raised the Follow-up at #100 as you requested, please take a look.

Issue #19, opened 2026-06-26
Perf: optimize LLM API call usage in cognify (concurrency, batch pipelining, prompt caching)


GraDKh:
## Summary

Investigate and optimize how the pipeline issues **LLM API calls** during `cognify` (fact/graph extraction, summarization, temporal extraction). The goal is to reduce end-to-end latency and wall-clock time without changing output correctness. Focus areas: concurrency control, batch pipelining, prompt caching, and retry behaviour.

This is the LLM slice of the external-API performance investigation. Sibling issues cover embeddings and database batching. Complements #18 (pure-CPU profiling) — this one targets I/O-bound LLM latency.

## Current state (verified)

**LLM client / trait**
- `crates/llm/src/llm_trait.rs` — `Llm` trait (`generate`, `create_structured_output_raw`, `create_structured_output_with_messages_raw`).
- `crates/llm/src/adapters/openai.rs` — `OpenAIAdapter`. Strict JSON-schema mode → function-calling → JSON-mode fallback. Network retries (3, exp backoff 1→2→4s, cap 30s), structured-output retries (5). 600s per-request timeout.

**Where LLM calls happen**
- **Graph/fact extraction** — `crates/cognify/src/tasks.rs:391` + `crates/cognify/src/fact_extraction/extractor.rs:81`. One LLM call per chunk. Chunks processed in batches of `data_per_batch`; within a batch, work is `tokio::spawn`ed and bounded by a `Semaphore(max_parallel_extractions)` (default 20). **Batches are processed sequentially** — batch N+1 does not start until batch N fully drains.
- **Summarization** — `crates/cognify/src/tasks.rs:883` + `crates/cognify/src/summarization/extractor.rs:136`. One LLM call per chunk. At the task level batches run **sequentially** (`for batch in chunks(summarization_batch_size) { summarize_chunks(batch).await }`); the per-chunk spawn inside the extractor is **not bounded by a semaphore**, unlike graph extraction.
- **Temporal extraction** — `crates/cognify/src/tasks.rs:1196`. Two-stage: per-chunk event extraction (semaphore-bounded, good) then **batch-level** entity enrichment (`crates/cognify/src/temporal_extraction/entity_enricher.rs:48` sends one call per batch — the pattern we want elsewhere).

**Config** — `crates/cognify/src/config.rs:210` (`max_parallel_extractions`=20, `data_per_batch`, `summarization_batch_size`=50).

## Optimization opportunities to investigate

[...]

PR #40, opened 2026-06-30
feat(llm): native Anthropic Messages API adapter (Tier 2)

RajdeepKushwaha5:
## Summary

Tier 2 of #17 — add a native Anthropic adapter. Unlike the Tier 1 providers (ollama/mistral/gemini/custom), Anthropic is not OpenAI-compatible, so it needs a dedicated adapter against the Messages API. Builds on #30 (Tier 1 routing).

`AnthropicAdapter` implements the shared `Llm` trait and reuses the existing `Message` / `GenerationOptions` / `GenerationResponse` types, the `LlmError` enum, and the same transient-retry / structured-output-repair shape as `OpenAIAdapter`.

## What differs from the OpenAI path (and how it is handled)

| Anthropic Messages API | Handling |
|---|---|
| Auth via `x-api-key` + `anthropic-version` headers | sent on every request |
| No `system` role inside `messages` | system messages are hoisted into the top-level `system` field; the rest map to `user`/`assistant` |
| `max_tokens` is required | always sent (configurable default, 4096) |
| Structured output via tool use, not `response_format` | a single **forced** `tool_use` whose `input_schema` is the response schema; the tool `input` is returned as the result |
| Trailing slash on the base URL | normalized so request paths never double up |
| litellm-style `anthropic/` model prefix | stripped, matching the Python SDK |

The structured-output path retries up to 5 attempts (instructor parity) on a missing or malformed forced-tool block, and terminal errors (400/401/404) fail fast without burning attempts.

## Wiring

- `ComponentManager::init_llm_adapter` routes `anthropic` to `AnthropicAdapter`  (not the OpenAI-compatible factory).
- The HTTP server's `wire_llm` does the same.
- Transcription stays `None` for `anthropic` (the Messages API has no Whisper route),  consistent with the graceful no-audio behavior for ollama/mistral/gemini.

## Tests / docs

- Unit tests: system hoisting + role mapping, `anthropic/` prefix stripping, base-URL  trailing-slash normalization, tool-schema sanitization, and response parsing  (text concatenation + forced `tool_use` extraction with usage mapping).
- `crates/llm/tests/integration_anthropic.rs`: gracefully-skipped live test (generate  + structured output), skipping when `ANTHROPIC_API_KEY` is absent so CI stays green  without secrets, mirroring `integration_openai.rs` / `integration_providers.rs`.
- `docs/configuration.md` describes the native adapter (auth, endpoint default,  prefix stripping, no transcription).
- Verified: `fmt`, `clippy --all-targets -D warnings` (cognee-llm / cognee-lib /  cognee-http-server), the 5 unit tests, and the integration tests compile + skip.

## Scope

Part of #17 — does not close it. Tier 3 (native Azure and Bedrock) follows separately.
Builds on #30; until that merges this PR's diff includes the Tier 1 commits, after which GitHub drops the overlap.


RajdeepKushwaha5:
Stacked on #30 - the `Test (stable)` failure is inherited from there (see my note on #30). Everything that doesn't need a live key passes locally; happy to target the specific failing test if the log can be shared.

GraDKh (changes requested):
Thanks for this — the adapter is well-structured and the unit tests cover the tricky mapping bits (system hoisting, prefix stripping, tool extraction). I reviewed it against the Python cognee Anthropic path (`litellm_instructor/.../anthropic/adapter.py`, `get_llm_client.py`, `config.py`, `retry_config.py`) to check parity. One issue blocks merge; the rest are non-blocking.

## 🔴 Blocking — `max_tokens` handling diverges from Python and breaks both caller shapes

Anthropic requires `max_tokens`, so the adapter has to supply a value. The current logic (`base_request`: `opts.max_tokens.unwrap_or(self.default_max_tokens)`) mishandles both ways callers express "use the model's budget", and neither path matches Python:

**a) `None`-options callers send 16384 → HTTP 400 on the default model.**
`GenerationOptions::default()` sets `max_tokens: Some(16384)` (`crates/llm/src/types.rs`), so a caller passing `None` (e.g. `crates/search/src/utils/feedback_detection.rs`) resolves via `unwrap_or_default()` to `Some(16384)` — the `DEFAULT_MAX_TOKENS = 4096` fallback never runs. `claude-3-5-sonnet-20241022` (the model in the tests/docs) caps output at 8192, so the Messages API returns `400 "max_tokens: 16384 > 8192"`, which `call_api` treats as terminal. Every such call fails.

**b) cognify's `max_tokens: None` becomes a hard 4096 cap → truncation / silent data loss.**
`fact_extraction/extractor.rs`, `temporal_extraction/event_extractor.rs`, and `entity_enricher.rs` deliberately pass `Some(GenerationOptions { max_tokens: None, .. })`, with comments stating this is Python parity for "no output cap … a small cap truncates large graphs mid-JSON." On Anthropic that `None` becomes 4096. `with_max_tokens(...)` is never called at either construction site (`component_manager.rs`, `http-server/src/wiring.rs`), and there's no `stop_reason == "max_tokens"` truncation guard — a truncated `tool_use.input` is returned as-is, failing downstream deserialization (aborts cognify in fact extraction; **silently swallowed to an empty `Vec` in the temporal paths**). So the parity comments are inaccurate for Anthropic, and 4096 is 2–4× smaller than the budget Python grants.

**Python does neither.** It carries the same `llm_max_completion_tokens = 16384` ceiling (`config.py`) but clamps it against the model's real limit before use:
```python
model_max = get_model_max_completion_tokens(llm_config.llm_model)  # from litellm.model_cost[model]["max_tokens"]
max_completion_tokens = min(model_max, llm_config.llm_max_completion_tokens) if model_max else llm_config.llm_max_completion_tokens
# ... anthropic/adapter.py:
self.llm_args.setdefault("max_tokens", max_completion_tokens)
```
So Python sends `min(model_max, 16384)` — 8192 for Claude 3.5 Sonnet, higher for newer models, never 4096.

**Suggested fix:** wire `llm_max_completion_tokens` into the adapter and clamp the effective `max_tokens` to the target model's documented output limit (mirroring Python's `min(model_max, ceiling)`), rather than the split between an unreachable-4096 fallback and a pass-through 16384.

## 🟡 Non-blocking

**Misleading comment + inconsistency on 400 handling (`create_structured_output_with_messages_raw`).**
The comment says terminal errors `(400/401/404)` are surfaced immediately, but the fast-fail arm matches only `AuthenticationError` (401) and `ModelNotFound` (404); a 400 arrives as `InvalidResponse` and is re-issued `structured_output_retries` (5) times. Two notes: there's no backoff in this loop (it's 5 rapid attempts — backoff lives only in `call_api`'s network loop, which a 400 short-circuits), and retrying 400 is actually *faithful* to Python (tenacity's `retry_if_not_exception_type` excludes only 404/401). So the retry behavior is fine to keep, but please fix the comment — and note this diverges from the OpenAI adapter, which propagates a 400 in ~2 calls. Distinguishing a terminal 400 from the retryable "no tool_use block" case would need a separate error marker since they currently share `InvalidResponse`.

**No schema validation / reask for structured output (parity gap with instructor).**
The adapter returns the first `tool_use.input` unvalidated; a malformed-but-present input then fails hard at the outer `serde_json::from_value` with no retry. Python/instructor validates the tool input against the schema and, on `ValidationError`, appends a reask message ("Validation Error found… Recall the function correctly, fix the errors") before retrying. Matching that would give the same self-correction (larger change; validate `input` against `json_schema` and reask). Note the current retry loop's "missing tool_use block" branch is near-dead code, since a *forced* `tool_choice` requires Anthropic to emit the block.

**Wiring + `call_api` duplication (maintainability).**
The empty-key check, endpoint `Option`, `retries.max(1)`, and the `.with_structured_output_retries().with_network_retries()` chain are hand-duplicated across `wiring.rs` and `component_manager.rs` (diverging on the missing-key path: `warn!`+`None` vs `ComponentError`), and the `call_api` retry loop / backoff formula / 600s `Client::builder()` are near-verbatim copies of `OpenAIAdapter`. Anthropic legitimately can't route through the OpenAI-compatible factory (Azure is similarly special-cased), so the ask is just to extract the shared boilerplate (e.g. `build_llm_http_client`, a `backoff_delay_ms` helper, an `is_terminal_error` predicate).

## 🔵 Optional (nit)
`generate()` with a system-only message vec would emit `"messages": []` (Anthropic 400s on an empty array), because `split_messages` hoists all `System` messages into `system`. No current caller hits this and Python behaves the same, but a guard in `base_request` (synthesize a user turn, or return a clear error when `converted` is empty) would harden the public method.

RajdeepKushwaha5:
hi sir @GraDKh 
thanks for the thorough review, the max_tokens catch was spot on.

**Blocking, max_tokens: fixed.** The adapter now clamps the per-request `max_tokens` to the model's documented output cap, as you suggested. The effective value is `min(caller value or the llm_max_completion_tokens ceiling, model_max)`, mirroring Python's `min(model_max, llm_max_completion_tokens)`. So:
- a None-options caller (`GenerationOptions::default` sets 16384) now sends 8192 on Claude 3.5 Sonnet, not 16384, so no more 400,
- cognify's `max_tokens: None` now resolves to the clamped ceiling (8192), not a truncating 4096.

I wired `llm_max_completion_tokens` through `LlmInputs` (Settings to the factory to the adapter), added a `model_max_output_tokens` lookup (3.5/3.7 to 8192, Claude 3 to 4096, 4.x and unknown fall back to the ceiling, which is well under their real caps), and added a unit test covering all four cases.

**Rebased onto #56.** Since #56 moved LLM wiring into the pluggable registry, I re-homed the Anthropic integration from the inline `component_manager`/`wiring` code into a proper `AnthropicLlmFactory` registered in `with_builtins()`. That resolves the merge conflict and removes the two hand-duplicated wiring sites you flagged (one factory now, instead of the `component_manager` + `wiring.rs` pair that diverged on the missing-key path).

**Non-blocking:**
- Misleading 400 comment: fixed. It now states that 401/404 are terminal and a 400 is retried (faithful to Python's tenacity, which excludes only 401/404), and that it shares `InvalidResponse` with the "missing tool_use block" case. Behavior unchanged.
- Schema validation / reask: agreed this is the right instructor-parity improvement, but it is a larger change (validate the tool input against the schema, append a reask on `ValidationError`, retry). I would rather land it as a focused follow-up than grow this PR.
- `call_api` / backoff / client duplication with `OpenAIAdapter`: real cleanup, but a broader refactor touching both adapters (a shared `build_llm_http_client` / `backoff_delay_ms` / `is_terminal_error`). Better as a separate PR so this one stays scoped to the Tier 2 adapter.
- Empty system-only-messages guard (the nit): no current caller hits it and Python behaves the same, so I left it out to keep the diff focused. Happy to add the guard if you would like it in.

Verified: fmt, `clippy --all-targets -D warnings` (llm / components / lib / http-server), the Anthropic unit tests including the max_tokens clamp, and the integration test compiles and skips. Rebased onto latest main.

RajdeepKushwaha5:
I have added the missing `llm.api_call` span to `AnthropicAdapter::call_api`, matching the OpenAI adapter (`cognee.llm.model` + `cognee.llm.provider = "anthropic"`, with the `url` field recorded after it is built). Added a mirrored span integration test (`crates/llm/tests/anthropic_span_instrumentation.rs`) that asserts the span and its fields against a mocked `/messages` endpoint, so Anthropic traffic is now visible to telemetry on par with OpenAI.

Verified: fmt, `clippy --all-targets -D warnings`, and the `cognee-llm` tests (the new span test plus the existing OpenAI span tests still pass).

RajdeepKushwaha5:
Hello @GraDKh,

Could you please review this PR when you have a moment? Thanks

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,568 @@
...
+    /// The `max_tokens` to send: the caller's value, or the configured ceiling
+    /// when unset, clamped to the model's documented output cap so Anthropic
+    /// never 400s on `max_tokens > model limit` (and so callers that leave it
+    /// unset get the full model budget rather than a hard 4096 cap). Mirrors
+    /// Python's `min(model_max, llm_max_completion_tokens)`.
+    fn effective_max_tokens(&self, opts: &GenerationOptions) -> u32 {
+        let requested = opts.max_tokens.unwrap_or(self.max_completion_tokens);
+        match Self::model_max_output_tokens(&self.model) {
+            Some(cap) => requested.min(cap),
+            None => requested,
```
GraDKh: **🔴 correctness — raised `LLM_MAX_COMPLETION_TOKENS` 400s every request on Claude 4.x / unknown models**

`model_max_output_tokens()` only recognises the `claude-3` family; for any Claude 4.x model (the current default family) it returns `None`, so `effective_max_tokens` forwards `requested` (line 143) **unclamped**.

If an operator raises `llm_max_completion_tokens` above a Claude 4 model's documented output cap — e.g. `40000` with `claude-opus-4` (32k cap) — the adapter sends `max_tokens=40000`; Anthropic rejects `max_tokens > model limit` with HTTP 400 (mapped to `InvalidResponse`), so every `generate()` fails and every structured-output call exhausts its retries into `MaxRetriesExceeded`.

The doc comment's assumption that the ceiling is "comfortably under the 4.x-family output limits" only holds for the default `16384`, not a configured value. This is the same blocking `max_tokens` divergence GraDKh flagged — the clamp was added, but the fix stops at Claude 3.x, so the exact class of bug (a configured ceiling exceeding the real cap → terminal 400) survives for the newer models. Consider clamping the 4.x/unknown case to a documented per-family default (≥32k) rather than passing through unclamped.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,568 @@
...
+        )))
+    }
+
+    /// Build the base request body shared by completion and structured output.
+    fn base_request(&self, messages: &[Message], opts: &GenerationOptions) -> Value {
+        let (system, converted) = Self::split_messages(messages);
+        let mut body = json!({
+            "model": self.model,
+            "max_tokens": self.effective_max_tokens(opts),
+            "messages": converted,
```
GraDKh, line 354: **🟠 correctness — system-only message list emits `"messages": []`, a terminal Anthropic 400**

`split_messages` hoists every `System` message into the top-level `system` field and returns an empty `converted` vec, so `base_request` emits `"messages": []`. Any caller invoking the public `generate` / structured-output methods with only system-role messages gets HTTP 400 ("messages: at least one message is required"), which `call_api` classifies as non-retryable and fails hard. The OpenAI backend tolerates the same input.

This is GraDKh's 🔵 nit, still unaddressed — no in-tree caller hits it today, so it's a latent break in the public API surface. A guard here (synthesize a user turn, or return a clear error when `converted` is empty) would harden it.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,568 @@
...
+                .unwrap_or_else(|_| request_body.to_string());
+            eprintln!("\n[COGNEE_DEBUG_LLM_REQUEST] POST {url}\n{pretty}\n");
+        }
+
+        let mut last_error = LlmError::NetworkError("No attempt made".to_string());
+
+        for attempt in 0..=self.network_retries {
+            debug!(attempt, "Anthropic API attempt");
+            if attempt > 0 {
+                let delay_ms = (1_000u64 * 2u64.saturating_pow(attempt as u32 - 1)).min(30_000);
```
GraDKh: **🟡 cleanup — inline no-jitter backoff reintroduces the thundering-herd problem `retry_backoff` was added to fix**

This delay is deterministic: `(1000 * 2^(attempt-1)).min(30000)`, no jitter. A batch of concurrent cognify requests all hitting 429 at the same instant will retry at exactly the same moments (1s, 2s, 4s…), re-tripping the rate limit in lockstep — the exact regression `crate::retry::retry_backoff` (equal jitter, issue #19) exists to prevent. The OpenAI adapter already uses it (`openai.rs:219`, `:975`).

Reusing `crate::retry::retry_backoff(attempt as u32)` here removes the duplicated cap/backoff math and restores jitter. Partially addresses GraDKh's "call_api duplication" 🟡 note.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,568 @@
...
+    #[serde(default)]
+    output_tokens: u32,
+}
+
+impl From<AnthropicUsage> for TokenUsage {
+    fn from(u: AnthropicUsage) -> Self {
+        TokenUsage {
+            prompt_tokens: u.input_tokens,
+            completion_tokens: u.output_tokens,
+            total_tokens: u.input_tokens + u.output_tokens,
```
GraDKh: **🔵 cleanup — unguarded `u32 + u32` for `total_tokens`**

`u.input_tokens + u.output_tokens` panics on overflow in debug builds and wraps in release. Extremely unlikely with realistic token counts, but `saturating_add` (or u64 accumulation) is safe and free. Minor.

RajdeepKushwaha5:
Thanks, all four addressed (rebased onto latest main first, which brought in `crate::retry`):

- **max_tokens on Claude 4.x:** `model_max_output_tokens` now clamps every model, not just the 3.x family: Opus 4 to 32k, Sonnet 4 / Haiku 4 to 64k, and any unknown/future model to a conservative 32k floor. The unclamped passthrough branch is gone, so a raised `llm_max_completion_tokens` can never exceed the model's real cap and 400 the request. `effective_max_tokens` is now just `requested.min(model_max)`. Added regression tests: Opus 4 at a 40k ceiling clamps to 32k, Sonnet 4 at 40k passes through (under its 64k cap), and an unknown model at 40k clamps to 32k.
- **system-only message list:** `base_request` now folds a system-only list into a single user turn, so it no longer emits `"messages": []` (the terminal 400). This matches the OpenAI backend's tolerance for the same input. Added a unit test.
- **backoff jitter:** `call_api` now uses `crate::retry::retry_backoff(attempt as u32)`, the same equal-jitter helper the OpenAI adapter uses, instead of the inline deterministic `2^n` delay. Jitter is restored and the duplicated cap/backoff math is gone, so a batch that all hit 429 at once no longer retries in lockstep.
- **total_tokens overflow:** now `saturating_add`.

Verified: rustfmt clean, and the updated unit tests (the max_tokens clamp cases and the system-only guard). clippy and the full test run go through CI.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,633 @@
...
+        let mut last_error =
+            LlmError::InvalidResponse("No structured-output attempt made".to_string());
+
+        for attempt in 0..self.structured_output_retries {
+            if attempt > 0 {
+                debug!(attempt, "retrying Anthropic structured output");
+            }
+            match self.call_api(body.clone()).await {
+                Ok(response) => match response.tool_input(STRUCTURED_OUTPUT_TOOL) {
+                    Some(input) => return Ok(input),
```
GraDKh: **🔴 correctness (blocking) — structured output has no schema-validation repair retry**

This loop only retries when the forced `tool_use` block is *entirely absent* (the `None` arm). When the block is present, `Some(input) => return Ok(input)` returns immediately **with no validation** — so a JSON-valid-but-*incomplete* tool input (the model hit `max_tokens` mid-object, `stop_reason == "max_tokens"`) is returned as `Ok`, then fails downstream in `LlmExt::create_structured_output_with_messages` at `serde_json::from_value` with `DeserializationError`. A single recoverable extraction aborts the whole cognify chunk.

Root cause: this adapter implements `create_structured_output_with_messages_raw` but does **not** override `create_structured_output_with_messages_raw_validated`, so the trait's default impl discards the validator (`llm_trait.rs`: `let _ = validator;`) and never re-asks. The OpenAI adapter does the opposite — it overrides `_raw_validated`, builds `schema_required_validator`, and on a validation miss appends a corrective instruction and retries (`openai.rs` `append_corrective_instruction` + the tool-outcome loop). Python's `instructor` path also retries validation failures, so this is both a parity gap and a real abort.

Please port the `_raw_validated` override here — simpler than OpenAI's, since there is exactly one mode (the forced tool) with no legacy/JSON fallback chain: on a present-but-invalid `input`, re-ask with a corrective message; and inspect `stop_reason == "max_tokens"` (currently only surfaced as `finish_reason` in `generate`, never checked in this path) so truncation is retried rather than returned as `Ok`.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,633 @@
...
+                // Auth (401) and model-not-found (404) are terminal: surface them
+                // immediately rather than burning retries. A 400 arrives as
+                // InvalidResponse and is retried below (up to the retry cap),
+                // which matches Python (tenacity's retry_if_not_exception_type
+                // excludes only 401/404); it also shares the InvalidResponse type
+                // with the retryable "missing tool_use block" case above.
+                Err(e @ (LlmError::AuthenticationError(_) | LlmError::ModelNotFound(_))) => {
+                    return Err(e);
+                }
+                Err(e) => last_error = e,
```
GraDKh: **🟠 correctness / efficiency — retry loop re-invokes `call_api` after its own retries are already exhausted, with no outer backoff**

`call_api` already retries 429 / 5xx / network `network_retries` times with jittered backoff and returns `MaxRetriesExceeded` when it gives up (≈ line 282). This `Err(e) => last_error = e` arm then continues the outer loop and calls `call_api` **again from scratch**. With the factory wiring both knobs from `ctx.llm.max_retries` (`builtins/llm.rs`), a persistent 429 becomes up to `structured_output_retries × (network_retries + 1)` = 5 × 4 = **20** HTTP calls against a rate-limited endpoint — and there is no `sleep` between outer iterations (only a `debug!`), so each restart begins backoff at attempt 0 and hammers the API.

Two fixes: (1) treat a `MaxRetriesExceeded` from `call_api` as terminal here (return / break), the way the OpenAI loop breaks on `Err` instead of re-looping the identical request; (2) the outer loop must still retry the missing-tool-block and validation-repair cases (see the blocking comment above) — for those, add backoff between outer iterations. Python uses `wait_exponential_jitter(8, 128)` *between* structured retries, which this path currently omits.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,633 @@
...
+        self
+    }
+
+    /// Documented maximum output tokens for known Anthropic model families,
+    /// mirroring the `litellm.model_cost[model]["max_tokens"]` lookup the Python
+    /// adapter uses. Every branch returns a value so the configured ceiling is
+    /// always clamped: an operator-raised `llm_max_completion_tokens` can never
+    /// be forwarded above a model's real cap and 400 the request. Unknown /
+    /// future models clamp to the smallest documented Claude 4.x cap (32k).
+    fn model_max_output_tokens(model: &str) -> u32 {
```
GraDKh, line 169: **🟡 correctness — hardcoded per-family output caps are already stale, not just at risk of drift**

The doc comment frames this as Python's `litellm.model_cost[model]["max_tokens"]` lookup, but as a hardcoded table it is already wrong against current documented Anthropic limits:
- `claude-3-7` → 8192, but 3.7 Sonnet supports **64K** output (extended) — clamps well under the real cap.
- `opus-4` matches Opus **4.5 / 4.6 / 4.7 / 4.8** (64K–128K output) → all clamped to 32k.
- `sonnet-4` covers Sonnet **4.6** (up to 128K) → clamped to 64k.
- the `else` branch clamps every unknown / future model *down* to 32k, so each new release silently under-budgets until the table is hand-edited.

This is mitigated — it only bites when the operator raises `LLM_MAX_COMPLETION_TOKENS` above the cap, and it fails safe (reduced output budget, never a 400) — so it is not blocking. But since the intent is to mirror `litellm.model_cost`, consider making it data-driven (a static map keyed by documented model IDs, or a caps lookup) so it does not need a source edit per model release, and refresh the current values (3.7 Sonnet, Opus 4.5+, Sonnet 4.6).

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,633 @@
...
+                    .unwrap_or_else(|_| "Unknown error".to_string());
+                let err = match status.as_u16() {
+                    401 => LlmError::AuthenticationError(error_body),
+                    429 => LlmError::RateLimitExceeded(error_body),
+                    400 => LlmError::InvalidResponse(format!("Bad request: {error_body}")),
+                    404 => LlmError::ModelNotFound(error_body),
+                    _ => LlmError::ApiError(format!("HTTP {status}: {error_body}")),
+                };
+                // Non-retryable: bad request, auth, or unknown model.
+                if matches!(status.as_u16(), 400 | 401 | 404) {
```
GraDKh: **🟡 parity — `402 Payment Required` is missing from the terminal (non-retryable) set**

Retrying 400 here is intentional and correct — the comment in the structured loop rightly notes tenacity's `retry_if_not_exception_type` retries 400s. The gap is that Python excludes **more** than 401/404 from retry: its policy also excludes `LLMPaymentRequiredError` (HTTP 402). Here a 402 falls into the `_ => ApiError` arm, so `call_api` retries it and then the structured loop retries it again — burning the full retry budget on a billing error that can never succeed. Add a `402 => …` mapping and include it in this terminal `matches!(…)` set (and correspondingly in the structured loop's terminal-error arm) for parity.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,633 @@
...
+            32_000
+        }
+    }
+
+    /// The `max_tokens` to send: the caller's value, or the configured ceiling
+    /// when unset, clamped to the model's documented output cap so Anthropic
+    /// never 400s on `max_tokens > model limit`. Mirrors Python's
+    /// `min(model_max, llm_max_completion_tokens)`.
+    fn effective_max_tokens(&self, opts: &GenerationOptions) -> u32 {
+        let requested = opts.max_tokens.unwrap_or(self.max_completion_tokens);
```
GraDKh: **🟡 cleanup — configured `LLM_MAX_COMPLETION_TOKENS` is shadowed for default-options callers**

`opts.max_tokens.unwrap_or(self.max_completion_tokens)` consults the configured ceiling (wired from `llm_max_completion_tokens` via `with_max_completion_tokens`) **only** when the caller passes `max_tokens: None`. The core cognify extractors do pass `None` deliberately (a documented Python-parity invariant), so they honor it correctly. But `GenerationOptions::default()` sets `max_tokens: Some(16384)` (`types.rs`), so `generate()` callers and any structured caller using default options silently bypass a *lower* configured ceiling — an operator setting `4096` to cap cost still gets `min(16384, model_cap)`.

Not blocking. Consider applying the ceiling as a `min` even when `opts.max_tokens` is `Some(...)`, so `llm_max_completion_tokens` is a true upper bound regardless of call site (Python treats it as the effective ceiling on all paths), or document why default-options callers are intentionally exempt.

GraDKh (changes requested):
## Review: native Anthropic Messages API adapter — changes requested

Thanks for the iteration. The previous round's items are all resolved in the current head: system-only input now folds into a user turn, `call_api` uses the shared jittered `retry_backoff`, `total_tokens` uses `saturating_add`, and the `max_tokens > model cap` 400 is now clamped for every family. The wire-format mapping (system hoisting, forced single-tool structured output, `tool_use` extraction), the factory wiring, and the unit tests are solid.

The findings below are from a high-effort review cross-checked against `openai.rs` and the Python `litellm_instructor` reference. **One blocking item** (structured output has no validation-repair) plus four should-fix / consider cleanups. Each is inline at the relevant line.

**Blocking**
- 🔴 Structured output has no schema-validation repair retry: a truncated-but-present `tool_use.input` is returned as `Ok(partial)` and aborts the cognify chunk at deserialization, where the OpenAI adapter re-asks.

**Should fix**
- 🟠 The structured retry loop re-enters `call_api` after it has already exhausted its network retries → up to `structured_output_retries × (network_retries + 1)` HTTP calls, with no backoff between outer iterations.
- 🟡 Hardcoded per-family output caps are already stale vs current Anthropic limits (3.7 Sonnet, Opus 4.5+, Sonnet 4.6).

**Consider**
- 🟡 `402 Payment Required` is missing from the terminal-error set (Python excludes it from retry).
- 🟡 Configured `LLM_MAX_COMPLETION_TOKENS` is shadowed for callers passing default `GenerationOptions`.

**Follow-up (not blocking this PR).** This is now the second hand-rolled copy of the `call_api` retry / error-mapping / structured-retry skeleton, and Tier 3 (native Azure / Bedrock, issue #17) will add more. Every finding below is a variant of "copied the HTTP skeleton, dropped the OpenAI adapter's hard-won repair/validation logic" — that failure mode will recur. Recommend a tracked follow-up to extract a shared `llm::http` core (post-with-retries + the corrective-retry/validator loop, generic over "extract candidate payload from response") and port both adapters onto it *before* Tier 3 lands. Doing that extraction inside this PR would balloon its scope and churn the parity-critical `openai.rs`, so it should be its own change — but it should be tracked now, not discovered again in Bedrock.

RajdeepKushwaha5:
Thanks for the high-effort pass. All five items are addressed.

**Blocking**

- **Structured output now has schema-validation repair.** The adapter overrides `create_structured_output_with_messages_raw_validated` and both structured paths go through one `structured_output_impl`. A present-but-invalid tool input (fails the schema validator, or was truncated at `max_tokens` mid-object, `stop_reason == "max_tokens"`) is now re-asked with a corrective instruction carrying the reason, instead of being returned as `Ok(partial)` and aborting the cognify chunk at deserialization. This mirrors the OpenAI adapter's reask.

**Should fix**

- **The structured loop no longer re-enters an exhausted `call_api`.** A `MaxRetriesExceeded` from `call_api` (and auth / model-not-found / the new payment-required) is now terminal in the outer loop, so a persistent 429 can no longer fan out to `structured_output_retries x (network_retries + 1)` attempts. The remaining retryable cases (missing tool block, validation repair) now back off between outer iterations via the shared jittered `retry_backoff`, so re-asks are not fired in a tight loop.
- **Model output caps are data-driven and refreshed.** `model_max_output_tokens` is now a substring-matched, order-sensitive table (narrower point-release patterns before broader ones) with a documented conservative fallback, so adding a model is a data edit rather than a new branch. Values updated to the ones you flagged: 3.7 Sonnet -> 64K, Opus 4.5+ -> 64K, Sonnet 4.6 -> 128K. A test pins the ordering so a broad pattern cannot shadow a narrow one. (Sourcing this from a real caps dataset is noted as belonging with the shared-http follow-up below.)

**Consider**

- **402 Payment Required is now terminal.** Added `LlmError::PaymentRequired`, mapped from HTTP 402 in `call_api`, and included in the terminal set of both `call_api` and the structured loop, matching Python's policy of excluding `LLMPaymentRequiredError` from retry.
- **`LLM_MAX_COMPLETION_TOKENS` is now a true upper bound on every path.** `effective_max_tokens` is `min(caller value, configured ceiling, model cap)`, so a default-options caller (`GenerationOptions::default()` carries `Some(16384)`) no longer bypasses a lower configured ceiling. Added a test for the default-options path.

**On the copy-paste concern:** rather than hand-roll a second validator, I extracted `schema_required_validator` into `crate::schema` and both adapters now share it (the OpenAI call site and its test are unchanged, kept as a thin delegate). I fully agree with the larger follow-up: a shared `llm::http` core (post-with-retries + the corrective-retry/validator loop, generic over "extract candidate payload from response") is the right home before Tier 3 lands, and it should be its own change so it does not churn the parity-critical `openai.rs` here. Happy to open and track that issue.

Verified: rustfmt clean, `cargo clippy --all-targets -D warnings` green, and the `cognee-llm` unit + integration tests pass.

RajdeepKushwaha5:
Heads-up on the red `Test (no secrets)` here: it is not coming from this PR. It is a pre-existing failure on `main`, and this PR only inherits it because CI tests the branch merged into `main`.

**What is happening**

#83 made `KnowledgeGraph.nodes`/`edges` required (dropping `serde(default)`) so a nodes-only payload drives a corrective retry instead of silently yielding a 0-edge graph, and it fixed `MockLlm`'s canned fallback to use the real field name. It missed one spot: `crates/cognify/tests/integration_text_summary_payload.rs` seeds its own response via `MockLlm::new(..)` and still used `relationships`, a name the derive has always ignored. With `edges` now required that payload no longer deserializes, so the cognify pipeline aborts at the graph-extraction task.

**Evidence that it is main, not this branch**

`main`'s own CI (`Test (stable)` on `35c738b`) is red with the identical failure:

```
Summary [458.162s] 2403 tests run: 2401 passed, 2 failed, 4 skipped
TRY 1 FAIL  cognee-cognify::integration_text_summary_payload text_summary_payload_contains_text_field
TRY 2 FAIL  cognee-cognify::integration_text_summary_payload text_summary_payload_contains_text_field
  cognify pipeline should succeed with mock backends:
    Execute("task 2 failed after 1 attempt(s): LLM error: Deserialization error:
             Failed to deserialize structured output: missing field `edges`")
```

Exactly one test out of 2414 fails in this PR's lane, and it is that same one.

**The fix**

I opened #87 with the one-line fix against `main` (seed `edges` instead of `relationships`, matching the model and `MockLlm`'s canned fallback; the test still asserts an empty graph). This PR carries the same one-liner so it can go green in the meantime. Once #87 lands, the duplicate drops out of this branch on the next rebase.

(For completeness: `main`'s `Test (stable)` also shows a second, unrelated failure, `cognee-search::temporal_retriever_multiple_events`. Neither this PR nor #87 touches it.)

**One correction to my previous comment**

I said I had extracted `schema_required_validator` into `crate::schema` so both adapters share it. I have since reverted that: `openai.rs` and `schema.rs` are back to byte-identical with `main`, and the Anthropic adapter carries its own local validator. That matches your guidance that the shared extraction should be its own change rather than churn the parity-critical `openai.rs` here, and it keeps this PR scoped to the Tier 2 adapter. All five review items are otherwise unchanged and still in place.

Rebased onto latest `main`.

RajdeepKushwaha5:
hi sir @GraDKh 
could you please review it now ?

crates/components/src/builtins/llm.rs
```diff
@@ -80,6 +80,50 @@ impl LlmFactory for OpenAiCompatibleLlmFactory {
...
+        ANTHROPIC_PROVIDER
+    }
+
+    async fn build(&self, ctx: &BackendBuildContext) -> Result<Arc<dyn Llm>, ComponentError> {
+        if ctx.llm.api_key.trim().is_empty() {
+            return Err(ComponentError::Config(
+                "anthropic provider requires an API key (set LLM_API_KEY)".to_string(),
+            ));
+        }
+        let endpoint = if ctx.llm.endpoint.trim().is_empty() {
```
GraDKh: **Endpoint leak — misroutes Anthropic traffic (requesting change).**

`ctx.llm.endpoint` is aliased from `OPENAI_URL` (`crates/lib/src/config.rs:292`), which is a documented-required var in this repo's own setup. A user who only flips `LLM_PROVIDER=anthropic` while `OPENAI_URL` is still set passes the OpenAI host as `base_url`, so every request POSTs to `{openai_url}/messages` with an `x-api-key` header → 404/401 on all cognify/search calls.

Parity check: Python's `AnthropicAdapter` never receives an endpoint at all — `get_llm_client.py` passes only api_key/model/max_completion_tokens/instructor_mode/llm_args and builds `AsyncAnthropic(api_key=...)` with no `base_url`. Suggest ignoring `OPENAI_URL` here and only honoring an Anthropic-specific endpoint override that was explicitly set for this provider.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+
+    fn model(&self) -> &str {
+        &self.model
+    }
+
+    fn supports_function_calling(&self) -> bool {
+        true
+    }
+
+    fn max_context_length(&self) -> u32 {
```
GraDKh, line 622: **Missing `transcribe_image` → image cognify aborts (requesting change).**

This `impl Llm` overrides neither `transcribe_image` nor `supports_vision`, so both fall back to the trait default (`crates/llm/src/llm_trait.rs`) returning `FeatureNotSupported`. `ImageLoader` is fail-fast (`crates/ingestion/src/loaders/image.rs`), so a dataset containing an image aborts the whole cognify run under `LLM_PROVIDER=anthropic` — even though Claude 3+/4 are vision-capable and `OpenAIAdapter` overrides this (`openai.rs:666`).

Parity: Python's Anthropic path supports vision via the inherited `GenericAPIAdapter.transcribe_image` (litellm routes bare `claude-*` to Anthropic). Please add a `transcribe_image` override.

crates/components/src/builtins/llm.rs
```diff
@@ -80,6 +80,50 @@ impl LlmFactory for OpenAiCompatibleLlmFactory {
...
+            None
+        } else {
+            Some(ctx.llm.endpoint.clone())
+        };
+        let adapter =
+            AnthropicAdapter::new(ctx.llm.model.clone(), ctx.llm.api_key.clone(), endpoint)
+                .map_err(|e| ComponentError::Llm(e.to_string()))?
+                .with_structured_output_retries(ctx.llm.max_retries)
+                .with_network_retries(ctx.llm.max_retries)
+                .with_max_completion_tokens(ctx.llm.max_completion_tokens);
```
GraDKh: **`LLM_ARGS` silently dropped (requesting change).**

The builder chain never applies `ctx.llm.llm_args` — contrast `OpenAiCompatibleLlmFactory` at `crates/components/src/builtins/llm.rs:52` (`.with_extra_args(...)`). This is bigger than a missing factory call: `AnthropicAdapter` has no extra-args mechanism at all (no field, no `with_extra_args`, no merge in `call_api`). Python's Anthropic adapter honors `llm_args` via `merged_kwargs = {**self.llm_args, **kwargs}` and even depends on it to inject `max_tokens`. So operator-configured `LLM_ARGS` (e.g. `top_p`) have no effect under `anthropic`. Fix needs adapter support **plus** this factory wiring.

crates/llm/src/error.rs
```diff
@@ -31,6 +31,12 @@ pub enum LlmError {
     #[error("Authentication failed: {0}")]
     AuthenticationError(String),
 
+    /// HTTP 402 — billing/credit exhausted. Terminal: retrying can never make a
+    /// billing failure succeed, so it is excluded from retry the way Python's
+    /// `retry_if_not_exception_type` excludes `LLMPaymentRequiredError`.
+    #[error("Payment required: {0}")]
+    PaymentRequired(String),
```
GraDKh, line 38: **402 handling now inconsistent across adapters (requesting change).**

Good addition, but it's only mapped/terminal in the Anthropic adapter. The OpenAI adapter's `send_chat_request` (`crates/llm/src/adapters/openai.rs`) still matches only 401/429/400 and lets 402 fall into the retryable `_ => ApiError` catch-all, so a billing-exhausted OpenAI-compatible endpoint burns the full retry budget instead of failing fast. Python excludes 402 from retry on the OpenAI path too (`LLMPaymentRequiredError`). Please map 402 → `PaymentRequired` and mark it terminal in the OpenAI adapter as well, so the new variant is applied consistently.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+    /// caller would otherwise silently bypass a lower operator-configured ceiling
+    /// (an operator setting 4096 to cap cost would still get 8192). Python treats
+    /// `llm_max_completion_tokens` as the effective ceiling on all paths. The
+    /// model cap is applied last so Anthropic never 400s on
+    /// `max_tokens > model limit`.
+    fn effective_max_tokens(&self, opts: &GenerationOptions) -> u32 {
+        let requested = opts.max_tokens.map_or(self.max_completion_tokens, |v| {
+            v.min(self.max_completion_tokens)
+        });
+        requested.min(Self::model_max_output_tokens(&self.model))
```
GraDKh: **Floor `max_tokens` at 1 (robustness).**

`effective_max_tokens` is `min(requested, ceiling, model cap)` with no lower bound, so `llm_max_completion_tokens=0` (accepted by the plain `parse::<u32>()` in config) sends `"max_tokens": 0`, which Anthropic rejects with a terminal 400 → every call hard-fails. Add a `.max(1)`. (Python has the same gap via `setdefault("max_tokens", ...)`, so this is a robustness improvement rather than a parity fix, but it's cheap and worth it.)

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+                // 400 stays retryable (Python's tenacity excludes only
+                // 401/402/404), which is also the "model rejected the schema"
+                // repair case.
+                Err(
+                    e @ (LlmError::AuthenticationError(_)
+                    | LlmError::PaymentRequired(_)
+                    | LlmError::ModelNotFound(_)
+                    | LlmError::MaxRetriesExceeded(_)),
+                ) => return Err(e),
+                Err(e) => last_error = e,
```
GraDKh: **400 (schema rejected) retried with an identical body.**

A 400 → `InvalidResponse` lands in the catch-all `Err(e) => last_error = e` here, not the terminal set, and corrective instructions are only appended on the `Ok` arms — so a genuinely-rejected schema is re-POSTed byte-identical `structured_output_retries` times (each after a backoff) before failing. The comment at :477 notes Python's tenacity also retries 400s identically, so this matches Python; but consider either appending a corrective note on the `Err` 400 path too, or treating a schema-validation 400 as terminal, to avoid burning the full retry budget + latency on a request that can't succeed unchanged.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+        }
+    }
+
+    /// Append a corrective instruction to the last user turn so the next attempt
+    /// carries the failure reason, the way instructor reasks with the validation
+    /// error (mirrors `OpenAIAdapter::append_corrective_instruction`).
+    fn append_corrective_instruction(body: &mut Value, reason: Option<&str>) {
+        if let Some(messages) = body["messages"].as_array_mut()
+            && let Some(last) = messages.last_mut()
+            && last["role"] == "user"
```
GraDKh: **Corrective retry silently no-ops on a non-user last turn.**

The `last["role"] == "user"` guard means that when a caller passes messages ending in an assistant turn (via `create_structured_output_with_messages_raw`), a validation/parse failure re-sends an identical body every retry. Low impact — `base_request` guarantees a user turn for system-only input and this is a verbatim copy of the OpenAI guard (`openai.rs:532`) — but worth handling the assistant-last-turn case (append a fresh user turn) when the shared `llm::http` helper is extracted.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+        name = "llm.api_call",
+        level = "info",
+        skip(self, request_body),
+        fields(
+            url = tracing::field::Empty,
+            cognee.llm.model = self.model.as_str(),
+            cognee.llm.provider = "anthropic",
+        ),
+    )]
+    async fn call_api(&self, request_body: Value) -> LlmResult<AnthropicResponse> {
```
GraDKh: **Avoidable per-attempt deep clone (cleanup).**

`call_api` takes `request_body: Value` by value but only ever reads it by reference (`.json(&request_body)`, pretty-print). The structured loop calls `self.call_api(body.clone())` every attempt, deep-cloning the whole JSON tree. Take `&Value` instead. (The OpenAI copy's by-value signature is justified because it mutates the body for the `LLM_ARGS` merge — this adapter doesn't, so there's no reason to keep it here.)

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+    /// Build a schema-aware validator for the type-erased raw path (which has no
+    /// Rust type to deserialize into).
+    ///
+    /// Enforces that every property named in the schema's top-level `required`
+    /// array is present and non-null, so a tool input omitting a required field
+    /// drives a corrective retry instead of being returned as `Ok`. Mirrors the
+    /// OpenAI adapter's validator; folding the two into one shared helper belongs
+    /// with the `llm::http` extraction follow-up rather than churning `openai.rs`
+    /// here.
+    fn schema_required_validator(schema: &Value) -> impl Fn(&Value) -> Result<(), String> + '_ {
```
GraDKh, line 382: **Duplication (cleanup, already flagged in-code).**

`schema_required_validator`, `append_corrective_instruction`, and the `call_api` transient-retry loop are near-verbatim copies of the OpenAI adapter's (`openai.rs:469` / `:532` / `:278`). The doc comment already defers folding these into a shared `llm::http` helper — just flagging that any future fix (e.g. the 402 handling above, backoff changes) must be applied in both places until that lands, and can silently diverge.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+    /// Split messages into the top-level `system` string and the `user`/
+    /// `assistant` message list. Anthropic has no `system` role inside
+    /// `messages`, so all system messages are concatenated into `system`.
+    fn split_messages(messages: &[Message]) -> (String, Vec<Value>) {
+        let mut system_parts: Vec<&str> = Vec::new();
+        let mut converted: Vec<Value> = Vec::new();
+        for m in messages {
+            match m.role {
+                MessageRole::System => system_parts.push(m.content.as_str()),
+                MessageRole::User => {
```
GraDKh: **Collapsible match arms (nit).**

The `User` and `Assistant` arms differ only in the role string literal; map the role to a `&str` and push once (as `openai.rs` does).

crates/components/src/builtins/llm.rs
```diff
@@ -80,6 +80,50 @@ impl LlmFactory for OpenAiCompatibleLlmFactory {
...
+        }
+        let endpoint = if ctx.llm.endpoint.trim().is_empty() {
+            None
+        } else {
+            Some(ctx.llm.endpoint.clone())
+        };
+        let adapter =
+            AnthropicAdapter::new(ctx.llm.model.clone(), ctx.llm.api_key.clone(), endpoint)
+                .map_err(|e| ComponentError::Llm(e.to_string()))?
+                .with_structured_output_retries(ctx.llm.max_retries)
```
GraDKh: **Non-blocking / considered — keep as-is.**

`max_retries` feeds both `with_structured_output_retries` and `with_network_retries`, and the loops nest, so the worst case is `max_retries * (network_retries + 1)` calls. On review this is fine to keep: `MaxRetriesExceeded` is terminal in the outer loop so a *persistently*-failing transport costs only `network_retries + 1`; it matches the existing OpenAI wiring; and it's actually milder than Python (tenacity ~5–6 attempts / 240s × instructor 2 × SDK transport retries). Noting for the record, no change requested.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,832 @@
...
+                );
+                tokio::time::sleep(delay).await;
+            }
+
+            match self.call_api(body.clone()).await {
+                Ok(response) => {
+                    // `stop_reason == "max_tokens"` means the tool input was cut
+                    // off mid-object: present and JSON-parseable, but incomplete.
+                    // Re-ask instead of returning a partial object.
+                    let truncated = response.stop_reason.as_deref() == Some("max_tokens");
```
GraDKh, line 481: **Non-blocking / considered — keep as-is.**

A `stop_reason == "max_tokens"` response is treated as truncated and re-asked before validation, discarding a possibly-complete tool_input. This is a faithful port of instructor's behavior (`function_calls.py` raises `IncompleteOutputException` on `stop_reason == "max_tokens"` before validation), which is what Python's `anthropic_tools` mode runs — so keeping it is the parity-correct choice. Flagging only so it's a conscious decision; no change requested.

GraDKh (changes requested):
## Reconciled code review — native Anthropic Messages API adapter

High-effort automated review plus a second pass cross-checking every finding against the Python cognee reference (`cognee/infrastructure/llm/`). 12 findings survived verification. Requesting changes on the parity divergences and correctness gaps; two findings were **considered and are intentional** (they faithfully match Python's instructor/tenacity behavior) and are flagged non-blocking rather than as change requests.

**Requested changes — parity divergences vs. Python cognee:**
1. `OPENAI_URL` leaks into the Anthropic `base_url` and misroutes all traffic.
2. No `transcribe_image` override → image cognify hard-aborts, though Claude supports vision (Python's Anthropic path does too).
3. `LLM_ARGS` silently dropped — the adapter has no extra-args mechanism at all; Python honors it.
4. New `PaymentRequired` (402) variant is applied only to Anthropic; the OpenAI adapter still retries 402 (Python treats it terminal on both paths).

**Also worth addressing:** floor `max_tokens` at 1 (0 → terminal 400); the 400 retry-with-identical-body path; the corrective-retry no-op on a non-user last turn; and the duplication / per-attempt clone / collapsible match-arm cleanups (the code already flags the shared-helper follow-up).

**Considered, no change requested:** the `max_tokens` truncation reask and the nested retry policy — both faithful ports of Python's behavior.

Per-thread detail inline.

RajdeepKushwaha5:
Thanks for the thorough reconciled pass. All the requested changes are in, plus the cheap correctness/cleanup items.

**Requested changes (parity divergences):**

- **Endpoint leak — fixed.** `AnthropicLlmFactory` no longer inherits `ctx.llm.endpoint` (which aliases `OPENAI_URL`); it passes `None`, so the adapter always targets `https://api.anthropic.com/v1`. A stray `OPENAI_URL` can no longer misroute Anthropic traffic. This matches Python, which builds `AsyncAnthropic(api_key=...)` with no base URL. Documented in `docs/configuration.md`.
- **`transcribe_image` — added.** The adapter now overrides `transcribe_image` (and `supports_vision`) using Anthropic's base64 `image` content block, so an image in a dataset no longer aborts cognify under `LLM_PROVIDER=anthropic`. Built directly (not via `base_request`) so `LLM_ARGS` don't bleed into the description request, matching the OpenAI adapter.
- **`LLM_ARGS` — honored.** Added an `extra_args` field + `with_extra_args`, merged into every Messages body as `{**llm_args, **explicit}` (the adapter's explicit keys win), and wired `.with_extra_args(ctx.llm.llm_args.clone())` in the factory. So operator-configured `LLM_ARGS` (e.g. `top_p`) now take effect. Test added for the merge precedence.
- **402 consistency — fixed.** `402 → PaymentRequired` is now mapped and terminal in the OpenAI adapter's `send_chat_request` (and the transcription path) as well, so a billing-exhausted OpenAI-compatible endpoint fails fast instead of burning the retry budget. The variant is now applied on both paths.

**Also addressed:**

- **Floor `max_tokens` at 1** — `effective_max_tokens` now `.max(1)`, so `llm_max_completion_tokens=0` no longer 400s every call. Test added.
- **Corrective-retry no-op** — `append_corrective_instruction` now appends a new user turn when the last turn is not `user` (was a silent drop). Test added.
- **400 retry-with-identical-body** — the only non-terminal error reaching the outer loop is a raw API 400 (429/5xx/network exhaust to the terminal `MaxRetriesExceeded` inside `call_api`); it now appends the reason before retrying, so the next attempt differs rather than re-POSTing an identical request.
- **Per-attempt deep clone** — `call_api` now takes `&Value`; the structured loop passes `&body` instead of `body.clone()`.
- **Collapsible match arms** — `split_messages` maps the role to a `&str` and pushes once.

**Considered, kept as-is (as you flagged):** the `stop_reason == "max_tokens"` truncation reask (faithful to instructor's `IncompleteOutputException`) and the nested retry policy.

**Follow-up:** the shared `llm::http` core (folding the duplicated `schema_required_validator` / `append_corrective_instruction` / retry skeleton across both adapters) remains the right pre-Tier-3 change; I'll track it as its own issue so it doesn't churn `openai.rs` here.

Verified locally: rustfmt clean; `cargo clippy -p cognee-llm --all-targets -D warnings` clean; the cognee-llm unit tests pass (95, including the four new ones). clippy on the full workspace and the keyless suite run in CI. Rebased onto latest `main`.

crates/llm/src/error.rs
GraDKh, line 38: **⛔ Blocker — mandatory before merge.**

This new variant is correctly terminal/non-retryable here, but it's dropped one layer up. `map_llm_error` in `crates/http-server/src/routers/llm.rs` (~L245–253, not touched by this PR) has:

```rust
ConfigError | ContentPolicyViolation | InvalidResponse => BAD_REQUEST,
_ => ApiError::LlmError(StatusCode::CONFLICT, ...)  // 409
```

There is no arm for `PaymentRequired`, so it falls into the `_` wildcard and returns **HTTP 409 Conflict** — the same bucket as transient/retryable errors (network, rate-limit, 5xx). Retry middleware that treats 409 as a transient conflict will keep re-issuing against a billing-dead account, defeating the precise retry-on-billing behavior this variant was added to stop.

Please add an explicit arm mapping `PaymentRequired` to a terminal status (402 Payment Required is the natural choice) before merge.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,978 @@
...
+            (
+                String::new(),
+                vec![json!({ "role": "user", "content": system })],
+            )
+        } else {
+            (system, converted)
+        };
+        let mut body = json!({
+            "model": self.model,
+            "max_tokens": self.effective_max_tokens(opts),
```
GraDKh, line 353: **Correctness — should fix.**

`max_tokens` is set once here (`effective_max_tokens`) and is never modified anywhere in the structured-output retry loop. When a response comes back with `stop_reason == "max_tokens"`, the truncation branch only appends the "return ONE complete object" corrective (~L413) and re-sends with the **same** `max_tokens`.

If the JSON result inherently needs more than the clamped budget (e.g. a large cognify chunk yielding many entities), every one of the `structured_output_retries` attempts truncates again, sleeps through the backoff, and the call finally fails with `MaxRetriesExceeded` — an extraction a larger budget would have completed never succeeds.

Consider raising `max_tokens` on the truncation-retry path (up to the model cap), or surfacing a distinct non-retryable error, rather than re-asking with an identical budget.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,978 @@
...
+        ("haiku-4", 64_000),
+    ];
+
+    /// Cap applied to models absent from [`MODEL_OUTPUT_CAPS`](Self::MODEL_OUTPUT_CAPS).
+    ///
+    /// Deliberately conservative: forwarding an operator-raised ceiling unclamped
+    /// 400s whenever it exceeds the model's real limit (a terminal error), while
+    /// clamping low only reduces the output budget. It does mean a newly released
+    /// model under-budgets until the table above is refreshed.
+    const UNKNOWN_MODEL_OUTPUT_CAP: u32 = 32_000;
```
GraDKh, line 162: **Cleanup — non-blocking.**

`MODEL_OUTPUT_CAPS` (L137) is a hand-maintained ordered substring table, and any model not listed clamps to `UNKNOWN_MODEL_OUTPUT_CAP = 32_000`. A future model with a larger real output limit will silently under-budget until someone edits the table.

Not blocking — the doc comment already flags sourcing caps from litellm `model_cost` as the durable fix, and the default 16384 ceiling isn't affected. Flagging so the follow-up doesn't get lost.

crates/http-server/src/config.rs
```diff
@@ -669,6 +669,9 @@ impl HttpServerConfig {
                 // The CLI/ComponentManager path wires `LLM_ARGS` via
                 // `cognee_lib::Settings`.
                 llm_args: serde_json::Map::new(),
+                // The HTTP server does not expose an LLM completion-tokens knob
+                // yet; use the same ceiling as the SDK default (Settings).
+                max_completion_tokens: 16384,
```
GraDKh: **Cleanup — non-blocking.**

`16384` is hardcoded here (comment: "same ceiling as the SDK default") and independently duplicated in `crates/lib/src/config.rs`, `AnthropicAdapter::DEFAULT_MAX_COMPLETION_TOKENS` (anthropic.rs:59), and the test fixtures in `builtins/database.rs` / `registry.rs`. If the SDK default ever changes, the HTTP server keeps sending the stale value and silently diverges. A single shared constant would keep them in lockstep.

GraDKh (changes requested):
## Review — native Anthropic Messages API adapter

Nice adapter, and the terminal `PaymentRequired` modeling is the right instinct. One **mandatory blocker** before this can merge, one correctness issue worth fixing, and two non-blocking cleanups. Comments are inline.

**Mandatory before merge:**
- **`PaymentRequired` is silently downgraded to HTTP 409 at the server boundary** (see comment on `crates/llm/src/error.rs`). The variant is terminal by design, but `map_llm_error` in `crates/http-server/src/routers/llm.rs` (untouched by this PR) has no arm for it, so it lands in the `_ => CONFLICT` bucket alongside transient/retryable errors. That re-enables the exact retry-on-billing behavior the variant exists to prevent. Needs an explicit terminal mapping.

**Should fix (correctness):**
- The structured-output truncation re-ask never raises `max_tokens`, so an output that genuinely overflows the budget fails deterministically on every retry. Details inline.

**Non-blocking (cleanup / follow-up):**
- Hand-maintained `MODEL_OUTPUT_CAPS` substring table.
- Hardcoded `16384` duplicated across four sites instead of a shared constant.

---

_For transparency: two additional candidate findings from the first automated pass were investigated and **dropped** after a second verification against this commit — (a) "a 400 is resent with an identical body" is false, the `Err(e)` arm calls `append_corrective_instruction`; and (b) "the corrective is a no-op when the last message is an assistant turn" is false, there's a `_ => push a new user turn` fallback with a pinning unit test. A claimed OpenAI/Anthropic 402-mapping divergence was also stale — this PR maps 402 in both adapters. Not holding the PR on any of those._

RajdeepKushwaha5:
Thanks @GraDKh  for the detailed re-review. All four points are addressed.

### 1. `PaymentRequired` downgraded to 409 at the server boundary (blocker)

Fixed. `map_llm_error` in `crates/http-server/src/routers/llm.rs` now has an explicit terminal arm:

```rust
LlmError::PaymentRequired(_) => {
    ApiError::LlmError(StatusCode::PAYMENT_REQUIRED, err.to_string())
}
```

You were right that the `_ => CONFLICT` wildcard was folding a billing failure into the "transient, retry me" bucket, re-enabling exactly the retry-on-billing loop the adapter layer was careful to exclude. It now surfaces a terminal 402. Added unit tests pinning the mapping: `PaymentRequired` -> 402, transient
(`NetworkError` / `RateLimitExceeded`) -> 409, and the value-error family -> 400, so a future reshuffle of the match can't silently drop the terminal arm back into the wildcard.

### 2. Truncation re-ask never raised `max_tokens` (correctness)

Fixed. On `stop_reason == "max_tokens"` the retry now raises the budget toward the model's documented cap instead of re-asking with the same clamped ceiling:

```rust
Some(_) if truncated => {
    let model_cap = Self::model_max_output_tokens(&self.model);
    let current = body["max_tokens"].as_u64().unwrap_or(0) as u32;
    if current >= model_cap {
        return Err(LlmError::InvalidResponse(format!(
            "Anthropic structured output was truncated at the model's \
             {model_cap}-token output cap and cannot be completed within \
             that budget"
        )));
    }
    body["max_tokens"] = json!(model_cap);
    ...
}
```

This takes both routes you suggested: it raises the budget when there is headroom, and when we are already at the model cap (a larger budget is impossible) it fails terminally with a distinct non-retryable error rather than re-asking at the same ceiling until `MaxRetriesExceeded`.

Two `httpmock` integration tests in `crates/llm/tests/anthropic_truncation_retry.rs` pin the behavior: one asserts the retry arrives with `max_tokens` raised from the 1000-token ceiling to the model cap (64000) and then succeeds; the other asserts
that a truncation already at the model cap fails after a single exchange (no loop).

### 3. `MODEL_OUTPUT_CAPS` hand-maintained table (non-blocking)

Agreed this is the right long-term shape, sourcing the caps from litellm's `model_cost` rather than a hand-maintained substring table. Keeping it as a tracked follow-up so it does not expand the scope of this PR, as discussed.

### 4. `16384` duplicated across sites (non-blocking)

Done. Introduced a single shared constant `cognee_llm::DEFAULT_MAX_COMPLETION_TOKENS` (defined next to `GenerationOptions` in `crates/llm/src/types.rs`) and pointed all of the previously-hardcoded sites at it:

- `GenerationOptions::default().max_tokens`
- `AnthropicAdapter::DEFAULT_MAX_COMPLETION_TOKENS` (now an alias of the shared const)
- `Settings` default in `crates/lib/src/config.rs`
- `HttpServerConfig` LLM inputs in `crates/http-server/src/config.rs`
- the `BackendBuildContext` test fixtures in `crates/components`

They now move in lockstep.

GraDKh:
Thank you for the thorough work on this adapter and the detailed review rounds — really appreciate the care that went into the token-cap handling, structured-output retries, and the Python-parity error mapping. Merging now; I'll follow up with fixes for the review findings (truncation-at-budget handling, token-ceiling wiring, and the adapter dedup) in a separate change. 🙏

PR #41, opened 2026-06-30
feat(llm): Azure OpenAI support (Tier 3)

RajdeepKushwaha5:
## Summary

Tier 3 of #17 -  add Azure OpenAI support. Azure is wire-compatible with the OpenAI chat API, so it is implemented as a thin variant of the existing request path rather than a new adapter. Builds on #30 (Tier 1) and the Tier 2 Anthropic PR.

Azure differs from OpenAI in two ways, both handled on `OpenAIAdapter` behind an optional `api_version`:

- Auth: the `api-key` header instead of `Authorization: Bearer`.
- URL: an `?api-version=<v>` query parameter on every request; the deployment is  encoded in the endpoint URL (`{base}/openai/deployments/{deployment}`), so  `{base_url}/chat/completions?api-version=<v>` is the Azure request URL.

The Azure path is fully gated on `api_version` being set, so the standard OpenAI / OpenAI-compatible path is byte-for-byte unchanged when it is unset. The OpenAI-only request quirks already key off the `api.openai.com` host, so they never fire on Azure.

## Changes

- `OpenAIAdapter`: optional `api_version` field + `with_api_version(..)`; `endpoint_url`  appends the api-version query in Azure mode; `apply_auth` selects `api-key` vs  `Authorization: Bearer`. Empty/whitespace api-version is treated as unset.
- Wire `azure` into `ComponentManager::init_llm_adapter` and the HTTP server's  `wire_llm`. Both require `LLM_API_KEY`, `LLM_ENDPOINT` (the deployment URL), and  `LLM_API_VERSION`.
- Add `llm_api_version` to the HTTP server config (env `LLM_API_VERSION`).
- Docs: `docs/configuration.md` describes the Azure conventions; the `.env.example`  Azure block now uses `LLM_PROVIDER=azure` with the deployment URL.

## Behaviour

No change to existing providers. OpenAI and the Tier 1 OpenAI-compatible providers keep Bearer auth and no api-version query (api_version stays `None`). Azure is the only path that sends `api-key` + `?api-version=`.

## Tests

- New unit tests: the api-version query is appended and `api-key` auth is selected in  Azure mode; the standard OpenAI path keeps Bearer auth and no query; an empty  api-version stays on the OpenAI path.
- The full `cognee-llm` lib suite (78 tests) passes, confirming the shared-adapter  refactor does not regress the OpenAI path.
- Verified: `fmt`, `clippy --all-targets -D warnings` (cognee-llm / cognee-lib /  cognee-http-server).

## Scope

Part of #17 - does not close it. Bedrock (the larger Tier 3 item: SigV4 signing, per-region/model payloads) is deferred per the issue unless there is concrete demand. Stacked on #30 and the Tier 2 PR; until those merge this PR's diff includes their commits, after which GitHub drops the overlap.

### DIFF
--- a/crates/http-server/src/wiring.rs
+++ b/crates/http-server/src/wiring.rs
@@ -279,6 +279,30 @@ fn wire_responses_client(cfg: &HttpServerConfig) -> Option<Arc<dyn ResponsesClie
         return None;
     }
 
+    // The OpenAI Responses client speaks the OpenAI /responses convention
+    // (Authorization: Bearer, no api-version). Allowlist the providers that use
+    // that convention *and* can serve a /responses route: openai (empty means the
+    // openai default), plus the Bearer-auth OpenAI-compatible gateways — LiteLLM
+    // proxy and vLLM (>=0.9) expose an OpenAI-style /responses route and are
+    // typically configured as `custom` / `openai_compatible`. Excluded because
+    // they would 401/404 on every POST /api/v1/responses: Azure (api-key +
+    // api-version), Anthropic (no /responses route), and the ollama / mistral /
+    // gemini gateways (no /responses route). The feature is already opt-in via
+    // `responses_client_enabled` (config.rs), so a `custom` gateway that does not
+    // implement the route only 404s when its operator explicitly enables and
+    // calls it — add/cognify/search are unaffected either way.
+    let provider = cfg.llm_provider.to_ascii_lowercase();
+    if !matches!(
+        provider.as_str(),
+        "openai" | "" | "custom" | "openai_compatible"
+    ) {
+        tracing::info!(
+            "responses client not wired for provider '{provider}' \
+             (its gateway does not serve the OpenAI /responses route)"
+        );
+        return None;
+    }
+
     let api_key = cfg.llm_api_key.expose_secret().to_string();
     if api_key.is_empty() {
         tracing::warn!("responses client enabled but llm api key is missing; wiring as None");
@@ -407,5 +431,42 @@ mod tests {
             msg.contains("postgres connection string") && msg.contains("VECTOR_DB_PROVIDER"),
             "expected actionable pgvector error, got: {msg}"
         );
+    }
+
+    #[test]
+    fn wire_responses_client_allowlist_covers_bearer_compatible_gateways() {
+        use secrecy::SecretString;
+
+        let cfg_for = |provider: &str| HttpServerConfig {
+            llm_provider: provider.to_string(),
+            llm_api_key: SecretString::new("sk-test".to_string().into()),
+            responses_client_enabled: true,
+            ..Default::default()
+        };
+
+        // openai plus the Bearer-auth OpenAI-compatible gateways (LiteLLM / vLLM,
+        // configured as custom / openai_compatible) can serve the /responses route.
+        for provider in ["openai", "", "custom", "openai_compatible"] {
+            assert!(
+                wire_responses_client(&cfg_for(provider)).is_some(),
+                "provider '{provider}' should wire a responses client"
+            );
+        }
+
+        // api-key auth (azure) or no /responses route (anthropic / ollama /
+        // mistral / gemini) stay excluded.
+        for provider in ["azure", "anthropic", "ollama", "mistral", "gemini"] {
+            assert!(
+                wire_responses_client(&cfg_for(provider)).is_none(),
+                "provider '{provider}' must not wire a responses client"
+            );
+        }
+
+        // The opt-in flag still gates an allowlisted provider.
+        let disabled = HttpServerConfig {
+            responses_client_enabled: false,
+            ..cfg_for("openai")
+        };
+        assert!(wire_responses_client(&disabled).is_none());
     }
 }
--- a/crates/llm/src/adapters/openai.rs
+++ b/crates/llm/src/adapters/openai.rs
@@ -49,6 +49,10 @@ pub struct OpenAIAdapter {
     model: String,
     api_key: String,
     base_url: String,
+    /// When `Some`, the adapter targets Azure OpenAI: requests authenticate with
+    /// the `api-key` header (not `Authorization: Bearer`) and carry an
+    /// `?api-version=<v>` query parameter. `None` is the standard OpenAI path.
+    api_version: Option<String>,
     client: Client,
     structured_output_retries: usize,
     /// Number of times to retry the HTTP request on transient network/server errors.
@@ -100,6 +104,79 @@ pub struct OpenAIAdapter {
     /// matching the historical [`GenerationOptions::default`] cap so an adapter
     /// built without config behaves exactly as before.
     default_max_tokens: Option<u32>,
+    /// Whether this adapter's `(model, base_url)` pair is an OpenAI reasoning
+    /// model that needs the reasoning parameter shape. Computed once at
+    /// construction (both inputs are fixed there) so the per-request build path
+    /// does not re-parse `base_url` on every `is_reasoning_model()` call. See
+    /// [`compute_reasoning_model`].
+    reasoning_model: bool,
+}
+
+/// Whether `model` is an OpenAI reasoning family (`gpt-5*`, `o1*`, `o3*`, `o4*`)
+/// served from a host that requires the reasoning parameter shape.
+///
+/// Detection is name-based and host-agnostic for remote hosts, so it fires on
+/// both official OpenAI and Azure deployments (`*.openai.azure.com`) of these
+/// models, which both require the reasoning parameter shape. (A host gate on
+/// `api.openai.com` used to leave Azure o-series/gpt-5 deployments sending
+/// `max_tokens`+`temperature`, which Azure 400s on every call.) It is suppressed
+/// only for a local host (see [`is_local_base_url`]) — loopback, or Ollama's
+/// default port on a private-network address — which keeps accepting the legacy
+/// parameters even when a served model name collides with a reasoning prefix.
+///
+/// A self-hosted gateway on a private network but a non-Ollama port (e.g. a LAN
+/// vLLM at `http://192.168.1.5:8000`) is treated as remote and gets the
+/// reasoning shape; such a host is often a proxy to real OpenAI, where that
+/// shape is correct. A deployment that needs the opposite can point the model at
+/// a loopback bind or use Ollama's port.
+fn compute_reasoning_model(model: &str, base_url: &str) -> bool {
+    let m = model.to_lowercase();
+    let is_reasoning_name =
+        m.starts_with("gpt-5") || m.starts_with("o1") || m.starts_with("o3") || m.starts_with("o4");
+    is_reasoning_name && !is_local_base_url(base_url)
+}
+
+/// Heuristic for a local / non-OpenAI-compatible host that does not accept the
+/// reasoning-model parameter shape.
+///
+/// Matches on the parsed host authority, not a substring scan of the whole URL,
+/// so a genuinely remote endpoint whose URL merely contains `localhost` as a
+/// subdomain (`o3.localhost.example.com`) or `127.0.0.1` in a path/query is not
+/// misclassified as local. Local means either:
+/// - a loopback host (`localhost`, `127.0.0.0/8`, `::1`), any port; or
+/// - Ollama's default port `11434` on a private-network host (loopback or an
+///   RFC-1918 / link-local address). The port shortcut is gated on a private
+///   host so a genuinely remote endpoint that merely listens on `11434`
+///   (`https://gateway.example.com:11434`) is not classified local and still
+///   gets the reasoning shape for a real o-series/gpt-5 model.
+fn is_local_base_url(base_url: &str) -> bool {
+    let Ok(url) = reqwest::Url::parse(base_url) else {
+        // An unparseable base_url can't be classified as local; treat it as
+        // remote so a reasoning model still gets the correct parameter shape.
+        return false;
+    };
+    let Some(host) = url.host_str() else {
+        return false;
+    };
+    // url keeps brackets on IPv6 literals; strip them so `[::1]` parses as `::1`.
+    let host = host.trim_start_matches('[').trim_end_matches(']');
+    if host.eq_ignore_ascii_case("localhost") {
+        return true;
+    }
+    let Ok(ip) = host.parse::<std::net::IpAddr>() else {
+        // A named remote host (not the `localhost` label) is not local.
+        return false;
+    };
+    if ip.is_loopback() {
+        return true;
+    }
+    let is_private = match ip {
+        std::net::IpAddr::V4(v4) => v4.is_private() || v4.is_link_local(),
+        // Unique-local (`fc00::/7`) detection is not stable on IpAddr; loopback
+        // (handled above) covers the realistic local-IPv6 case.
+        std::net::IpAddr::V6(_) => false,
+    };
+    is_private && url.port() == Some(11434)
 }
 
 impl OpenAIAdapter {
@@ -150,22 +227,30 @@ impl OpenAIAdapter {
         // that legitimately contain a slash (e.g. Baseten's `openai/gpt-oss-120b`).
         let model: String = model.into();
 
+        // Normalise a trailing slash so request URLs built as
+        // `{base_url}/chat/completions` never produce a double slash. The
+        // Gemini OpenAI-compat base ends in `/v1beta/openai/`, and a
+        // user-supplied endpoint may too; both would otherwise 404.
+        let base_url = base_url
+            .map(|u| u.trim_end_matches('/').to_string())
+            .unwrap_or_else(|| Self::DEFAULT_BASE_URL.to_string());
+
+        // Both `model` and `base_url` are fixed for the adapter's lifetime, so
+        // classify once here instead of re-parsing `base_url` on every request.
+        let reasoning_model = compute_reasoning_model(&model, &base_url);
+
         Ok(Self {
             model,
             api_key: api_key.into(),
-            // Normalise a trailing slash so request URLs built as
-            // `{base_url}/chat/completions` never produce a double slash. The
-            // Gemini OpenAI-compat base ends in `/v1beta/openai/`, and a
-            // user-supplied endpoint may too; both would otherwise 404.
-            base_url: base_url
-                .map(|u| u.trim_end_matches('/').to_string())
-                .unwrap_or_else(|| Self::DEFAULT_BASE_URL.to_string()),
+            base_url,
+            api_version: None,
             client,
             structured_output_retries: Self::DEFAULT_STRUCTURED_OUTPUT_RETRIES,
             network_retries: Self::DEFAULT_NETWORK_RETRIES,
             transcription_model,
             extra_args: serde_json::Map::new(),
             default_max_tokens: Some(Self::DEFAULT_MAX_COMPLETION_TOKENS),
+            reasoning_model,
         })
     }
 
@@ -269,6 +354,59 @@ impl OpenAIAdapter {
         format!("Bearer {}", self.api_key)
     }
 
+    /// Enable Azure OpenAI mode by setting the API version. An empty/whitespace
+    /// value is treated as unset (stays on the standard OpenAI path). In Azure
+    /// mode the `base_url` is expected to be the deployment endpoint
+    /// (`https://<resource>.openai.azure.com/openai/deployments/<deployment>`),
+    /// so `{base_url}/chat/completions?api-version=<v>` is the Azure request URL.
+    pub fn with_api_version(mut self, api_version: impl Into<String>) -> Self {
+        let v = api_version.into();
+        let trimmed = v.trim();
+        // Store trimmed: `endpoint_url` interpolates this raw, so trailing
+        // whitespace would percent-encode to `?api-version=...%20` and Azure
+        // 400s every request. (Matches the embedding path's api-version handling.)
+        self.api_version = if trimmed.is_empty() {
+            None
+        } else {
+            Some(trimmed.to_string())
+        };
+        self
+    }
+
+    /// Build a request URL for `path`, appending `api-version=<v>` in Azure mode.
+    ///
+    /// The api-version is added through `Url::query_pairs_mut`, so the query
+    /// separator is chosen correctly even if `base_url` already carries a query
+    /// (no malformed double-`?`) and the value is percent-encoded rather than
+    /// interpolated raw. If `base_url` does not parse as a URL we fall back to a
+    /// manual append that still picks `?` vs `&` from the existing query, so the
+    /// api-version is never silently dropped (Azure 400s without it).
+    fn endpoint_url(&self, path: &str) -> String {
+        let base = format!("{}/{path}", self.base_url);
+        match &self.api_version {
+            Some(v) => match reqwest::Url::parse(&base) {
+                Ok(mut url) => {
+                    url.query_pairs_mut().append_pair("api-version", v);
+                    url.into()
+                }
+                Err(_) => {
+                    let sep = if base.contains('?') { '&' } else { '?' };
+                    format!("{base}{sep}api-version={v}")
+                }
+            },
+            None => base,
+        }
+    }
+
+    /// Apply the provider's auth header: `api-key` for Azure, `Authorization:
+    /// Bearer` for standard OpenAI / OpenAI-compatible endpoints.
+    fn apply_auth(&self, req: reqwest::RequestBuilder) -> reqwest::RequestBuilder {
+        match &self.api_version {
+            Some(_) => req.header("api-key", &self.api_key),
+            None => req.header("Authorization", self.auth_header()),
+        }
+    }
+
     /// Whether to request non-thinking mode for local Qwen OpenAI-compatible endpoints.
     fn should_disable_thinking(&self) -> bool {
         self.model.to_lowercase().starts_with("qwen") && !self.base_url.contains("api.openai.com")
@@ -278,15 +416,10 @@ impl OpenAIAdapter {
     /// that reject `temperature`/`top_p`/`frequency_penalty`/`presence_penalty`
     /// overrides and require `max_completion_tokens` in place of `max_tokens`.
     ///
-    /// Gated on the official `api.openai.com` base URL so custom OpenAI-compatible
-    /// proxies (Ollama, vLLM, …) keep accepting legacy parameters even when the
-    /// configured model name happens to match a reasoning-family prefix.
+    /// Returns the value classified once at construction (see
+    /// [`compute_reasoning_model`]); no per-call URL parsing.
     fn is_reasoning_model(&self) -> bool {
-        if !self.base_url.contains("api.openai.com") {
-            return false;
-        }
-        let m = self.model.to_lowercase();
-        m.starts_with("gpt-5") || m.starts_with("o1") || m.starts_with("o3") || m.starts_with("o4")
+        self.reasoning_model
     }
 
     /// Insert `max_tokens` (or `max_completion_tokens` on reasoning models) into a
@@ -337,7 +470,7 @@ impl OpenAIAdapter {
         ),
     )]
     async fn send_chat_request(&self, request_body: Value) -> LlmResult<OpenAIResponse> {
-        let url = format!("{}/chat/completions", self.base_url);
+        let url = self.endpoint_url("chat/completions");
         tracing::Span::current().record("url", url.as_str());
         let debug_enabled = std::env::var("COGNEE_DEBUG_LLM_REQUEST")
             .map(|v| cognee_utils::parse_env_bool(&v))
@@ -366,9 +499,7 @@ impl OpenAIAdapter {
             }
 
             let response = match self
-                .client
-                .post(&url)
-                .header("Authorization", self.auth_header())
+                .apply_auth(self.client.post(&url))
                 .header("Content-Type", "application/json")
                 .json(&request_body)
                 .send()
@@ -1232,7 +1363,7 @@ impl OpenAIAdapter {
         &self,
         form: reqwest::multipart::Form,
     ) -> LlmResult<WhisperResponse> {
-        let url = format!("{}/audio/transcriptions", self.base_url);
+        let url = self.endpoint_url("audio/transcriptions");
         tracing::Span::current().record("url", url.as_str());
 
         // We cannot clone a multipart Form, so the first attempt uses the
@@ -1252,9 +1383,7 @@ impl OpenAIAdapter {
         // the form on each attempt.
 
         let response = self
-            .client
-            .post(&url)
-            .header("Authorization", self.auth_header())
+            .apply_auth(self.client.post(&url))
             .multipart(form)
             .send()
             .await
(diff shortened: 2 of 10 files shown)


RajdeepKushwaha5:
Stacked on #30 - the `Test (stable)` failure is inherited from there (see my note on #30). Everything that doesn't need a live key passes locally; happy to target the specific failing test if the log can be shared.

GraDKh:
@RajdeepKushwaha5 could you please rebase this one on the latest main

RajdeepKushwaha5:
> @RajdeepKushwaha5 could you please rebase this one on the latest main

done

GraDKh (commented):
**Heads-up on rebasing: `main` has changed substantially under this PR.**

This PR is stacked on the unmerged Tier-1 (#30) and Tier-2 branches, so its diff currently includes a multi-provider `crates/llm/src/factory.rs` with a `SUPPORTED_PROVIDERS` constant and expanded `wire_llm`/`init_llm` arms (ollama/mistral/gemini/custom/openai_compatible). None of that exists on today's `main`: there is no `factory.rs`, `ComponentManager::init_llm` supports only `openai` + `mock` (litert now errors to a closed crate), and `wire_llm` hard-gates to `openai`. A few of the comments below (especially the provider-list one) are framed against the stacked base; once #30/Tier-2 land or this rebases, please re-verify those spots against the consolidated single-implementation routing in `component_manager.rs`, so the divergence doesn't silently reintroduce the duplicate provider lists.

_Automated review — findings verified against the PR head; flag any false positives._

crates/llm/src/adapters/openai.rs
```diff
@@ -146,6 +151,34 @@ impl OpenAIAdapter {
...
+    pub fn with_api_version(mut self, api_version: impl Into<String>) -> Self {
+        let v = api_version.into();
+        self.api_version = if v.trim().is_empty() { None } else { Some(v) };
+        self
+    }
+
+    /// Build a request URL for `path`, appending `?api-version=<v>` in Azure mode.
+    fn endpoint_url(&self, path: &str) -> String {
+        match &self.api_version {
+            Some(v) => format!("{}/{path}?api-version={v}", self.base_url),
```
GraDKh: **Azure deployments of reasoning models (o1/o3/o4/gpt-5) will fail on every call.**

This comment anchors on the Azure request path, but the actual issue is in `is_reasoning_model()` a few lines below: it early-returns `false` unless `base_url` contains `api.openai.com`, so on an Azure base URL (`*.openai.azure.com`) it is never true. An Azure deployment of an o-series/gpt-5 model therefore gets `max_tokens` (not `max_completion_tokens`) and keeps `temperature` — and since `GenerationOptions::default()` sets `temperature: Some(0.0)` and `max_tokens: Some(16384)`, this fires on *every* call. Azure returns HTTP 400 ("Unsupported parameter"), which `call_api` treats as non-retryable, so the deployment is unusable.

Rather than just adding `*.openai.azure.com` to the host check (which patches Azure but breaks again for the next OpenAI-compatible gateway), consider making the detection **name-based**: match the o-series/gpt-5 model name and only *suppress* it for known local/non-compatible hosts (Ollama/vLLM). That removes the whole bug class. FWIW the `genai` crate detects reasoning models exactly this way — by model name, host-agnostic.

crates/lib/src/component_manager.rs
```diff
@@ -609,6 +613,60 @@ impl ComponentManager {
...
+                    ));
+                }
+                let endpoint = if llm_endpoint.trim().is_empty() {
+                    None
+                } else {
+                    Some(llm_endpoint.clone())
+                };
+                let retries = llm_max_retries.max(1);
+                let adapter =
+                    AnthropicAdapter::new(llm_model.clone(), llm_api_key.clone(), endpoint)
```
GraDKh: **The Anthropic adapter is hard-capped at 4096 output tokens, which can fail cognify runs.**

`AnthropicAdapter::with_max_tokens` is defined but never called from either wiring layer (here or `crates/http-server/src/wiring.rs`), so every request falls back to `DEFAULT_MAX_TOKENS = 4096` via `opts.max_tokens.unwrap_or(self.default_max_tokens)`.

There is a pre-existing gap that is not unique to this PR: `LLM_MAX_TOKENS` / `llm_max_completion_tokens` is not wired into *any* provider's generation path. But on the OpenAI path that's harmless — `write_max_tokens` omits the field entirely when `None` (uncapped). The Messages API *requires* `max_tokens`, so for this new adapter `None` becomes a hard, unconfigurable 4096 cap.

That collides with cognify, which deliberately passes `max_tokens: None` ("no artificial cap" — see `fact_extraction/extractor.rs` and `temporal_extraction/entity_enricher.rs`). On a dense chunk whose extraction exceeds 4096 output tokens, the response truncates (`stop_reason: max_tokens`), the forced `tool_use` block is incomplete, `tool_input()` returns `None`, and the identical request is retried until `MaxRetriesExceeded` fails the run.

Suggest wiring `llm_max_completion_tokens` into `.with_max_tokens(...)` in both wiring layers (and adding a max-tokens field to `HttpServerConfig`, which currently has none), or at least raising `DEFAULT_MAX_TOKENS`.

crates/http-server/src/wiring.rs
```diff
@@ -271,6 +272,66 @@ async fn wire_embedding_engine(cfg: &HttpServerConfig) -> Option<Arc<dyn Embeddi
...
+                tracing::warn!("anthropic llm not wired: {err}");
+                None
+            }
+        };
+    }
+
+    // Azure OpenAI reuses the OpenAI request path with api-key auth and an
+    // api-version query (issue #17, Tier 3); LLM_ENDPOINT must be the deployment
+    // URL and LLM_API_VERSION must be set.
+    if cfg.llm_provider.eq_ignore_ascii_case("azure") {
```
GraDKh: **The OpenAI responses client is auto-wired against Azure/Anthropic endpoints and will 401 on every request.**

(Anchored on the azure block; the issue is in the separate `wire_responses_client` function.) It is enabled by default whenever `LLM_API_KEY` is non-empty (`config.rs` — opt-out only via `COGNEE_RESPONSES_CLIENT_ENABLED=false`), and it ignores `llm_provider`/`llm_api_version`: it reuses `llm_api_key`/`llm_endpoint` with `Authorization: Bearer` and no `?api-version`.

So under this PR's own documented `LLM_PROVIDER=azure` recipe (with `COGNEE_RESPONSES_CLIENT_ENABLED` unset), the server wires an `OpenAIResponsesClient` that POSTs to `{azure-deployment}/responses` with Bearer auth and no api-version → 401/404 on every `POST /api/v1/responses`. Same for `anthropic` (targets `api.openai.com` with an `sk-ant-` key, or an Anthropic endpoint with no `/responses` route).

Newly reachable: pre-PR the factory rejected `azure`/`anthropic`, so there was no working config to hit this. The blast radius is limited to the `/api/v1/responses` surface (add/cognify/search are unaffected), but it fails silently — wiring logs success. Suggest gating `wire_responses_client` on an OpenAI-compatible provider, or teaching it the Azure api-key/api-version conventions.

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,472 @@
...
+                    None => {
+                        last_error = LlmError::InvalidResponse(
+                            "Anthropic response did not contain the forced tool_use block"
+                                .to_string(),
+                        );
+                    }
+                },
+                // call_api already returns terminal errors (400/401/404) without
+                // retrying; surface those immediately rather than burning attempts.
+                Err(e @ (LlmError::AuthenticationError(_) | LlmError::ModelNotFound(_))) => {
```
GraDKh: **The structured-output retry loop re-sends errors that retrying can't fix.**

1. A deterministic HTTP 400 (mapped to `LlmError::InvalidResponse` in `call_api`) is not caught by this short-circuit, which only matches `AuthenticationError | ModelNotFound`. It falls into the generic arm below, so the identical body is re-sent up to `DEFAULT_STRUCTURED_OUTPUT_RETRIES` (5) times — directly contradicting the comment right above that says 400 is surfaced immediately. A schema Anthropic rejects burns 5 identical requests and then reports `MaxRetriesExceeded`, masking the real 400.

2. `call_api` already runs its own network-retry ladder and returns `MaxRetriesExceeded` on exhaustion; that also lands in the generic arm, so the outer loop re-enters `call_api`. Worst case is 5 × (network_retries+1) = 20 HTTP attempts with three separate backoff ladders on a persistently failing endpoint. The OpenAI loops avoid this by propagating `call_api` errors with `?` and retrying only validation failures.

Fix subtlety: `InvalidResponse` is overloaded — it's also the legitimately-retryable "response didn't contain the forced tool_use block" case. So propagate `call_api` errors (or at least `MaxRetriesExceeded` + the 400 origin) out of the loop and reserve the outer retries for the missing-tool_use case, e.g. via a distinct error variant rather than matching on `InvalidResponse` broadly.

crates/llm/src/adapters/openai.rs
```diff
@@ -146,6 +151,34 @@ impl OpenAIAdapter {
...
     }
 
+    /// Enable Azure OpenAI mode by setting the API version. An empty/whitespace
+    /// value is treated as unset (stays on the standard OpenAI path). In Azure
+    /// mode the `base_url` is expected to be the deployment endpoint
+    /// (`https://<resource>.openai.azure.com/openai/deployments/<deployment>`),
+    /// so `{base_url}/chat/completions?api-version=<v>` is the Azure request URL.
+    pub fn with_api_version(mut self, api_version: impl Into<String>) -> Self {
+        let v = api_version.into();
+        self.api_version = if v.trim().is_empty() { None } else { Some(v) };
```
GraDKh: **nit:** `with_api_version` validates the trimmed value but stores it untrimmed (`self.api_version = if v.trim().is_empty() { None } else { Some(v) }`), and `endpoint_url` interpolates it raw. A value with trailing whitespace — `LLM_API_VERSION="2024-12-01-preview "` — passes the empty-check, gets percent-encoded to `...?api-version=...%20`, and Azure 400s every request with only a `%20` in the traced URL as a clue.

Low severity: dotenv trims unquoted `.env` values, so the common path is safe; whitespace only survives via quoted `.env` values, shell exports, docker-compose `env_file`, or the settings.json/HTTP/bindings surfaces. Trivial fix: store `Some(v.trim().to_string())`. (The embedding path already does `val.trim().to_string()` for `EMBEDDING_API_VERSION`, so this also aligns the two.)

crates/llm/src/adapters/anthropic.rs
```diff
@@ -0,0 +1,472 @@
...
+        let mut out = schema.clone();
+        if let Some(obj) = out.as_object_mut() {
+            obj.remove("$schema");
+        }
+        out
+    }
+
+    /// POST `request_body` to `/messages` with the standard headers and a
+    /// transient-retry loop with exponential backoff.
+    async fn call_api(&self, request_body: Value) -> LlmResult<AnthropicResponse> {
```
GraDKh: **`call_api` drops the `llm.api_call` tracing span, making Anthropic traffic invisible to telemetry.**

This `call_api` is a near-verbatim copy of `OpenAIAdapter::call_api`, but it omits the `#[instrument(name = "llm.api_call", …, cognee.llm.model = …, cognee.llm.provider = "openai")]` attribute the OpenAI version carries. That span is a tested contract — `crates/llm/tests/openai_span_instrumentation.rs` asserts on a `llm.api_call` span with `cognee.llm.model`/`cognee.llm.provider`, and those field names are workspace constants in `crates/utils/src/tracing_keys.rs`. Any subscriber or dashboard keyed on `llm.api_call` will record zero Anthropic calls and can't tell providers apart.

Suggest adding the matching `#[instrument]` with `cognee.llm.provider = "anthropic"` (and ideally a mirrored span test).

crates/http-server/src/wiring.rs
```diff
@@ -271,6 +272,66 @@ async fn wire_embedding_engine(cfg: &HttpServerConfig) -> Option<Arc<dyn Embeddi
 }
 
 fn wire_llm(cfg: &HttpServerConfig) -> Option<Arc<dyn Llm>> {
+    // Native Anthropic uses its own Messages-API adapter (issue #17, Tier 2),
+    // not the OpenAI-compatible factory.
+    if cfg.llm_provider.eq_ignore_ascii_case("anthropic") {
```
GraDKh: **The anthropic and azure wiring is duplicated across both wiring layers, bypassing the factory built to prevent exactly this.**

These ~55 lines are near-identical to the copy in `crates/lib/src/component_manager.rs` (same empty-key check, empty-endpoint→`None` mapping, `retries.max(1)`, builder chain). `crates/llm/src/factory.rs` is the designed home for provider routing — its doc comment says "Centralising that here keeps the two wiring paths in sync," and its error contract already anticipates the hard-fail (component manager) vs warn-and-`None` (HTTP server) split. The two copies have already diverged in error handling.

Azure fits the factory cleanly: it produces an `OpenAIAdapter` and differs from the `openai` arm by a single `.with_api_version()` call. Anthropic is more constrained — it's a different concrete type, so the current `-> LlmResult<OpenAIAdapter>` signature can't return it (transcriber wiring depends on the concrete type); it'd need a sibling `build_llm(...) -> LlmResult<Arc<dyn Llm>>` entry point.

Cost: every future change to these providers (default endpoint, retry floor, wiring `LLM_MAX_TOKENS`) has to land in both places or the embedded SDK and HTTP server silently diverge. (See also the rebase note — on `main` this consolidation looks different.)

crates/lib/src/component_manager.rs
```diff
@@ -619,7 +677,7 @@ impl ComponentManager {
             _ => {
                 return Err(ComponentError::Config(format!(
                     "Unsupported llm_provider '{provider}'. \
-                     Supported: openai, ollama, mistral, gemini, custom, mock.",
+                     Supported: openai, ollama, mistral, gemini, custom, anthropic, azure, mock.",
```
GraDKh: **Keep the supported-provider list in one place — and on `main` that place has moved.**

This PR (via its stacked Tier-1 base) also updates a `SUPPORTED_PROVIDERS` list in `factory.rs`, so there are now two hand-maintained lists that must agree. But note that `factory.rs` doesn't exist on current `main` — provider routing there is consolidated into this single implementation, `ComponentManager::init_llm`, whose fallthrough arm is the only canonical list (on `main` it reads `"Supported: openai, mock."`). When this rebases onto `main`, please make sure `azure` (and `anthropic` from Tier-2) are added to *this* list rather than reintroducing a second one in a resurrected `factory.rs` — otherwise the two drift, which is exactly the failure being flagged.

Also worth deciding: the standalone `wire_llm` path (`crates/http-server/src/wiring.rs`) currently hard-gates to `openai` and warns "not supported by standalone wiring yet" for everything else — so azure/anthropic won't be reachable via the HTTP server's standalone wiring unless that gate is updated too.

GraDKh (changes requested):
Requesting changes. Detailed findings are in the inline comments on this PR (review [#4629841948](https://github.com/topoteretes/cognee-rs/pull/41#pullrequestreview-4629841948)). The blocking ones:

- **Azure reasoning models fail on every call** — `is_reasoning_model()` is gated on the `api.openai.com` host, so Azure o-series/gpt-5 deployments send `max_tokens`/`temperature` and get a non-retryable 400.
- **Anthropic is hard-capped at 4096 output tokens** — `with_max_tokens` is never wired, so cognify's `max_tokens: None` becomes a 4096 cap and truncated extractions fail the run.
- **The responses client 401s under azure/anthropic** — it's auto-wired with Bearer auth / no api-version against the configured endpoint.

Also: the Anthropic structured-output retry loop re-sends unfixable errors, plus lower-severity items (api-version trimming, the missing `llm.api_call` span, duplicated wiring, and the provider-list single-source question). See the per-line comments for specifics and suggested fixes, and the earlier review body for the rebase heads-up (`main` has diverged from this PR's stacked base).

RajdeepKushwaha5:
Thanks, this was a sharp review. I rebased this azure-only onto latest main (the #56 registry), so several of the stacked-base points resolve on their own. Point by point:

**Blocking:**
- **Azure reasoning models: fixed.** `is_reasoning_model` is now name-based and host-agnostic, so Azure o-series/gpt-5 deployments get `max_completion_tokens` (not `max_tokens`) and drop `temperature`, instead of a 400 on every call. It is suppressed only for local / non-OpenAI-compatible hosts (loopback, the Ollama port) that reject the shape. Added a test covering an Azure deployment and a remote gateway.
- **Responses client 401s: fixed.** `wire_responses_client` now skips `azure` and `anthropic`, since the OpenAI responses client speaks only the Bearer / no-api-version convention. Add/cognify/search were unaffected either way.
- **Anthropic 4096 cap:** out of scope here now. Rebasing azure-only dropped the anthropic commit from this PR; the max_tokens clamp is fixed in #40.

**Non-blocking:**
- **api-version trimming: fixed.** Stored trimmed, so a trailing space in `LLM_API_VERSION` no longer percent-encodes into the query and 400s. Aligns with the embedding path.
- **Duplicated wiring: gone.** Azure is now an `AzureLlmFactory` registered in the #56 component registry: it builds the shared `OpenAIAdapter` against the explicit deployment endpoint, then applies `.with_api_version()`. One wiring path, not two. `api_version` is threaded through `LlmInputs`.
- **Provider-list single source:** moot on current main. #56 consolidated routing into the registry, so there is no `factory.rs` provider list to drift against; Azure registers itself with the registry.
- **Anthropic `call_api` tracing span and the anthropic 400 retry loop:** both Anthropic, so out of scope on this azure-only PR. The 400 retry is actually faithful to Python (you confirmed this on #40, where I fixed the misleading comment); the missing tracing span I will add on #40.

Verified: fmt, `clippy --all-targets -D warnings` (llm / components / lib / http-server), and the `cognee-llm` tests (74 lib + integration, including the api-version and reasoning-model cases). Rebased onto latest main.

crates/llm/src/adapters/openai.rs
```diff
@@ -155,15 +196,31 @@ impl OpenAIAdapter {
...
+        let is_reasoning_name = m.starts_with("gpt-5")
+            || m.starts_with("o1")
+            || m.starts_with("o3")
+            || m.starts_with("o4");
+        is_reasoning_name && !self.is_local_host()
+    }
+
+    /// Heuristic for a local / non-OpenAI-compatible host (loopback or the Ollama
+    /// default port) that does not accept the reasoning-model parameter shape.
+    fn is_local_host(&self) -> bool {
```
GraDKh: **🟠 correctness — `is_local_host` substring matching misclassifies genuinely remote endpoints**

`contains("localhost") / contains("127.0.0.1") / contains(":11434")` matches anywhere in the URL, so a **remote** host is treated as local when its URL merely contains one of these substrings:

- `https://gw.corp:11434/v1` — a remote server exposed on the Ollama port
- `https://o3.localhost.example.com/...` — `localhost` as a subdomain label
- any path/query containing `127.0.0.1`

When such a remote endpoint serves a reasoning model, `is_reasoning_model()` wrongly returns `false`, so the adapter sends `max_tokens`+`temperature` to a reasoning deployment and Azure/OpenAI 400s ("Unsupported parameter") on every call.

Match on the parsed host authority (and check loopback exactly / port `== 11434`) rather than substring-scanning the whole `base_url`. Lower severity — these are unusual URLs — but the substring match is the more concrete of the two `is_reasoning_model` heuristic concerns.

crates/http-server/src/wiring.rs
```diff
@@ -279,6 +279,21 @@ fn wire_responses_client(cfg: &HttpServerConfig) -> Option<Arc<dyn ResponsesClie
...
     }
 
+    // The OpenAI Responses client speaks the OpenAI /responses convention
+    // (Authorization: Bearer, no api-version). Azure needs api-key auth + an
+    // api-version query, and Anthropic has no /responses route, so wiring it for
+    // those providers would 401/404 on every POST /api/v1/responses. Skip them
+    // rather than wire a client that cannot work (add/cognify/search are
+    // unaffected either way).
+    let provider = cfg.llm_provider.to_ascii_lowercase();
+    if matches!(provider.as_str(), "azure" | "anthropic") {
```
GraDKh: **🟡 cleanup — `wire_responses_client` uses a blocklist, so non-OpenAI providers with no `/responses` route are still wired to a 404**

The prior review asked to gate this on an OpenAI-compatible provider (an allowlist). This blocklist (`matches!(provider, "azure" | "anthropic")`) fixes only those two and reintroduces the same class for the rest: `OPENAI_COMPATIBLE_PROVIDERS` (builtins/llm.rs) also includes `ollama`, `mistral`, `gemini`, `custom`, `openai_compatible` — none of which expose the OpenAI **Responses** route. With `LLM_PROVIDER=ollama` (etc.) and the responses client enabled (default when `LLM_API_KEY` is set), the server still wires an `OpenAIResponsesClient` that POSTs to `{endpoint}/responses`, so every `POST /api/v1/responses` 404s while wiring logs success.

Note "OpenAI-compatible for chat completions" ≠ "has a `/responses` route" — that surface is OpenAI-specific. An allowlist of the providers that actually serve `/responses` (effectively `openai`, plus any proxy you've verified) is the robust gate; the blocklist has to be updated for every future non-OpenAI provider.

GraDKh (changes requested):
Thanks for the Azure adapter and the clean rebase onto main — two inline items remain (the `is_local_host` substring heuristic and the `wire_responses_client` blocklist that still wires ollama/mistral/gemini to a 404 `/responses` route) before this is good to merge.

RajdeepKushwaha5:
Thanks, I addressed both points (rebased onto latest main first):

- **`is_local_host` substring matching:** now parses the base URL with `reqwest::Url` and matches on the authority, not a substring scan of the whole string. It suppresses the reasoning-model shape only for an exact loopback host (`localhost` / `127.0.0.1` / `::1`) or the parsed Ollama port `11434`. So `o3.localhost.example.com` (loopback as a subdomain) and a `127.0.0.1` in a path/query are correctly treated as remote and still get `max_completion_tokens`. Added regression tests for the subdomain-`localhost`, `127.0.0.1`-in-path, and real-loopback cases.
- **`wire_responses_client` blocklist:** switched to an allowlist. The OpenAI `/responses` route is OpenAI-specific ("OpenAI-compatible for chat completions" does not imply a `/responses` endpoint), so the client is now wired only for `openai` (empty provider means the openai default). That stops ollama / mistral / gemini / custom / openai_compatible from being wired to a `/responses` route that 404s, not just azure/anthropic. `add`/`cognify`/`search` are unaffected.

Verified: rustfmt clean, and the updated openai unit tests (the reasoning-model detection cases, including the new remote-host regressions). clippy and the full test run go through CI.

crates/llm/src/adapters/openai.rs
```diff
@@ -204,15 +245,46 @@ impl OpenAIAdapter {
...
-        if !self.base_url.contains("api.openai.com") {
-            return false;
-        }
         let m = self.model.to_lowercase();
-        m.starts_with("gpt-5") || m.starts_with("o1") || m.starts_with("o3") || m.starts_with("o4")
+        let is_reasoning_name = m.starts_with("gpt-5")
+            || m.starts_with("o1")
+            || m.starts_with("o3")
+            || m.starts_with("o4");
+        is_reasoning_name && !self.is_local_host()
```
GraDKh: Making reasoning detection host-agnostic to fix Azure 400s makes sense, and the doc comment lays out the rationale well. One edge the new logic opens up: suppression now relies solely on `is_local_host()` (loopback + port 11434). A **non-loopback** self-hosted gateway on a custom port — e.g. a LAN vLLM/LM Studio at `http://192.168.1.5:8000` — that serves a model whose name matches an `o1`/`o3`/`o4`/`gpt-5` prefix now gets the reasoning shape (`max_completion_tokens`, no `temperature`/`top_p`) where it previously got the legacy shape. Truly-local setups are covered by loopback, so this is a narrow case (and arguably such a gateway is usually a proxy to real OpenAI, where the reasoning shape is correct) — but the doc comment's "a local vLLM" only holds for loopback binds. Worth a note, or gate reasoning-shape on an explicit provider/config signal rather than inferring from host+name.

crates/http-server/src/wiring.rs
```diff
@@ -279,6 +279,24 @@ fn wire_responses_client(cfg: &HttpServerConfig) -> Option<Arc<dyn ResponsesClie
...
+    // (Authorization: Bearer, no api-version, an OpenAI-hosted /responses route).
+    // That route is OpenAI-specific: "OpenAI-compatible for chat completions" does
+    // not imply a /responses endpoint. Azure (api-key + api-version), Anthropic
+    // (no /responses route), and the ollama/mistral/gemini/custom gateways would
+    // all 401/404 on every POST /api/v1/responses. So allowlist the providers that
+    // actually serve /responses (openai; empty means the openai default) rather
+    // than blocklisting a known-bad few and silently 404ing the rest —
+    // add/cognify/search are unaffected either way.
+    let provider = cfg.llm_provider.to_ascii_lowercase();
+    if !matches!(provider.as_str(), "openai" | "") {
```
GraDKh: The rationale here is sound for Azure/Anthropic/ollama, but the allowlist is stricter than "only OpenAI serves /responses" in practice: **LiteLLM proxy and vLLM (>=0.9) both expose an OpenAI-style `/responses` route**, and those are typically configured as `openai_compatible`/`custom`. For such a deployment that previously had `responses_client_enabled` (which defaults to "api key present", config.rs:555) and a working `POST /api/v1/responses`, this guard now returns `None` and the route 500s (`"responses client is not wired"`, responses.rs). Since the feature is already opt-in via `responses_client_enabled`, consider treating that flag as the opt-in (or adding an explicit override) rather than a hard provider allowlist — that keeps the Azure/Anthropic safety without silently regressing compatible gateways that do implement the route.

crates/llm/src/adapters/openai.rs
```diff
@@ -204,15 +245,46 @@ impl OpenAIAdapter {
...
+    fn is_local_host(&self) -> bool {
+        match reqwest::Url::parse(&self.base_url) {
+            Ok(url) => {
+                let host_is_loopback = url.host_str().is_some_and(|h| {
+                    // url keeps brackets on IPv6 literals; strip them so `[::1]`
+                    // compares equal to `::1`.
+                    let h = h.trim_start_matches('[').trim_end_matches(']');
+                    h.eq_ignore_ascii_case("localhost") || h == "127.0.0.1" || h == "::1"
+                });
+                host_is_loopback || url.port() == Some(11434)
```
GraDKh: `url.port() == Some(11434)` is ORed unconditionally, so it also matches a *genuinely remote* endpoint that happens to listen on 11434 (e.g. `https://gateway.example.com:11434/v1`), classifying it local and suppressing the reasoning shape for a real o-series/gpt-5 model there. Very narrow (only triggers on an explicit `:11434`), but if the intent is "local Ollama", consider requiring the host to also be loopback/private before the port shortcut applies.

crates/llm/src/adapters/openai.rs
```diff
@@ -195,6 +200,42 @@ impl OpenAIAdapter {
...
+        } else {
+            Some(trimmed.to_string())
+        };
+        self
+    }
+
+    /// Build a request URL for `path`, appending `?api-version=<v>` in Azure mode.
+    fn endpoint_url(&self, path: &str) -> String {
+        match &self.api_version {
+            Some(v) => format!("{}/{path}?api-version={v}", self.base_url),
```
GraDKh: Two small robustness gaps in the Azure URL build:
1. The `?` is hardcoded, so if `base_url` already carries a query component the result is a malformed double-`?` (`...?existing=x?api-version=v`). The docs steer users to a plain deployment URL, but pasting the portal URL verbatim is an easy mistake — choosing the separator on `base_url.contains('?')`, or building via `Url` + `query_pairs_mut()`, would harden it.
2. `api_version` is trimmed (good) but not percent-encoded before interpolation. Realistic values are safe, but an unexpected char would go in raw.

crates/llm/src/adapters/openai.rs
```diff
@@ -204,15 +245,46 @@ impl OpenAIAdapter {
...
+
+    /// Heuristic for a local / non-OpenAI-compatible host (loopback or the Ollama
+    /// default port) that does not accept the reasoning-model parameter shape.
+    ///
+    /// Matches on the parsed host authority (exact loopback host, or the Ollama
+    /// default port 11434), not a substring scan of the whole URL, so a genuinely
+    /// remote endpoint whose URL merely contains `localhost` as a subdomain
+    /// (`o3.localhost.example.com`) or `127.0.0.1` in a path/query is not
+    /// misclassified as local.
+    fn is_local_host(&self) -> bool {
```
GraDKh: Minor (not blocking): `is_local_host()` runs `reqwest::Url::parse(&self.base_url)` on every call, and it's called on each `is_reasoning_model()` — a couple of times per request build (more across structured-output retries/fallbacks). Both `base_url` and `model` are fixed at construction, so the local-host (or whole reasoning) classification could be computed once in the builder and stored as a `bool`. Cheap either way — parsing a short constant — just avoidable repeated work on the request path.

GraDKh (commented):
Reviewed the Azure adapter changes (high-effort pass, findings independently re-verified against the diff). The Azure support itself looks solid — `api-key` vs `Bearer`, `api-version`, and the deployment-URL shape all hang together. I left 5 inline notes, all on side effects of the reasoning-detection rework and a couple of robustness edges. None is a merge blocker in my view (the two biggest are documented, intentional decisions), so this is a comment rather than a change request — flagging them so you can decide.

**Summary**
1. `is_reasoning_model` host gate removal — new LAN-gateway edge (openai.rs:262)
2. `/responses` provider allowlist — regresses openai_compatible gateways that *do* serve /responses, e.g. LiteLLM/vLLM (wiring.rs:292)
3. `is_local_host` port-11434 check fires on remote hosts too (openai.rs:282)
4. `endpoint_url` can build a double-`?` URL + api_version not percent-encoded (openai.rs:225)
5. minor: `is_local_host` re-parses base_url on every call (openai.rs:273)

RajdeepKushwaha5:
Thanks @GraDKh  for the high-effort pass. Addressed all five, code and tests pushed.

### 1. `is_reasoning_model` host gate — LAN-gateway edge (openai.rs)

Reworked the local-host detection so the doc comment and the behavior line up.
Local now means:

- a loopback host (`localhost`, `127.0.0.0/8`, `::1`), any port; or
- Ollama's default port `11434` on a private-network host (RFC-1918 / link-local).

A self-hosted gateway on a private network but a non-Ollama port (your `http://192.168.1.5:8000` LAN vLLM example) is treated as remote and gets the reasoning shape. As you noted, such a host is often a proxy to real OpenAI where that shape is correct, so I kept it as remote rather than guessing, and the doc comment now says so explicitly instead of implying "any local vLLM." A deployment that wants the opposite can bind loopback or use Ollama's port; if a stronger signal is ever needed, an explicit provider/config flag is the durable route, noted as a follow-up.

### 2. `/responses` provider allowlist — regressed compatible gateways (wiring.rs)

Fixed. The allowlist now includes the Bearer-auth OpenAI-compatible providers that can serve an OpenAI-style `/responses` route:

```rust
matches!(provider.as_str(), "openai" | "" | "custom" | "openai_compatible")
```

That restores LiteLLM proxy and vLLM (>=0.9), which are configured as `custom` / `openai_compatible`, while still excluding Azure (api-key + api-version), Anthropic (no `/responses`), and the ollama / mistral / gemini gateways. Since the feature is already opt-in via `responses_client_enabled`, a `custom` gateway that does not implement the route only 404s when its operator explicitly enables and calls it, and add/cognify/search are unaffected.

### 3. `is_local_host` port-11434 check fired on remote hosts (openai.rs)

Fixed. The `11434` shortcut is now gated on the host being private/loopback, so `https://gateway.example.com:11434` is classified remote and a real o-series/gpt-5 model there gets the reasoning shape. Added tests for remote-on-11434 (remote), LAN Ollama `192.168.x:11434` (local), and LAN gateway `192.168.x:8000` (remote).

### 4. `endpoint_url` double-`?` + api_version not percent-encoded (openai.rs)

Fixed via `Url::query_pairs_mut`, which chooses the query separator correctly (no malformed double-`?` when `base_url` already carries a query) and percent-encodes the value instead of interpolating it raw. If `base_url` does not parse as a URL, the manual fallback still picks `?` vs `&` from the existing query so the api-version is never dropped (Azure 400s without it). Added a test covering the double-query and reserved-character cases.

### 5. `is_local_host` re-parsed `base_url` on every call (openai.rs)

Fixed. `(model, base_url)` are fixed at construction, so the reasoning classification is computed once in the constructor and stored as a `bool`; `is_reasoning_model()` now just returns it, with no per-request URL parsing on the structured-output retry/fallback paths.

Issue #22, opened 2026-06-26
Perf umbrella: optimize external API & I/O usage (LLM, embeddings, DB) for latency


GraDKh:
## Goal

Reduce end-to-end **latency and wall-clock time** of the cognee-rs pipeline by optimizing how it uses **external / I/O-bound resources** — LLM APIs, embedding models, and databases — without changing output correctness (90%+ parity target with the Python SDK preserved).

This is the umbrella tracking the investigation. Each slice has a dedicated issue with verified findings, file/line anchors, and acceptance criteria. Complements #18 (pure-CPU pipeline profiling) — that issue covers CPU-bound work; these cover I/O latency.

## Sub-issues

- [x] #19 — **LLM API calls**: batch pipelining (`buffer_unordered` instead of serial batch drains), bounded summarization concurrency, prompt caching of the static prefix, retry jitter.
- [x] #20 — **Embeddings**: eliminate the confirmed **double-embedding pass** (chunks/entities/summaries are embedded twice per `cognify`), Ollama batch input, batch-size tuning.
- [ ] #21 — **Databases**: two confirmed N+1 loops (`has_edges`, `update_last_accessed`), transactional grouping of bulk writes, connection-pool config, native batch search.

## Highest-confidence wins (start here)

| # | Win | Impact |
|---|-----|--------|
| #20 | Stop embedding chunks/entities/summaries twice per `cognify` (first pass is used only for a `.len()` count, never persisted) | ~2× fewer embedding calls for those types |
| #21 | `has_edges` — one query per edge → single `unnest` query | O(E) → O(1) round-trips on bulk edge ingest |
| #21 | `update_last_accessed` — fetch+update per id → bulk `UPDATE ... WHERE IN` | 2N → 1 round-trip |
| #19 | Pipeline extraction/summarization batches instead of draining each serially | removes inter-batch stalls |

## Cross-cutting acceptance criteria

- [ ] A repeatable benchmark for a representative large-document `add → cognify → search` run capturing: wall-clock, LLM call count, embedding call count, DB round-trip count, and p50/p95 per-stage latency. Reuse the #18 mock-replay harness where it applies.
- [ ] Before/after numbers recorded for each implemented optimization.
- [ ] No regression in cross-SDK parity tests.

## Related prior work (Linear)

COG-4160 (relational DB grouping, Python), COG-5064 (provider RPM/latency under parallel load), COG-3146 (configurable cognify batch size), COG-2522 (async LLM call utilization).

PR #43, opened 2026-06-30
perf(cognify): bound summarization concurrency and add retry jitter

RajdeepKushwaha5:
## Summary

First, measurable slice of #19. This PR lands the two LLM-call optimizations that improve correctness/robustness on their own and do not depend on a wall-clock benchmark to justify:

1. **Bound summarization concurrency.** Summarization spawned one `tokio::spawn` per chunk with no ceiling, so a large document could open one in-flight LLM request per chunk , a rate-limit and memory risk, and unlike graph extraction
   which is already bounded. `summarize_chunks` now runs a bounded  `buffer_unordered(max_parallel)` pipeline (default 20, matching the graph extractor's `max_parallel_extractions`), wired from `CognifyConfig`. Output order and fail-on-first-error semantics are preserved.

2. **Retry jitter.** The adapter's network backoff was deterministic (1s, 2s, 4s …), so a batch of requests that all hit HTTP 429 at the same instant also retried in lockstep, a thundering herd that keeps tripping the limit. Backoff now uses equal jitter in `[base/2, base]` via a shared `retry_backoff` helper.

Both changes are correctness/robustness wins: they reduce the chance of *causing* rate-limit stalls, which is the real-life bottleneck #19 calls out.

## What is intentionally not in this PR

The larger latency items in #19 , removing inter-batch barriers for graph extraction (`buffer_unordered` over all chunks), multi-chunk-per-request batching, and prompt caching , are worth doing but only land credibly *with measurements*. The maintainer has rightly noted that neither an instant `ReplayLlm` benchmark nor a low-sample live-API run represents the real-life high-load case where rate limits are the bottleneck. I propose to land those behind a deterministic load harness (see the issue comment) so each change ships with before/after numbers under a rate-limited regime, rather than guesses.

## Tests

- `retry_backoff` unit tests: stays within the equal-jitter bounds for every attempt, caps at 30s, and is actually randomized. (Also fixed a latent multiply overflow at very large attempt counts, found by the cap test.)
- `summarize_chunks` concurrency test: a mock LLM records peak in-flight calls; with 10 chunks and a cap of 3 the peak never exceeds 3, all 10 are summarized, and the output preserves input order.
- `fmt`, `clippy --all-targets -D warnings`.

## Relation to the LLM-adapter PRs (#30, #40, #41)

This continues the LLM-layer work from the #17 tiers: provider routing (#30), the native Anthropic adapter (#40), and Azure (#41). The retry jitter lands in the same `OpenAIAdapter` request loop those PRs hardened, and the prompt-caching follow-up
maps directly onto the native Anthropic adapter from #40 (Anthropic prompt caching is set on the `system`/tools blocks the adapter already builds). So the LLM-call optimization in #19 is a natural continuation of that adapter work, by the same
author who owns those code paths.

## Scope

Part of #19 (the LLM slice of the #22 external-API performance umbrella). Does not close it: the measured pipelining/batching/caching follow-ups remain.

### DIFF
--- a/crates/cognify/src/summarization/extractor.rs
+++ b/crates/cognify/src/summarization/extractor.rs
@@ -33,6 +33,41 @@ fn default_summary_options() -> GenerationOptions {
 /// drift guard.
 const DEFAULT_SUMMARY_PROMPT: &str = include_str!("prompts/summarize_content.txt");
 
+/// Summarize a single chunk of text. Shared by [`SummaryExtractor::extract_summary`]
+/// and the bounded [`SummaryExtractor::summarize_chunks`] pipeline so neither has
+/// to fabricate a throwaway extractor per call.
+async fn summarize_one(
+    llm: &Arc<dyn Llm>,
+    summary_schema: &Option<serde_json::Value>,
+    text: &str,
+    custom_prompt: Option<&str>,
+) -> Result<SummarizedContent, CognifyError> {
+    let system_prompt = custom_prompt.unwrap_or(DEFAULT_SUMMARY_PROMPT);
+    let options = Some(default_summary_options());
+
+    match summary_schema {
+        None => llm
+            .create_structured_output(text, system_prompt, options)
+            .await
+            .map_err(|e| CognifyError::LlmError(e.to_string())),
+        Some(schema) => {
+            let raw: serde_json::Value = llm
+                .create_structured_output_raw(text, system_prompt, schema, options)
+                .await
+                .map_err(|e| CognifyError::LlmError(e.to_string()))?;
+            let summary = raw.get("summary").and_then(|v| v.as_str()).ok_or_else(|| {
+                CognifyError::LlmError(
+                    "summary_schema output missing string `summary` field".to_string(),
+                )
+            })?;
+            Ok(SummarizedContent {
+                summary: summary.to_string(),
+                description: String::new(),
+            })
+        }
+    }
+}
+
 /// Summary extractor for text chunks.
 ///
 /// Uses an LLM (via the Llm trait) to generate hierarchical summaries from text chunks.
@@ -108,35 +143,7 @@ impl SummaryExtractor {
         text: &str,
         custom_prompt: Option<&str>,
     ) -> Result<SummarizedContent, CognifyError> {
-        let system_prompt = custom_prompt.unwrap_or(DEFAULT_SUMMARY_PROMPT);
-        let options = Some(default_summary_options());
-
-        match &self.summary_schema {
-            None => {
-                let summarized: SummarizedContent = self
-                    .llm
-                    .create_structured_output(text, system_prompt, options)
-                    .await
-                    .map_err(|e| CognifyError::LlmError(e.to_string()))?;
-                Ok(summarized)
-            }
-            Some(schema) => {
-                let raw: serde_json::Value = self
-                    .llm
-                    .create_structured_output_raw(text, system_prompt, schema, options)
-                    .await
-                    .map_err(|e| CognifyError::LlmError(e.to_string()))?;
-                let summary = raw.get("summary").and_then(|v| v.as_str()).ok_or_else(|| {
-                    CognifyError::LlmError(
-                        "summary_schema output missing string `summary` field".to_string(),
-                    )
-                })?;
-                Ok(SummarizedContent {
-                    summary: summary.to_string(),
-                    description: String::new(),
-                })
-            }
-        }
+        summarize_one(&self.llm, &self.summary_schema, text, custom_prompt).await
     }
 
     /// Summarize multiple text chunks in parallel.
@@ -187,12 +194,8 @@ impl SummaryExtractor {
                 let model_name = model_name.clone();
                 let prompt = custom_prompt.clone();
                 async move {
-                    let extractor = SummaryExtractor {
-                        llm,
-                        summary_schema,
-                        max_parallel: 1,
-                    };
-                    let summarized = extractor.extract_summary(&text, prompt.as_deref()).await?;
+                    let summarized =
+                        summarize_one(&llm, &summary_schema, &text, prompt.as_deref()).await?;
                     let summary =
                         TextSummary::from_summarized_content(chunk_id, summarized, model_name);
                     Ok::<(usize, TextSummary), CognifyError>((index, summary))
--- a/crates/cognify/src/summarization/extractor.rs
+++ b/crates/cognify/src/summarization/extractor.rs
@@ -58,14 +58,23 @@ pub struct SummaryExtractor {
     /// When `Some`, the dynamic schema path is taken instead of the typed
     /// `SummarizedContent` path (Python `summarization_model` parity).
     summary_schema: Option<serde_json::Value>,
+    /// Maximum number of concurrent summarization LLM calls. Bounds in-flight
+    /// requests on large documents (see issue #19).
+    max_parallel: usize,
 }
 
 impl SummaryExtractor {
+    /// Default cap on concurrent summarization LLM calls. Matches the graph
+    /// extractor's default `max_parallel_extractions` so both stages share the
+    /// same in-flight ceiling.
+    pub const DEFAULT_MAX_PARALLEL: usize = 20;
+
     /// Create a new summary extractor using the built-in `SummarizedContent` schema.
     pub fn new(llm: Arc<dyn Llm>) -> Self {
         Self {
             llm,
             summary_schema: None,
+            max_parallel: Self::DEFAULT_MAX_PARALLEL,
         }
     }
 
@@ -78,7 +87,15 @@ impl SummaryExtractor {
         Self {
             llm,
             summary_schema: schema,
+            max_parallel: Self::DEFAULT_MAX_PARALLEL,
         }
+    }
+
+    /// Set the maximum number of concurrent summarization LLM calls. Values
+    /// below 1 are coerced to 1.
+    pub fn with_max_parallel(mut self, max_parallel: usize) -> Self {
+        self.max_parallel = max_parallel.max(1);
+        self
     }
 
     /// Extract a summary from text.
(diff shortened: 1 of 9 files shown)


GraDKh (commented):
Reviewed the two optimizations — both are correct and well-tested: output order is preserved via the index-sort, the retry jitter bounds / overflow / 30s cap are all correct, and `buffer_unordered`'s drop-on-error actually improves on the old detached-`tokio::spawn` behavior (which kept firing already-cancelled requests). No correctness bugs found. The inline comments below are consistency/cleanup follow-ups.

**One gap worth fixing before merge — noted here because it's outside the diff:** the retry-jitter change was wired into only 2 of the 3 identical retry sites. `crates/llm/src/responses_client.rs:215` still uses the deterministic backoff `(1_000u64 * 2u64.saturating_pow(attempt as u32 - 1)).min(30_000)`. A batch of Responses-API requests that all hit HTTP 429 at the same instant will still retry in lockstep — the exact thundering herd this PR set out to fix, left un-fixed on one path. Route that site through `crate::retry::retry_backoff` as well so all three loops share the jittered backoff.

crates/llm/src/retry.rs
```diff
@@ -0,0 +1,75 @@
...
+        .min(30_000)
+}
+
+/// Exponential backoff with **equal jitter** for retry `attempt` (1-indexed).
+///
+/// Returns a duration in `[base/2, base]`, where `base` is the capped
+/// exponential backoff. Keeping at least half the backoff preserves the growing
+/// delay, while the random half spreads simultaneous retries to avoid a
+/// thundering herd (e.g. a batch that all hit HTTP 429 at once).
+pub(crate) fn retry_backoff(attempt: u32) -> Duration {
```
GraDKh, line 37: **Reuse instead of a second retry module.** `cognee_utils::retry` (crates/utils/src/retry.rs) already implements capped exponential backoff **with jitter** — `RetryConfig` + `calculate_delay` + `with_jitter`. Adding a second backoff module inside the `llm` crate means two implementations of the same math that will drift over time. Strongly prefer either building this equal-jitter variant on top of `cognee_utils::retry::RetryConfig::calculate_delay`, or extracting the shared delay computation into that one module so both crates use a single source of truth.

crates/llm/src/retry.rs
```diff
@@ -0,0 +1,75 @@
...
+//! exponential backoff. A purely deterministic schedule (1s, 2s, 4s, …) means a
+//! batch of requests that all hit a rate limit at the same instant also retry at
+//! the same instants — a thundering herd that keeps tripping the limit. Adding
+//! jitter spreads those retries out. See issue #19.
+
+use std::time::Duration;
+
+/// Capped exponential backoff base for a 1-indexed `attempt`: 1s, 2s, 4s, …
+/// capped at 30s.
+fn base_backoff_ms(attempt: u32) -> u64 {
```
GraDKh, line 19: **Latent footgun:** `base_backoff_ms(0)` returns ~1000ms rather than 0, because `saturating_sub(1)` floors the exponent at `2^0`. It's only safe today because every caller guards `if attempt > 0`. A `debug_assert!(attempt >= 1)` (or a doc line stating `attempt` is 1-indexed) would prevent a future `retry_backoff(0)` from silently sleeping ~[500,1000]ms on a nominal 0th attempt.

crates/cognify/src/summarization/extractor.rs
```diff
@@ -142,45 +159,51 @@ impl SummaryExtractor {
...
+                        summary_schema,
+                        max_parallel: 1,
+                    };
+                    let summarized = extractor.extract_summary(&text, prompt.as_deref()).await?;
+                    let summary =
+                        TextSummary::from_summarized_content(chunk_id, summarized, model_name);
+                    Ok::<(usize, TextSummary), CognifyError>((index, summary))
+                }
+            })
+            .buffer_unordered(self.max_parallel)
```
GraDKh, line 204: **Concurrency idiom consistency.** The graph-extraction path bounds parallelism with `Arc<Semaphore>` + `tokio::spawn` + `join_all` (see `tasks.rs:389 / 771 / 1190`), whereas this introduces a second idiom — `stream::iter(...).buffer_unordered(...)` — for the identical need: "run N LLM calls, cap at `max_parallel_extractions`". Two competing bounded-fan-out patterns in the same area add maintenance friction (anyone changing concurrency semantics must reason about both). Consider a single shared helper, or adopting the existing semaphore idiom here.

crates/cognify/src/summarization/extractor.rs
```diff
@@ -142,45 +159,51 @@ impl SummaryExtractor {
...
+            .map(|(index, text, chunk_id)| {
+                let llm = Arc::clone(&self.llm);
+                let summary_schema = self.summary_schema.clone();
+                let model_name = model_name.clone();
+                let prompt = custom_prompt.clone();
+                async move {
+                    let extractor = SummaryExtractor {
+                        llm,
+                        summary_schema,
+                        max_parallel: 1,
```
GraDKh: **Awkward per-chunk construction.** Each future rebuilds a full `SummaryExtractor { llm, summary_schema, max_parallel: 1 }` just to call the `&self` method `extract_summary`, and the `max_parallel: 1` field is never read on that throwaway instance. Unlike the old `tokio::spawn` version, `buffer_unordered` futures don't need `'static`, so the closure can capture `&self.llm` / `&self.summary_schema` (or call a small free fn) directly instead of fabricating a nested extractor whose parallelism value is meaningless.

crates/cognify/src/tasks.rs
```diff
@@ -882,7 +882,8 @@ pub async fn summarize_text(
 
     let summaries = if config.enable_summarization && !non_dlt_chunks.is_empty() {
         let summary_extractor =
-            SummaryExtractor::new_with_schema(llm, config.summary_schema.clone());
+            SummaryExtractor::new_with_schema(llm, config.summary_schema.clone())
+                .with_max_parallel(config.max_parallel_extractions);
         let mut all_summaries = Vec::new();
 
         for batch in non_dlt_chunks.chunks(config.summarization_batch_size) {
```
GraDKh: **Double-bounding after this change.** `summarize_chunks` now internally caps concurrency at `max_parallel` (20) via `buffer_unordered`, so this outer `chunks(summarization_batch_size)` (50) loop no longer bounds in-flight requests — it only inserts a sequential barrier every 50 chunks (the fastest 49 summaries wait for the slowest before the next 50 start), reducing throughput vs. streaming all chunks through one `buffer_unordered`. Either drop the batch loop and pass all `non_dlt_chunks` to a single `summarize_chunks` call, or add a comment explaining why both a batch size (50) and a parallelism cap (20) are needed — otherwise `summarization_batch_size` now reads as a near-dead knob.

GraDKh (changes requested):
Thanks for the careful, well-tested work here — the concurrency bound and the jittered backoff are both genuinely correct, and the accompanying tests are a pleasure to review. Requesting changes only for the one gap that matters before merge: the retry-jitter fix should also cover `crates/llm/src/responses_client.rs:215`, which still retries in lockstep. The other inline comments are non-blocking cleanup/consistency suggestions — take or leave as you see fit.

RajdeepKushwaha5:
Thanks @GraDKh for the thorough review. I have pushed a follow-up commit addressing everything:

Blocking - 3rd retry site: Good catch. `responses_client.rs` now routes through `crate::retry::retry_backoff ` too, so all three transient-retry loops share the jittered backoff. No path retries in lockstep anymore.

Single source of truth for the backoff math: `retry_backoff` now derives its capped-exponential base from `cognee_utils::retry::RetryConfig::calculate_delay` (made `pub`) and only layers the equal jitter `[base/2, base]` on top - no second copy of the exponential/cap math. This is your first suggested option ("build the equal-jitter variant on top of `calculate_delay`"), and it keeps the existing 30s cap and tested bounds since the base values are identical.

`base_backoff_ms(0)` footgun: added `debug_assert!(attempt >= 1)` and a doc line stating `attempt` is 1-indexed.

Throwaway per-chunk `SummaryExtractor`: extracted a shared free fn `summarize_one(llm, schema, text, prompt)`; `extract_summary` delegates to it and `summarize_chunks` calls it directly, so there's no fabricated `max_parallel`: 1 extractor per future.

Double-bounding: dropped the outer batch loop - all chunks now stream through a single `summarize_chunks` call bounded internally at `max_parallel_extractions`, removing the every-N-chunks sequential barrier.

Concurrency idiom (semaphore vs `buffer_unordered`): left as `buffer_unordered` here since, as you noted, its drop-on-error behavior is an improvement over the detached `tokio::spawn`. Unifying the graph-extraction path onto the same idiom (or a shared helper) is a worthwhile but larger change I'd rather do as a focused follow-up than fold into this PR.

Verified: `fmt`, `clippy --all-targets -D warnings` (utils/llm/cognify), plus the retry, summarization-concurrency, and `cognee_utils` retry tests.

RajdeepKushwaha5:
hi sir @GraDKh 

i have opened a follow-up issue for unifying the two bounded-fan-out idioms so that thread has a concrete home: https://github.com/topoteretes/cognee-rs/issues/49
It captures the four sites (three `Semaphore` + `spawn` + `join_all`, one `buffer_unordered`) and proposes either a
shared `bounded_fan_out` helper or standardizing on `buffer_unordered`, with the behavioural differences (drop-on-error, `'static`, parallelism-vs-concurrency, ordering) written down so the decision is made once. Keeping it out of this PR to keep the diff focused.

RajdeepKushwaha5:
Hi Sir @GraDKh 

I addressed the last review loose end: after dropping the summarization batch loop, `summarization_batch_size` was read by nothing, so I removed it (field, builder, validation, tests) rather than leave a knob that silently does nothing. Concurrency is bounded solely by `max_parallel_extractions` via the `summarize_chunks` pipeline. Also rebased onto latest main (was 9 behind, now conflict-free). Local `fmt` + `clippy --all-targets -D warnings` clean.

crates/utils/src/retry.rs
```diff
@@ -59,8 +59,12 @@ impl RetryConfig {
...
-    /// Calculate the delay for a given attempt number.
-    fn calculate_delay(&self, attempt: u32) -> Duration {
+    /// Calculate the delay for a given attempt number (0-indexed).
+    ///
+    /// Public so other crates can reuse this single capped-exponential-backoff
+    /// implementation rather than re-deriving the math (see `cognee-llm`'s
+    /// `retry_backoff`).
+    pub fn calculate_delay(&self, attempt: u32) -> Duration {
         let base_delay =
             self.initial_delay_ms as f64 * self.backoff_multiplier.powi(attempt as i32);
```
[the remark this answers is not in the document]
GraDKh: Thanks, Copilot — good catch on the narrowing cast. In this PR's call paths it isn't actually reachable: `attempt` is always bounded by `network_retries` (a small config value), so it never approaches `i32::MAX` and the exponent can't wrap to a negative value. The delay therefore keeps growing and caps at 30s as intended.

That said, the point is fair now that `calculate_delay` is `pub` — a future external caller could pass an arbitrary `attempt`, so clamping the exponent before the `as i32` cast is worthwhile defensive hardening for the newly-public API. Capturing it as a non-blocking follow-up nit rather than a blocker for this PR. Resolving this thread.

GraDKh (approved):
Thanks for the great work on this, @RajdeepKushwaha5 🙏

This is a clean, well-scoped solution. Both changes are genuine robustness wins and the implementation is careful:

- **Bounded summarization concurrency** — replacing the unbounded per-chunk `tokio::spawn` with a `buffer_unordered(max_parallel)` pipeline is exactly the right shape. Output order is preserved via the index sort, the concurrency ceiling can never be zero (constructors default to `DEFAULT_MAX_PARALLEL = 20` and `with_max_parallel` clamps `<1 → 1`), and error propagation still surfaces on the first failing chunk. Reusing `max_parallel_extractions` so summarization and graph extraction share one ceiling is a nice touch.
- **Retry jitter** — the equal-jitter `[base/2, base]` backoff is correctly bounded and stays under the 30s cap, and delegating the capped-exponential math to `cognee_utils::retry::RetryConfig` keeps a single source of truth. Also a good catch fixing the latent multiply overflow.
- **Tests** — the concurrency-tracker mock (peak in-flight ≤ cap, order preserved, all chunks summarized) and the jitter bounds/cap/randomization tests directly exercise the new behavior. Removing the now-dead `summarization_batch_size` config cleanly (no lingering references) rounds it out.

I ran an independent multi-angle review (correctness, removed-behavior, cross-file, plus adversarial verification) and no correctness findings survived. The only open item is Copilot's defensive-hardening nit on the now-`pub` `calculate_delay` cast, which I've resolved as a non-blocking follow-up.

Approving. Nice, disciplined slice of #19. 🚀

PR #51, opened 2026-07-02
fix(search): fail loudly when NATURAL_LANGUAGE is unsupported by the backend

GraDKh:
## Problem

`search({ searchType: 'NATURAL_LANGUAGE' })` (and the `NATURAL_LANGUAGE` search type generally) returns an empty result with **no error** against the default local **ladybug** graph backend — even on a fully populated graph where other search types work.

## Root cause

`NaturalLanguageRetriever` asks the LLM to generate **Neo4j-style Cypher against typed node labels** (`:Entity`, `:TextDocument`, …; hardcoded in `NL_SYSTEM_PROMPT_TEMPLATE`). But ladybug/kuzu stores **every node under a single generic `:Node` label** (real kind in a `type` property) and every edge as `:EDGE`. So each generated query is rejected by the binder (`Table Entity does not exist`), the 3-attempt retry loop exhausts, and it returns `[]` — indistinguishable from "no results found".

This is a **faithful port of a pre-existing Python limitation** — I verified it end-to-end against Python `cognee` 1.2.2 on ladybug (populated graph, `GRAPH_COMPLETION` returns correct answers, `NATURAL_LANGUAGE` returns empty; generated Cypher used `:Entity` and hit `Table Entity does not exist` on all attempts). NATURAL_LANGUAGE/CYPHER are effectively **Neo4j-only** features — only Python's Neo4j adapter adds dynamic typed labels matching the prompt.

## Fix (fail loudly)

`execute_nl_query` now tracks whether any generated query actually executed against the graph. When **every** query that reached the backend was rejected and none executed, it returns a clear `SearchError::InvalidInput` explaining the `:Node`-vs-typed-label mismatch and pointing to Neo4j / an alternative search type — instead of a misleading empty result.

- A pure **LLM-generation failure** (no query ever reaches the graph) still returns an empty result → historical contract preserved (existing test still green).
- **CYPHER** search already surfaces backend errors via `?`, so it's unchanged.

## Tests

`cargo test -p cognee-search` — new `natural_language_retriever_errors_when_backend_rejects_all_queries` plus the 5 existing retriever tests all pass. `cargo check --all-targets`, `cargo clippy -p cognee-search --all-targets`, and `cargo fmt` are clean.

## Follow-up

The **proper fix** — teaching NATURAL_LANGUAGE to introspect the real `n.type` / `r.relationship_name` values and generate ladybug-shaped Cypher against the generic `:Node`/`:EDGE` model — is tracked in a separate issue.

### DIFF
--- a/crates/search/src/retrievers/cypher_nl_retrievers.rs
+++ b/crates/search/src/retrievers/cypher_nl_retrievers.rs
@@ -242,6 +242,11 @@ impl NaturalLanguageRetriever {
     async fn execute_nl_query(&self, query: &str) -> Result<Vec<Vec<Value>>, SearchError> {
         let (_node_schemas, edge_schemas) = self.get_graph_schema().await?;
         let mut previous_attempts = String::from("No attempts yet.");
+        // Track whether any generated query actually executed against the graph
+        // (Ok, even with zero rows) versus every one being rejected. See the
+        // fail-loudly branch after the loop.
+        let mut any_query_executed = false;
+        let mut last_query_error: Option<String> = None;
 
         for _ in 0..self.max_attempts {
             let cypher_query = match self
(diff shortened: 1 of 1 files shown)


GraDKh:
The proper ladybug-aware fix (generate Cypher against the generic `:Node`/`:EDGE` model instead of hardcoded Neo4j labels) is tracked in #52.

PR #56, opened 2026-07-02
refactor: extract cognee-components + pluggable adapter registry

GraDKh:
## What & why

The `ComponentManager` (cognee-lib) and the HTTP server's standalone `wiring.rs` each built the same seven pipeline backends from config, by hand — and had **drifted**: http-server lacked lancedb/brute-force vector backends, pggraph, mock-llm, `MOCK_EMBEDDING`, and had weaker sqlite parent-dir handling. This extracts the construction logic into a new **`cognee-components`** crate behind a pluggable **`ComponentRegistry`**, so both callers share one implementation and the drift is resolved by construction.

The registry is also the **extension seam** for the closed `cognee-cloud-rs` repo: external adapters (qdrant, litert) implement a factory trait and `register_*` it at their binary entry point — no OSS edits, no `match` arms to touch.

## Design

- New crate `crates/components`:
  - `ComponentError` (re-exported by `cognee-lib`, so the type is unchanged for downstream `#[from]`/`From` impls)
  - `BackendBuildContext` — the resolved, env-free input both configs lower into (all URL resolution + env reads happen at that boundary)
  - factory traits `VectorDbFactory` / `GraphDbFactory` / `LlmFactory` / `EmbeddingFactory` (`#[async_trait] + Send + Sync` so `Arc<dyn _>` is dyn-compatible and holders stay `Sync`)
  - `ComponentRegistry` (provider-id → factory) with `with_builtins()` + `register_*` + a single `build_*` path; free `build_storage` / `build_database`
  - drift-guard + sqlite-URL-shape tests
- Migration (backward-compatible on the OSS side):
  - `ComponentManager` gains a registry field; `new()` = `with_registry(with_builtins())`; each `init_*` / `transcriber()` delegates; version-cache + transcriber slot unchanged
  - http-server: `HttpServerConfig::backend_context()`, `wire_default_backends_with()`; the `Err→warn+None` downgrade and pgvector coherence guard stay in the wrappers; `mock-llm` stays default-off (production safety); `dev-mock` remapped to `cognee-components/testing`
  - bindings-common: `HandleState::from_settings_with_registry()` — the injection seam the py/ts/c cloud SDKs route through

## Extension example (closed repo)

```rust
let mut reg = ComponentRegistry::with_builtins();
reg.register_vector(Arc::new(QdrantVectorFactory));
reg.register_llm(Arc::new(LiteRtLlmFactory));
// -> ComponentManager::with_registry / HandleState::from_settings_with_registry / wire_default_backends_with
```

## Verification

- `cargo check --all-targets` (workspace) — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- Unit tests: components, lib (`component_manager`, `config`), http-server (`wiring`, `config`) — pass
- capi cargo-check **default + slim** (`--no-default-features --features sqlite,testing`) — clean
- Neon (TS) binding cargo-check — clean

## Review history

This branch went through two adversarial-review passes (design review before implementation, workflow code-review after). The second commit fixes the regressions the code-review surfaced — most notably a `sqlite://<relative>` startup break and a silent empty-DB footgun from a preemptive `File::create` (now delegated to the driver via `?mode=rwc`), plus restoring several diagnostic messages that the two old paths had.

## Follow-up (separate, in `cognee-cloud-rs`, after this lands + rev bump)

Register `QdrantVectorFactory` / `LiteRtLlmFactory` at the closed entry points.

## Open question for reviewers

The standalone http-server now accepts `VECTOR_DB_PROVIDER=brute-force` (ephemeral, data lost on restart) for parity with lib/CLI, where it previously hard-rejected it. Intentional per the design, but flagging in case the server should keep rejecting brute-force specifically.

### DIFF
--- a/crates/components/src/builtins/database.rs
+++ b/crates/components/src/builtins/database.rs
@@ -30,12 +30,18 @@ fn sqlite_fs_path(url: &str) -> Option<String> {
         return None;
     }
     let after = url.trim_start_matches("sqlite:");
+    // Collapse any run of leading slashes on the absolute forms to a single
+    // one, so an extra slash (e.g. `sqlite:////abs`) can't yield a
+    // double-leading-slash path that some platforms read as a UNC/network path.
     let path = if let Some(rest) = after.strip_prefix("//localhost/") {
-        format!("/{rest}")
-    } else if let Some(rest) = after.strip_prefix("///") {
-        format!("/{rest}")
+        format!("/{}", rest.trim_start_matches('/'))
     } else if let Some(rest) = after.strip_prefix("//") {
-        rest.to_string()
+        // `//<rest>`: absolute when `rest` starts with `/` (i.e. `sqlite:///…`),
+        // otherwise a relative path (`sqlite://data/…`).
+        match rest.strip_prefix('/') {
+            Some(abs) => format!("/{}", abs.trim_start_matches('/')),
+            None => rest.to_string(),
+        }
     } else {
         after.to_string()
     };
@@ -152,6 +158,9 @@ mod tests {
             sqlite_fs_path("sqlite://localhost/abs/db"),
             Some("/abs/db".into())
         );
+        // Extra leading slashes collapse to one (no double-leading-slash /
+        // accidental UNC path).
+        assert_eq!(sqlite_fs_path("sqlite:////abs/db"), Some("/abs/db".into()));
         // Query strings are stripped.
         assert_eq!(
             sqlite_fs_path("sqlite://data/db?mode=rwc"),
--- a/crates/components/src/builtins/database.rs
+++ b/crates/components/src/builtins/database.rs
@@ -9,69 +9,66 @@ use cognee_database::{DatabaseConnection, connect, initialize};
 use crate::context::BackendBuildContext;
 use crate::error::ComponentError;
 
+/// Extract the on-disk filesystem path from a SQLite URL, or `None` for
+/// non-file URLs (`postgres://…`, in-memory `sqlite::memory:`).
+///
+/// SQLite has no meaningful remote "host", so any `//<authority>` is treated as
+/// part of the path — matching the HTTP server's historical
+/// `strip_prefix("sqlite://")`, which is why a two-slash *relative* form like
+/// `sqlite://data/cognee.db` must still yield `data/cognee.db` (not be skipped
+/// as a host form). Shapes handled:
+///
+///   `sqlite:./rel/db`            → `./rel/db`        (1-slash relative)
+///   `sqlite:/abs/db`             → `/abs/db`         (1-slash absolute)
+///   `sqlite://rel/db`            → `rel/db`          (2-slash relative)
+///   `sqlite:///abs/db`           → `/abs/db`         (3-slash absolute)
+///   `sqlite://localhost/abs/db`  → `/abs/db`         (host form)
+///
+/// Any `?query` (e.g. `?mode=rwc`) is stripped.
+fn sqlite_fs_path(url: &str) -> Option<String> {
+    if !url.starts_with("sqlite:") || url.contains(":memory:") {
+        return None;
+    }
+    let after = url.trim_start_matches("sqlite:");
+    let path = if let Some(rest) = after.strip_prefix("//localhost/") {
+        format!("/{rest}")
+    } else if let Some(rest) = after.strip_prefix("///") {
+        format!("/{rest}")
+    } else if let Some(rest) = after.strip_prefix("//") {
+        rest.to_string()
+    } else {
+        after.to_string()
+    };
+    let path = path.split('?').next().unwrap_or(&path).to_string();
+    if path.is_empty() { None } else { Some(path) }
+}
+
 /// Connect to and initialize the relational database from the resolved URL.
 ///
-/// For SQLite file-backed databases, the parent directory is created first:
-/// sea-orm's `?mode=rwc` creates the *file* but not missing ancestor
-/// directories, so a settings override that redirects the DB to a new path
-/// (e.g. per-test isolation) would otherwise fail with "unable to open
-/// database file".
-///
-/// URL shapes handled:
-///   `sqlite:./rel/path/db`       (relative, 1-slash)
-///   `sqlite:///abs/path/db`      (absolute, 3-slash)
-///   `sqlite://localhost/abs/db`  (host form)
-/// All others (postgres, in-memory `sqlite::memory:`) are left alone.
+/// For SQLite file-backed databases, the parent directory is created first
+/// (non-fatal): sea-orm's `?mode=rwc` creates the *file* but not missing
+/// ancestor directories, so a settings override that redirects the DB to a new
+/// path (e.g. per-test isolation) would otherwise fail with "unable to open
+/// database file". File creation itself is left to the driver via `?mode=rwc`
+/// (callers that want auto-create put `mode=rwc` in the URL); this function
+/// never creates the DB file, so a mistyped path still fails loudly at connect
+/// instead of silently producing an empty database at the wrong location.
 pub async fn build_database(
     ctx: &BackendBuildContext,
 ) -> Result<Arc<DatabaseConnection>, ComponentError> {
     let url = &ctx.relational_db_url;
 
-    if url.starts_with("sqlite:") && !url.contains(":memory:") {
-        // Strip the sqlite: scheme and any leading host ("//localhost") or
-        // extra slashes to get the raw filesystem path (before '?').
-        let after_scheme = url.trim_start_matches("sqlite:");
-        let path_part = if after_scheme.starts_with("//localhost/") {
-            Some(&after_scheme["//localhost".len()..])
-        } else if after_scheme.starts_with("///") {
-            // sqlite:///abs/path — empty authority, absolute path.
-            Some(&after_scheme[2..])
-        } else if after_scheme.starts_with("//") {
-            // sqlite://somehost/... — genuine host form; leave entirely to the
-            // driver instead of attempting create_dir_all("//somehost").
-            None
-        } else {
-            Some(after_scheme)
-        };
-        // Drop query string (e.g. ?mode=rwc).
-        if let Some(path_part) = path_part {
-            let path_no_query = path_part.split('?').next().unwrap_or(path_part);
-            let db_path = Path::new(path_no_query);
-            if let Some(parent) = db_path.parent()
-                && !parent.as_os_str().is_empty()
-            {
-                // Non-fatal: an unusual-but-driver-valid URL must still reach
-                // sea-orm and surface the driver's own error.
-                if let Err(e) = std::fs::create_dir_all(parent) {
-                    tracing::warn!(
-                        "could not create SQLite parent directory '{}': {e}",
-                        parent.display()
-                    );
-                }
-            }
-            // Create the DB file when missing. `cognee-lib`'s default URL carries
-            // `?mode=rwc` (the driver would create it), but the HTTP server's
-            // `sqlite://{path}` form has no such query, so without this an
-            // absolute path to a not-yet-existing file fails to open. Creating
-            // an empty file is a valid empty SQLite DB and is idempotent for the
-            // `mode=rwc` form. Non-fatal — let the driver surface real errors.
-            if !db_path.as_os_str().is_empty()
-                && !db_path.exists()
-                && let Err(e) = std::fs::File::create(db_path)
-            {
+    if let Some(path) = sqlite_fs_path(url) {
+        let db_path = Path::new(&path);
+        if let Some(parent) = db_path.parent()
+            && !parent.as_os_str().is_empty()
+        {
+            // Non-fatal: an unusual-but-driver-valid URL must still reach
+            // sea-orm and surface the driver's own error.
+            if let Err(e) = std::fs::create_dir_all(parent) {
                 tracing::warn!(
-                    "could not create SQLite database file '{}': {e}",
-                    db_path.display()
+                    "could not create SQLite parent directory '{}': {e}",
+                    parent.display()
                 );
             }
         }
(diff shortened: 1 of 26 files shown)


crates/http-server/src/wiring.rs
```diff
@@ -33,20 +35,38 @@ fn ensure_dir(path: &Path) -> Result<(), ServerError> {
         .map_err(|e| ServerError::Other(anyhow!("create_dir_all({}): {e}", path.display())))
 }
 
+/// Wire the default standalone backends using the OSS built-in registry.
 pub async fn wire_default_backends(
     cfg: &HttpServerConfig,
+) -> Result<ComponentHandles, ServerError> {
+    wire_default_backends_with(cfg, &ComponentRegistry::with_builtins()).await
+}
```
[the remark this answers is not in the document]
GraDKh: This compiles as-is. The `&ComponentRegistry::with_builtins()` temporary is an argument in the function's *tail expression*, so its drop scope is the end of that whole `…await` statement — the borrow stays live for the entire await, not dropped mid-way. Verified: `cargo check --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, and the crate's tests all build/pass on this exact code. Leaving as-is.

crates/components/src/registry.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        reg
+    }
+
+    // ── registration (extension points) ───────────────────────────────────
+
+    /// Register (or override) a vector backend factory under `f.provider()`.
+    pub fn register_vector(&mut self, f: Arc<dyn VectorDbFactory>) {
+        self.vector.insert(f.provider().to_lowercase(), f);
+    }
```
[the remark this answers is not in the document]
GraDKh: Good catch — fixed in 708ea10. brute-force now registers under a single canonical key, and the spelling variants (`brute_force`/`bruteforce`) are canonicalized at lookup time in `build_vector`, so a `register_vector()` override applies consistently regardless of how config spells it.

crates/components/src/context.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Resolved LLM / transcriber inputs.
+    pub llm: LlmInputs,
+}
+
+/// Resolved embedding-engine inputs. Mapped to a `cognee_embedding::EmbeddingConfig`
+/// by [`crate::build_embedding_config`].
+#[derive(Clone)]
+pub struct EmbeddingInputs {
+    /// Lowercase provider string. Unknown values fall back to `onnx`, matching
+    /// the historical `ComponentManager` behavior.
```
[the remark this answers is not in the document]
GraDKh: Fixed in 708ea10 — the doc now states that empty defaults to onnx while a non-empty unrecognized provider is rejected, matching `DefaultEmbeddingFactory`.

crates/components/src/builtins/embedding.rs
```diff
@@ -0,0 +1,104 @@
...
+        _ => EmbeddingProvider::Onnx,
+    };
+
+    let mock_mode = if inputs.mock_deterministic {
+        MockVectorMode::Deterministic
+    } else {
+        MockVectorMode::Zero
+    };
+
+    EmbeddingConfig {
```
GraDKh: **[1 / blocking] `EmbeddingConfig` struct literal breaks under Cargo feature unification.** The `onnx` field's *existence* is gated on `cognee-embedding`'s `onnx` feature (`crates/embedding/src/config.rs:199`), but this literal initializes it only under `cognee-components`'s *own* `onnx` feature. Reproduced: `cargo check -p cognee-components -p cognee-embedding --no-default-features --features cognee-embedding/onnx` fails with `` error[E0063]: missing field `onnx` ``. No in-tree CI combo trips it today (all consumers pair the features), but it violates feature additivity and breaks the documented `default-features = false` usage for any consumer whose graph also enables `cognee-embedding/onnx`.

Fix is one line: use `..EmbeddingConfig::default()` functional-update syntax (a `Default` impl already exists at `config.rs:259`).

crates/components/src/builtins/embedding.rs
```diff
@@ -0,0 +1,104 @@
...
+    async fn build(
+        &self,
+        ctx: &BackendBuildContext,
+    ) -> Result<Arc<dyn EmbeddingEngine>, ComponentError> {
+        // A non-empty but unrecognized provider is a misconfiguration (typo /
+        // unsupported backend): surface it instead of silently falling back to
+        // ONNX. Empty means "use the default" and is allowed. `mock` short-
+        // circuits provider selection, so skip the check when mocking.
+        let provider = ctx.embedding.provider.trim();
+        if !ctx.embedding.mock && !provider.is_empty() && !is_known_embedding_provider(provider) {
```
GraDKh: **[2 / blocking] Unknown embedding providers now hard-error, silently breaking previously-working deployments — and the trait doc claims the opposite.** Old `init_embedding_engine` mapped any unrecognized `EMBEDDING_PROVIDER` to ONNX (`_ => EmbeddingProvider::Onnx`); this factory now returns `Err("unknown embedding provider ...")` and no layer above restores the fallback, so a deployment with e.g. `EMBEDDING_PROVIDER=azure` fails every cognify on upgrade. The hard error looks deliberate (arguably an improvement), but `crates/components/src/traits.rs:77` still documents that the default factory "preserves the historical unknown-provider → onnx fallback" — that doc is now false. At minimum fix the doc; ideally also call out the breaking change in the PR description / changelog.

crates/http-server/src/wiring.rs
```diff
@@ -109,208 +129,98 @@ pub async fn wire_default_backends(
...
-async fn wire_graph_db(cfg: &HttpServerConfig) -> Result<Arc<dyn GraphDBTrait>, ServerError> {
+async fn wire_graph_db(
+    cfg: &HttpServerConfig,
+    registry: &ComponentRegistry,
+    ctx: &cognee_components::BackendBuildContext,
+) -> Result<Arc<dyn GraphDBTrait>, ServerError> {
+    // The standalone server ships only the embedded ladybug graph; guard early
+    // with an actionable message rather than the registry's generic
+    // "unregistered provider" error.
     if !cfg.graph_provider.eq_ignore_ascii_case("ladybug") {
```
GraDKh: **[3] The hardcoded "only 'ladybug' is supported" guard runs before the registry lookup, defeating the extension seam this PR builds.** A closed entry point that registers a custom graph factory and calls `wire_default_backends_with` (the documented extension path, per the PR body's own example) still fails here. It even rejects `"kuzu"`, which the server's own `with_builtins()` registers — so the message is wrong in the same build. Unlike `wire_vector_db`, which correctly delegates to the registry (whose `unsupported_msg` already lists registered providers). Note pggraph parity has two other blockers too (the server Cargo.toml omits `pggraph`, and `backend_context()` hardcodes `graph_postgres_url: None`), but as written this guard makes registered graph factories unreachable.

crates/lib/src/config.rs
```diff
@@ -582,6 +610,218 @@ impl Settings {
...
+
+        let graph_provider = self.graph_database_provider.to_lowercase();
+        let graph_postgres_url = {
+            #[cfg(feature = "pggraph")]
+            {
+                if matches!(graph_provider.as_str(), "postgres" | "postgresql") {
+                    // Log the specific resolution failure (e.g. "Missing required
+                    // Postgres graph credentials") rather than swallowing it — the
+                    // factory only sees `None` and can't restate the cause.
+                    match self.resolved_graph_postgres_url() {
```
GraDKh: **[4] Postgres URL-resolution failures are downgraded to `tracing::warn!` + `None`, hiding the root cause from the error path.** Previously `graph_db()` returned the specific cause (`"Missing required Postgres graph credentials"`); now the factory can only say `"graph_database_provider=postgres requires a resolved Postgres URL"`, and the real cause is invisible to SDK users with no tracing subscriber. Since resolution only runs when the provider *is* postgres, the information loss hits exactly the users who need it. Consider carrying the resolution error string in `BackendBuildContext` so the factory can restate it. (Same pattern for the vector twin a few lines below.)

crates/components/src/registry.rs
```diff
@@ -0,0 +1,377 @@
...
+/// brute-force spelling variants (`brute_force`, `bruteforce`) onto the single
+/// registered key `brute-force`. All other providers pass through lowercased.
+fn canonical_vector_provider(provider: &str) -> String {
+    match provider.to_lowercase().as_str() {
+        "brute-force" | "brute_force" | "bruteforce" => "brute-force".to_string(),
+        other => other.to_string(),
+    }
+}
+
+fn unsupported_msg(field: &str, provider: &str, supported: &[String]) -> String {
```
GraDKh, line 244: **[5] `unsupported_msg` picks its feature hint from the provider string, ignoring which field failed — producing misleading errors.** `vector_db_provider=postgres` yields "rebuild with the `pggraph` crate feature" (a graph hint for a vector error, shown even when pggraph is already enabled). And a no-`ladybug` build (the CI-exercised capi slim build) with the default `graph_database_provider=ladybug` gets no feature hint at all — steering users to closed-adapter registration where origin/main said "requires the `ladybug` crate feature". A table of known-but-disabled provider→feature pairs *per component kind* would fix both.

crates/http-server/src/config.rs
```diff
@@ -559,6 +571,95 @@ impl HttpServerConfig {
...
+            vector_db_url: self.vector_db_url.clone(),
+            vector_postgres_url,
+            embedding_dimensions: self.embedding_dimensions as usize,
+            embedding: cognee_components::EmbeddingInputs {
+                provider: self.embedding_provider.trim().to_ascii_lowercase(),
+                model: self.embedding_model_name.clone(),
+                dimensions: self.embedding_dimensions as usize,
+                endpoint,
+                api_key,
+                batch_size: 36,
```
GraDKh: **[6] `backend_context()` newly hardcodes the BGE-Small paths and embedding defaults that the embedding crate already owns.** `batch_size: 36`, `max_completion_tokens: 8191`, `onnx_max_sequence_length: 512`, `onnx_batch_size: 32`, and the `./target/models/...` asset paths all duplicate `OnnxEmbeddingConfig::bge_small()` / `EmbeddingConfig::default()` (`crates/embedding/src/config.rs`). Pre-PR the http-server sourced these via `..Default::default()` and contained none of these literals — so this reintroduces, in miniature, the per-caller drift the PR exists to eliminate. If the default model path changes upstream, this copy silently points at a nonexistent file. Default `EmbeddingInputs` from the embedding crate's constructors instead.

crates/components/src/builtins/embedding.rs
```diff
@@ -0,0 +1,104 @@
...
+use cognee_embedding::{EmbeddingConfig, EmbeddingEngine, EmbeddingProvider, MockVectorMode};
+
+use crate::context::{BackendBuildContext, EmbeddingInputs};
+use crate::error::ComponentError;
+use crate::traits::EmbeddingFactory;
+
+/// Whether `provider` is a recognized embedding backend id (case-insensitive).
+/// An empty string is *not* recognized here; callers treat empty as "use the
+/// onnx default", while a non-empty unrecognized value is a misconfiguration.
+fn is_known_embedding_provider(provider: &str) -> bool {
```
GraDKh: **[7] The embedding provider-id set is encoded three times in this file.** `is_known_embedding_provider` (here), the `match` in `build_embedding_config` with its silent `_ => Onnx` arm, and the error-message string all list the same six providers. They agree today, but adding a provider means updating all three, or you either reject a valid id or silently build ONNX. One fallible parse into `EmbeddingProvider` (whose serde `snake_case` names already match these strings) replaces all of them — validity = `is_some`, config = `unwrap_or(Onnx)`.

crates/lib/src/component_manager.rs
```diff
@@ -104,590 +86,53 @@ impl ComponentManager {
...
-            "postgres" | "postgresql" => Err(ComponentError::Config(
-                "graph_database_provider=postgres requires the `pggraph` crate feature".into(),
-            )),
-
-            other => Err(ComponentError::Config(format!(
-                "Unsupported graph_database_provider '{other}'. Supported: ladybug, kuzu, postgres.",
-            ))),
-        }
+    async fn init_storage(&self) -> Result<Arc<dyn StorageTrait>, ComponentError> {
+        let ctx = self.build_context();
```
GraDKh: **[8] Every `init_*` rebuilds the full `BackendBuildContext`, so one config version does up to 7 context builds — and the Postgres credential-fallback warning now fires once per component instead of once.** The compute cost is cold-path noise (guarded by the version cache), but the duplicated `tracing::warn!` is a user-visible regression versus main, where it fired only in `init_graph_db`. When the graph provider is postgres with fallback creds, `storage()`, `database()`, `llm()`, … each re-emit the warning. Cache `(version, BackendBuildContext)` alongside the other versioned slots to fix both.

GraDKh (commented):
Automated review (medium effort, 8-angle finders → adversarial verify). The consolidation is well executed — the drift-guard tests and the shared `BackendBuildContext` boundary are genuinely good. 8 findings survived verification; **#1 (feature-unification compile break)** and **#2 (unflagged breaking behavior change)** are the ones I'd want resolved before merge. Details inline.

On the PR's open question about `brute-force` on the http-server: three independent finder angles flagged this as the most surprising behavior change. The server previously hard-failed on `VECTOR_DB_PROVIDER=brute-force` and now boots with a purely in-memory store whose vectors vanish on restart — a silent data-loss mode for a long-running server, unlike the CLI/lib case where ephemeral runs are normal. Suggest refusing it unless an explicit opt-in (e.g. `ALLOW_EPHEMERAL_VECTOR_DB=1`) is set, or at least a loud startup warning.

Two plausible-looking candidates were refuted and are not reported: the `sqlite://host/...` directory-creation change matches what sqlx actually does with such URLs, and the case-sensitive `"mock"` comparison in `build_llm` is unreachable because both lowering paths lowercase the provider.

crates/components/src/builtins/embedding.rs
GraDKh: Fixed in 1b146d1 — default-construct `EmbeddingConfig` then assign fields, so the cfg-gated `onnx` field is filled by `Default` when present regardless of *this* crate's `onnx` feature. No more E0063 under feature unification.

crates/components/src/builtins/embedding.rs
GraDKh: Fixed in 1b146d1 — corrected the `EmbeddingFactory` doc (empty→onnx default, non-empty-unknown→error, an intentional change vs the old silent onnx fallback). Kept the erroring behavior: fail loud on a misconfigured provider rather than embed with the wrong model.

crates/http-server/src/wiring.rs
GraDKh: Fixed in 1b146d1 — `wire_graph_db` delegates to the registry like `wire_vector_db`; the hardcoded ladybug-only guard is gone, so registered graph factories and the built-in `kuzu` alias are reachable.

crates/lib/src/config.rs
GraDKh: Fixed in 1b146d1 — the pg-url fields are now `Option<Result<String,String>>`; on resolution failure the specific cause flows to the factory and is restated in the returned error (visible without a tracing subscriber).

crates/components/src/registry.rs
GraDKh: Fixed in 1b146d1 — `unsupported_msg` keys the feature hint on component kind + provider (no cross-kind hint; the no-ladybug case now hints `ladybug`). Added a test.

crates/http-server/src/config.rs
GraDKh: Fixed in 1b146d1 — `backend_context` sources the embedding scalar + ONNX-asset defaults from `EmbeddingConfig::default()` / `OnnxEmbeddingConfig::default()` instead of magic literals.

crates/components/src/builtins/embedding.rs
GraDKh: Fixed in 1b146d1 — collapsed to one `parse_embedding_provider`; validity (`is_some`) and config mapping (`unwrap_or(Onnx)`) both derive from it. Added a test.

crates/lib/src/component_manager.rs
GraDKh: Fixed in 1b146d1 — `ComponentManager` caches the lowered `BackendBuildContext` per config version, so `backend_context` (env reads + the fallback warning) runs once per version, not once per component.

crates/components/src/builtins/database.rs
```diff
@@ -0,0 +1,199 @@
...
+/// Any `?query` (e.g. `?mode=rwc`) is stripped.
+fn sqlite_fs_path(url: &str) -> Option<String> {
+    if !url.starts_with("sqlite:") || url.contains(":memory:") {
+        return None;
+    }
+    let after = url.trim_start_matches("sqlite:");
+    let path = if let Some(rest) = after.strip_prefix("//localhost/") {
+        format!("/{rest}")
+    } else if let Some(rest) = after.strip_prefix("///") {
+        format!("/{rest}")
```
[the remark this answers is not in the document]
GraDKh: Fixed in the latest commit — `sqlite_fs_path` now collapses extra leading slashes on the absolute forms to a single `/` (added a `sqlite:////abs` test case). Note our own code produces at most 3 slashes (`sqlite://` + an absolute path), which was already handled; this hardens the general case.

crates/http-server/src/wiring.rs
```diff
@@ -415,23 +326,41 @@ mod tests {
...
+            ..Default::default()
+        };
+        let ctx = cfg.backend_context();
+        assert!(
+            ctx.embedding
+                .onnx_model_path
+                .to_string_lossy()
+                .contains("target/models"),
+            "unset ONNX model path must fall back to the ./target/models default"
+        );
```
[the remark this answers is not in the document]
GraDKh: Fixed — the assertion now normalizes `\` to `/` before checking for `target/models`, so it passes on Windows too.

PR #59, opened 2026-07-02
feat(llm): add ThrottleLlm rate-limit instrument for cognify load measurement

RajdeepKushwaha5:
## Summary

Part of #19. Adds `ThrottleLlm`, a deterministic instrument for measuring cognify's LLM-call behaviour under the one regime that matters for this issue: the high-load case where the provider rate limit, not the network, is the bottleneck.

The discussion on #19 kept returning to a single question: how do you measure the win in the real, rate-limited regime, given that `ReplayLlm` answers instantly (so it hides latency and never hits a rate limit) and a low-sample live run is flaky and never saturates the limit? This is the tool that answers it, offline and at zero token cost.

## What it does

`ThrottleLlm` wraps any `Llm` (in practice `ReplayLlm`) and adds the two properties the instant mock lacks:

- Request counting per outcome (total, allowed, rate_limited). Request count is the deterministic measure of RPM pressure, and it is the headline metric a caching or batching change moves, where wall-clock cannot. The same corpus always issues the same number of requests.
- A rolling 60s requests-per-minute budget. Calls over the budget return the same `LlmError::RateLimitExceeded` a real HTTP 429 maps to, so the retry and backoff paths are exercised deterministically.
- An optional per-call latency, so the overlap a pipelining change buys becomes visible against the instant mock.

With `rpm_limit = None` it is a pure request counter (the "before" measurement); with a limit set it reproduces the rate-limited regime. The simulated 429 is a model, not a claim to match a specific provider; the deterministic request count is the metric meant to carry a headline, and the 429 path is there to exercise retry and backoff.

## Why this shape

- No trait-signature change. It lives behind the existing `mock` feature and delegates every `Llm` method, so it composes with `ReplayLlm` and `RecordingLlm` without touching the core trait. Surfacing byte-exact token counts (for TPM) would need a return-type change on the structured-output path, which I propose as a separate follow-up rather than folding it in here. Shared, global accounting. Hold it behind an `Arc` and share it across the concurrent extraction and summarization tasks, so the counts and the budget are global, exactly as a single provider account would see them. The lock is never held across an await point.

## Tests

Five unit tests (`cargo test -p cognee-llm --features mock throttle`), all deterministic and offline:

- counts every call when unlimited,
- counts both the generate and structured-output paths,
- enforces the RPM budget within a window,
- budget refills as the window slides (using paused tokio time),
- injects the configured latency.

Verified locally: fmt, `clippy --all-targets -D warnings` (with the mock feature), and the five tests.

## What follows

The immediate next step is a cognify bench that drives a representative run through `ThrottleLlm` and prints the before/after request-count table, plus the token surfacing for TPM. This PR is the instrument those build on, kept separate so the tool is reviewable on its own.

## Scope

Part of #19. Does not close it.

### DIFF
--- a/crates/llm/Cargo.toml
+++ b/crates/llm/Cargo.toml
@@ -53,4 +53,4 @@ cognee-test-utils = { path = "../test-utils" }
 dotenv.workspace = true
 httpmock = "0.8"
 tempfile = { workspace = true }
-tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "test-util"] }
--- a/crates/llm/src/mock/mod.rs
+++ b/crates/llm/src/mock/mod.rs
@@ -2,6 +2,8 @@
 mod cassette;
 mod recording;
 mod replay;
+mod throttle;
 pub use cassette::{CassetteEntry, CassetteMethod, LlmCassette, input_hash, vision_hash};
 pub use recording::RecordingLlm;
 pub use replay::{MissPolicy, ReplayLlm};
+pub use throttle::{ThrottleConfig, ThrottleLlm, ThrottleMetrics};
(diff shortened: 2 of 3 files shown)


GraDKh:
@RajdeepKushwaha5 Thanks for efforts, this one looks good

Issue #49, opened 2026-07-01
refactor(cognify): unify the bounded-fan-out idiom for per-chunk LLM calls


RajdeepKushwaha5:
## Summary

The cognify pipeline runs "N per-chunk LLM calls, capped at `max_parallel_extractions`" in **two different idioms**. Raised during review of #43 (part of #19): converging on one keeps the concurrency semantics in a single place so anyone changing them does not have to reason about both.

## Current state (verified)

Four fan-out sites, same requirement, two patterns:

**`Arc<Semaphore>` + `tokio::spawn` + `join_all`, batched:**
- Graph extraction — `crates/cognify/src/tasks.rs:389` (semaphore) / `:403` (spawn) / `:413` (join_all)
- Second extraction path — `tasks.rs:771` / `:806` / `:816`
- Temporal event extraction — `tasks.rs:1193` / `:1206` / `:1216`

**`futures::stream::iter(...).buffer_unordered(max_parallel)`:**
- Summarization — `crates/cognify/src/summarization/extractor.rs` (`summarize_chunks`), added in #43

All four cap in-flight requests at `config.max_parallel_extractions`.

## Why it matters

Two competing bounded-fan-out patterns in the same area add maintenance friction: changing concurrency semantics (the cap, ordering, error handling, cancellation) means editing and reasoning about both. It is a consistency/maintainability issue, not a correctness bug - both idioms currently work.

Relevant behavioural differences to preserve in whichever direction we pick:
- **Error handling:** `buffer_unordered` + `try_collect` drops the remaining futures on the first error; the detached `tokio::spawn` version keeps firing
  already-needless requests. (`buffer_unordered` is the better default here.)
- **Parallelism vs concurrency:** `tokio::spawn` distributes work across runtime worker threads; `buffer_unordered` drives all futures on the current task. For await-bound network I/O (LLM calls) concurrency is what matters, so the spawn's cross-thread parallelism buys little.
- **`'static` bound:** `tokio::spawn` requires `'static` futures (hence the clones); `buffer_unordered` does not.
- **Ordering:** callers that need input order must sort by index (summarization already does).

## Proposed approaches

1. **Shared helper (preferred).** A small `bounded_fan_out(items, max_parallel, f)` utility (returning results in input order, propagating the first error) that all four sites call. One place to define the concurrency semantics.
[...]

PR #63, opened 2026-07-03
feat(cognify): batch multiple chunks per extraction request (#19)

### LINKED ISSUE #19: Perf: optimize LLM API call usage in cognify (concurrency, batch pipelining, prompt caching)

RajdeepKushwaha5:
## Summary

Part of #19. Adds `FactExtractor::extract_facts_grouped`, which sends `group_size` chunks per graph-extraction call instead of one, cutting the number of extraction requests from N to `ceil(N / group_size)`. This is the multi-chunk-batching lever from the issue, and it targets the metric a rate limit actually cares about: request count (RPM pressure).

## How it works

Each batched call asks the model for one knowledge graph per chunk (a `BatchedGraphs { graphs: Vec<KnowledgeGraph> }` response), and the graphs are mapped back to their chunks by input order. This mirrors the pattern the temporal entity enricher already uses.

Failure isolation: if a group's batched response fails to parse or does not return exactly one graph per chunk, that group alone falls back to per-chunk extraction, so one bad group can never drop the others. Output order always matches input order, so callers that pair graphs with chunk ids by position stay correct. `group_size <= 1` is exactly per-chunk extraction.

## The number

Measured with the `ThrottleLlm` instrument from #59, counting requests straight from the extraction path:

```
graph extraction, 12-chunk document, group_size=4:
scenario                   requests
per-chunk (pre-#19)              12
batched                           3
```

Requests drop from N to `ceil(N / group_size)`. That is deterministic and offline, no API key and no flakiness.

## Honest scope: opt-in, not defaulted

Batching trades requests for a larger per-call prompt and weaker per-chunk isolation, and whether it preserves extraction quality (does the model find the same entities and edges when K chunks share one call) is model-dependent. The issue frames this as a data-driven decision, so this PR lands the mechanism and the request-count proof, and keeps batching opt-in. The semantic-parity calibration on a real corpus (node and edge parity vs `group_size`) is the follow-up before wiring batching into the pipeline by default. I did not want to default it on without that data.

## Tests

Three deterministic, offline tests (`cargo test -p cognee-cognify --lib grouped_extraction`):
- request count drops from N to `ceil(N / group_size)`, counted via `ThrottleLlm`,
- one graph per chunk in input order, with the same empty-name to id post-processing as the per-chunk path,
- a bad batch (wrong graph count) falls back to per-chunk, so every chunk still produces a graph.

Verified: fmt, `clippy --all-targets -D warnings`, and the three tests.

## Scope

Part of #19. Stacked on #59 (`ThrottleLlm`). Does not close #19.

## Screenshot 
Screenshot 2026-07-03 110104

### DIFF
--- a/crates/cognify/src/fact_extraction/extractor.rs
+++ b/crates/cognify/src/fact_extraction/extractor.rs
@@ -50,6 +50,21 @@ fn normalize_node_names(graph: &mut KnowledgeGraph) {
     }
 }
 
+/// The `GenerationOptions` shared by the per-chunk and batched extraction calls.
+///
+/// `max_tokens: None` is a documented invariant (Python parity): the Python
+/// `acreate_structured_output` passes no cap, so the response uses the model's
+/// full default output budget. A small cap truncates a dense graph mid-JSON and
+/// aborts cognify with a deserialization error. Kept in one place so the two
+/// call sites cannot drift.
+fn extraction_options() -> GenerationOptions {
+    GenerationOptions {
+        temperature: Some(0.1),
+        max_tokens: None,
+        ..Default::default()
+    }
+}
+
 /// Why a batched group call could not be used as-is, so the caller can react to
 /// each cause differently.
 enum GroupError {
@@ -132,21 +147,7 @@ impl FactExtractor {
 
         let result: M = self
             .llm
-            .create_structured_output(
-                text,
-                system_prompt,
-                // Python parity: `acreate_structured_output` passes NO
-                // max_tokens/max_completion_tokens to the extraction call, so
-                // the response uses the model's full default output budget. A
-                // small cap here truncates large graphs mid-JSON on dense
-                // chunks, aborting cognify with a deserialization error. Leave
-                // max_tokens as None to match Python (no artificial cap).
-                Some(GenerationOptions {
-                    temperature: Some(0.1),
-                    max_tokens: None,
-                    ..Default::default()
-                }),
-            )
+            .create_structured_output(text, system_prompt, Some(extraction_options()))
             .await
             .map_err(|e| CognifyError::LlmError(e.to_string()))?;
 
@@ -261,7 +262,11 @@ impl FactExtractor {
     /// per-call prompt. Whether it preserves extraction quality depends on the
     /// model and is a data-driven decision (see the parity harness in the tests);
     /// keep it opt-in until a calibration run on your corpus supports the chosen
-    /// `group_size`.
+    /// `group_size`. The count win is also not unconditional: a mis-counted group
+    /// falls back to per-chunk, so a model that persistently mis-counts issues
+    /// `N + ceil(N / group_size)` requests, worse than plain per-chunk `N`. The
+    /// calibration run should therefore watch the mismatch rate, not just
+    /// node/edge parity.
     pub async fn extract_facts_grouped(
         &self,
         texts: Vec<String>,
@@ -272,7 +277,11 @@ impl FactExtractor {
         let mut out: Vec<KnowledgeGraph> = Vec::with_capacity(texts.len());
 
         for group in texts.chunks(group_size) {
-            if group_size == 1 {
+            // A single-chunk group (group_size == 1, or a trailing remainder like
+            // n=7,k=3 -> [3,3,1]) takes the per-chunk path: batching one chunk
+            // saves no requests and would send it as a `[{index,text}]` array with
+            // the batch instructions, a different prompt than plain `extract_facts`.
+            if group.len() == 1 {
                 out.push(
                     self.extract_facts(&group[0], custom_prompt.as_deref())
                         .await?,
@@ -338,13 +347,9 @@ impl FactExtractor {
             .create_structured_output(
                 &user_prompt,
                 &system_prompt,
-                // Same as per-chunk extraction: no output cap, so a dense batch
-                // is not truncated mid-JSON (Python parity).
-                Some(GenerationOptions {
-                    temperature: Some(0.1),
-                    max_tokens: None,
-                    ..Default::default()
-                }),
+                // Same options as the per-chunk path (no output cap); see
+                // `extraction_options`.
+                Some(extraction_options()),
             )
             .await
             .map_err(|e| {
@@ -361,6 +366,11 @@ impl FactExtractor {
         for graph in &mut graphs {
             normalize_node_names(graph);
         }
+        debug!(
+            "Batched extraction produced {} graph(s) for {} chunk(s)",
+            graphs.len(),
+            group.len()
+        );
         Ok(graphs)
     }
 
--- a/crates/cognify/src/fact_extraction/extractor.rs
+++ b/crates/cognify/src/fact_extraction/extractor.rs
@@ -5,7 +5,7 @@
 
 use std::sync::Arc;
 
-use cognee_llm::{GenerationOptions, Llm, LlmExt};
+use cognee_llm::{GenerationOptions, Llm, LlmError, LlmExt};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use tracing::{debug, warn};
@@ -37,6 +37,29 @@ struct BatchedGraphs {
     /// One graph per input chunk, in input order.
     #[serde(default)]
     graphs: Vec<KnowledgeGraph>,
+}
+
+/// Ensure every node has a non-empty name (Python `Node.__init__` compat: name
+/// defaults to id when empty). Shared by the per-chunk and batched paths so the
+/// rule can only ever live in one place.
+fn normalize_node_names(graph: &mut KnowledgeGraph) {
+    for node in &mut graph.nodes {
+        if node.name.is_empty() {
+            node.name = node.id.clone();
+        }
+    }
+}
+
+/// Why a batched group call could not be used as-is, so the caller can react to
+/// each cause differently.
+enum GroupError {
+    /// The provider rate-limited the batched call. Falling back to per-chunk
+    /// would issue `group_size` more requests and amplify the 429, so the caller
+    /// propagates this instead of fanning out.
+    RateLimited(CognifyError),
+    /// A parse / serialization / other failure. Safe to fall back to per-chunk
+    /// extraction, since that is a different (smaller) request shape.
+    Other(CognifyError),
 }
 
 /// Fact extractor for knowledge graph generation.
@@ -161,12 +184,7 @@ impl FactExtractor {
         );
 
         // Post-processing: ensure every node has a non-empty name (Python compat).
-        // In Python's Node.__init__, name defaults to id when empty.
-        for node in &mut graph.nodes {
-            if node.name.is_empty() {
-                node.name = node.id.clone();
-            }
-        }
+        normalize_node_names(&mut graph);
 
         Ok(graph)
     }
@@ -231,7 +249,13 @@ impl FactExtractor {
     /// order always matches input order, so callers that pair graphs with chunk
     /// ids by position stay correct.
     ///
-    /// `group_size <= 1` is exactly per-chunk extraction (one request per chunk).
+    /// `group_size <= 1` reduces to per-chunk extraction (one request per chunk).
+    /// That equivalence is on request count only: groups here are dispatched
+    /// serially, whereas [`extract_facts_batch`](Self::extract_facts_batch) runs
+    /// chunks concurrently, so the two are not latency-equivalent. Callers that
+    /// wire this into a concurrent stage should group chunks and push each group
+    /// through their existing parallelism bound rather than hand a whole document
+    /// to this serial loop.
     ///
     /// Note: this trades requests for weaker per-chunk isolation and a larger
     /// per-call prompt. Whether it preserves extraction quality depends on the
@@ -256,20 +280,32 @@ impl FactExtractor {
                 continue;
             }
 
-            match self.extract_group(group, custom_prompt.as_deref()).await {
-                Ok(graphs) if graphs.len() == group.len() => out.extend(graphs),
-                other => {
-                    if let Err(e) = other {
-                        warn!("Batched extraction failed ({e}); falling back to per-chunk.");
-                    } else {
-                        warn!(
-                            "Batched extraction returned the wrong graph count; \
-                             falling back to per-chunk for this group."
-                        );
-                    }
-                    for text in group {
-                        out.push(self.extract_facts(text, custom_prompt.as_deref()).await?);
-                    }
+            // A batched group either succeeds with the right count, or we decide
+            // whether to fall back to per-chunk. We never fall back on a provider
+            // rate limit: that would issue `group_size` more requests and amplify
+            // the 429 the batched call already hit, so it propagates instead.
+            let fall_back = match self.extract_group(group, custom_prompt.as_deref()).await {
+                Ok(graphs) if graphs.len() == group.len() => {
+                    out.extend(graphs);
+                    false
+                }
+                Ok(_) => {
+                    warn!(
+                        "Batched extraction returned the wrong graph count; \
+                         falling back to per-chunk for this group."
+                    );
+                    true
+                }
+                Err(GroupError::RateLimited(e)) => return Err(e),
+                Err(GroupError::Other(e)) => {
+                    warn!("Batched extraction failed ({e}); falling back to per-chunk.");
+                    true
+                }
+            };
+
+            if fall_back {
+                for text in group {
+                    out.push(self.extract_facts(text, custom_prompt.as_deref()).await?);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


RajdeepKushwaha5:
Hi Sir @GraDKh 
I'm framing this against the rest of the #19 work, since the three pieces are meant to fit together:

- **#43**: bound summarization concurrency + retry jitter. Metric: peak in-flight, 60 to 20.
- **#59**: `ThrottleLlm`, the deterministic, offline rate-limit instrument.
- **this PR**: the one lever that reduces request count directly, batching `group_size` chunks per extraction call. On a 12-chunk document at `group_size=4`, requests go from 12 to 3, counted straight from the extraction path by #59's instrument.

Two things I would value your steer on:

1. **Default vs opt-in.** The request-count drop is proven offline and deterministic. What is not proven offline is semantic parity: whether the model finds the same entities and edges as `group_size` grows. That is model-dependent and needs a calibration run on a real corpus, so I kept batching opt-in rather than guess. Happy to run that node/edge parity vs `group_size` calibration and post the table if you want it before we consider defaulting it on.

2. **Next lever.** The remaining ones from the issue are prompt caching (fewer input tokens, TPM pressure) and unifying the bounded-fan-out idiom (#49). Caching needs a small return-type change on the structured-output path to surface token counts for the before/after, which I would rather raise with you before touching the trait. Which would you like me to prioritize?

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +218,112 @@ impl FactExtractor {
...
+    /// ids by position stay correct.
+    ///
+    /// `group_size <= 1` is exactly per-chunk extraction (one request per chunk).
+    ///
+    /// Note: this trades requests for weaker per-chunk isolation and a larger
+    /// per-call prompt. Whether it preserves extraction quality depends on the
+    /// model and is a data-driven decision (see the parity harness in the tests);
+    /// keep it opt-in until a calibration run on your corpus supports the chosen
+    /// `group_size`.
+    pub async fn extract_facts_grouped(
```
GraDKh, line 270: **Non-blocking** — safe to merge as-is, but worth capturing before this gets wired into the pipeline.

Groups here are dispatched **sequentially**: the `for group in texts.chunks(group_size)` loop awaits each group's LLM call before starting the next (and the per-chunk fallback inside a failed group is serial too). That's fine while this method has no production callers, but two notes:

1. **Docstring wording.** "`group_size <= 1` is exactly per-chunk extraction" is only true on *request count* — execution is serial, whereas `extract_facts_batch` runs chunks concurrently. Worth qualifying so nobody swaps `extract_facts_batch` → `extract_facts_grouped(.., 1)` expecting equivalent latency.

2. **Wiring follow-up (the important one).** If this is later wired into `tasks.rs` by calling it with the full chunk list, the serial loop makes the extraction stage roughly an order of magnitude slower than today's path (e.g. 100 chunks @ `group_size=4` → 25 strictly-serial calls, each larger/slower than a single-chunk call). Today `tasks.rs` already bounds concurrency with `Semaphore(max_parallel_extractions)`. The wiring change should group chunks *there* and push each group through that existing semaphore, rather than delegating a whole document to this method's internal serial loop. Serial dispatch buys no extra RPM protection — grouping already cut the request count, and RPM is a rolling count, not a burst-shape limit.

3. **Optional hardening.** The fallback amplifies load under real rate limiting: a `RateLimitExceeded` from `extract_group` drops into the per-chunk fallback and issues `group_size` *more* requests (and any 429 there aborts the run via `?`). Consider propagating/backing off on `RateLimitExceeded` specifically instead of falling back.

(For reference: Python cognee has no multi-chunk batching — it's a bare per-chunk `asyncio.gather` — so there's no upstream parity precedent to preserve here.)

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +218,112 @@ impl FactExtractor {
...
+                }),
+            )
+            .await
+            .map_err(|e| CognifyError::LlmError(e.to_string()))?;
+
+        let mut graphs = batched.graphs;
+        // Same post-processing as extract_facts: empty name defaults to id.
+        for graph in &mut graphs {
+            for node in &mut graph.nodes {
+                if node.name.is_empty() {
```
GraDKh: **Optional / nit.** This empty-name→id normalization is now duplicated verbatim with the copy in `extract_facts` (the Python `Node.__init__` name-defaults-to-id parity rule). If that rule ever changes, both sites have to move together and it's easy to miss the batched one.

A tiny shared helper called from both paths would keep them in lockstep:

```rust
fn normalize_node_names(graph: &mut KnowledgeGraph) {
    for node in &mut graph.nodes {
        if node.name.is_empty() {
            node.name = node.id.clone();
        }
    }
}
```

Then `extract_facts` calls it on its single graph and `extract_group` maps it over `graphs`. Non-blocking.

RajdeepKushwaha5:
Hi Sir @GraDKh, a quick status and a question on where to take #19 next.

**Landed so far**

- #43 (merged): bounded summarization concurrency + retry jitter.
- #59 (merged): the `ThrottleLlm` load instrument, so request count under a deterministic rate-limited regime is now measurable offline.
- #63 (in review, review addressed): multi-chunk-per-request batching for graph extraction, opt-in, with the request-count proof from #59's instrument. I also folded in your last review notes (rate-limit propagation instead of per-chunk fan-out, a shared node-name normalizer, and the docstring latency caveat).

**What is still open from the issue**

1. Batch-pipelining for graph/fact extraction. This is the half of opportunity 1 that is not done yet: `tasks.rs` still drains each batch fully before starting the next (`for batch in chunks(data_per_batch) { join_all(...) }`), so a few slow chunks stall the whole batch. Summarization already moved to a single bounded `buffer_unordered` in #43; extraction has not.
2. Prompt caching for the static prompt prefix (Anthropic `cache_control`, OpenAI automatic prefix caching).
3. Token surfacing for TPM before/after, which caching needs to be quantified. This is the small return-type change on the structured-output path I flagged earlier.
4. Unifying the bounded-fan-out idiom (#49).
5. Semantic-parity calibration for #63 (node/edge parity vs `group_size` on a real corpus) before batching is defaulted on and wired into `tasks.rs`.

**What I propose to do next: finish batch-pipelining on the extraction path**

Mirror the #43 summarization change on graph/fact extraction: replace the serial `for batch { join_all }` with a single `futures::stream::iter(all_chunks).buffer_unordered(max_parallel_extractions)`, keeping the same global concurrency cap but removing the inter-batch barrier. I would measure it with #59's instrument (request count is unchanged, and with the per-call latency set, the peak-in-flight and wall-clock overlap become visible), so the change ships with before/after numbers rather than an assertion. It needs no trait change and no new instrument, and it completes a lever that is already half shipped.

**Where I need your steer**

- Prompt caching and the token-surfacing return-type change both touch the shared structured-output path, so I do not want to move on them before your call. If you want the token surfacing first (so caching has a metric), I will write up the exact `Llm` trait / return-type change as a small proposal for your review before touching it.
- Priority between prompt caching and unifying the fan-out idiom (#49). My instinct is extraction pipelining first (unblocked and measurable), then #49 once both stages share one idiom worth unifying, then caching once the token metric is in, but I will follow whatever order you prefer.

For #63 specifically: I can run the node/edge parity vs `group_size` calibration on a representative corpus and post the table, so the "default on vs opt-in" decision is data-driven rather than a guess. 


Let me know if you want those numbers and I will run it.

GraDKh (commented):
This looks generally solid and the prior review notes are well addressed — just a few small nits inline that are worth tidying up before this gets wired into the pipeline. Thanks!

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +236,134 @@ impl FactExtractor {
...
+                Err(GroupError::RateLimited(e)) => return Err(e),
+                Err(GroupError::Other(e)) => {
+                    warn!("Batched extraction failed ({e}); falling back to per-chunk.");
+                    true
+                }
+            };
+
+            if fall_back {
+                for text in group {
+                    out.push(self.extract_facts(text, custom_prompt.as_deref()).await?);
```
GraDKh, line 317: **Nit (low).** The fallback is per-*group*, not per-chunk-within-the-batch: a count-mismatched group costs `1 batched attempt + group.len()` requests, so a corpus/model that reliably mis-counts makes total requests `N + ceil(N/group_size)` — i.e. *worse* than plain per-chunk `N`. The overhead is bounded (≤50% at k=2, ≤25% at k=4) and it's opt-in, so this isn't blocking, but the docstring currently frames only quality as the risk. Worth a one-line note that persistent mis-counts can invert the request-count win, so the calibration run watches mismatch rate, not just node/edge parity.

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +236,134 @@ impl FactExtractor {
...
+        &self,
+        texts: Vec<String>,
+        custom_prompt: Option<String>,
+        group_size: usize,
+    ) -> Result<Vec<KnowledgeGraph>, CognifyError> {
+        let group_size = group_size.max(1);
+        let mut out: Vec<KnowledgeGraph> = Vec::with_capacity(texts.len());
+
+        for group in texts.chunks(group_size) {
+            if group_size == 1 {
```
GraDKh: **Nit (low).** The fast path is `group_size == 1` only. A *trailing* group of size 1 (e.g. n=7, k=3 → `[3,3,1]`) still goes through `extract_group`, so that lone chunk is sent as a `[{index,text}]` JSON array with `BATCH_GRAPH_INSTRUCTIONS` appended — a different prompt than the same chunk gets via per-chunk `extract_facts` (raw text). Same request count, but batch-mode framing for a degenerate single-chunk batch could shift output. Cheap fix: also take the per-chunk path when `group.len() == 1`.

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +236,134 @@ impl FactExtractor {
...
+                if matches!(e, LlmError::RateLimitExceeded(_)) {
+                    GroupError::RateLimited(ce)
+                } else {
+                    GroupError::Other(ce)
+                }
+            })?;
+
+        let mut graphs = batched.graphs;
+        // Same post-processing as extract_facts (shared helper keeps them in lockstep).
+        for graph in &mut graphs {
```
GraDKh, line 366: **Nit (trivial).** `extract_facts` logs `"Extracted graph with N nodes and M edges"`, but the batched path logs nothing on success (only the two fallback `warn!`s). Consider a per-group `debug!` here for observability parity when debugging a batched run.

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +236,134 @@ impl FactExtractor {
...
+
+        let batched: BatchedGraphs = self
+            .llm
+            .create_structured_output(
+                &user_prompt,
+                &system_prompt,
+                // Same as per-chunk extraction: no output cap, so a dense batch
+                // is not truncated mid-JSON (Python parity).
+                Some(GenerationOptions {
+                    temperature: Some(0.1),
```
GraDKh: **Nit (trivial).** This `GenerationOptions { temperature: Some(0.1), max_tokens: None, .. }` block is byte-identical to the one in `extract`, and the `max_tokens: None` rationale is a documented invariant. Since this PR already extracted `normalize_node_names` into a shared helper to keep the two paths in lockstep, a tiny `fn extraction_options() -> GenerationOptions` would apply the same treatment here and prevent drift.

crates/cognify/src/fact_extraction/extractor.rs
```diff
@@ -197,6 +236,134 @@ impl FactExtractor {
...
+    ) -> Result<Vec<KnowledgeGraph>, GroupError> {
+        let base = custom_prompt.unwrap_or(DEFAULT_GRAPH_PROMPT);
+        let system_prompt = format!("{base}{BATCH_GRAPH_INSTRUCTIONS}");
+
+        let input: Vec<serde_json::Value> = group
+            .iter()
+            .enumerate()
+            .map(|(index, text)| serde_json::json!({ "index": index, "text": text }))
+            .collect();
+        let user_prompt = serde_json::to_string(&input)
```
GraDKh, line 342: **Nit (trivial, no change needed).** `input` is a `Vec<serde_json::Value>` built from `{index: usize, text: String}`, so `to_string` can't realistically fail — this `map_err` is dead in practice. Propagating rather than `expect`ing is consistent with the repo's no-unwrap convention, so fine to leave; flagging only for completeness.

RajdeepKushwaha5:
Thanks. Addressed the four actionable nits in the new commit:

Single-chunk groups: now guarded on `group.len() == 1`, so a trailing remainder group (n=7,k=3 → [3,3,1]) takes the raw-text per-chunk path instead of batch-mode framing for one chunk.
Shared `extraction_options():` the identical `GenerationOptions` block in extract and `extract_group` is now one helper, same as `normalize_node_names`, so the `max_tokens: None` invariant can't drift.
Docstring: noted that a persistently mis-counting model issues `N + ceil(N/group_size)` requests (worse than plain per-chunk `N`), so the calibration watches mismatch rate, not just node/edge parity.
Observability: `extract_group` now `debug!`-logs graph/chunk counts on success, matching `extract_facts`.
Left the dead `map_err` as-is per your note. The wiring follow-up (grouping in `tasks.rs` through the existing semaphore) stays out of this PR.

GraDKh:
Thanks so much for this, @RajdeepKushwaha5 — the shared-helper refactor, the honest scoping around keeping batching opt-in, and the deterministic request-count proof make this a pleasure to review and merge!

Issue #57, opened 2026-07-02
bug(graph): UUID5 slug derived from name only — same-name entities of different types silently merge and corrupt the graph


SaviPandey:
## Summary

Two entities with different types but identical names are silently merged into a single graph node, causing data loss and incorrect 
graph traversal results. No error is raised.

## Root Cause

Node identity is computed from `name` alone — entity type is never included in the UUID seed.

In `crates/cognify/src/graph_integration/deduplication.rs`:

```rust
// Ontology individual path — type never included
let normalized = name.to_lowercase().replace(' ', "_");
let slug = Uuid::new_v5(&Uuid::NAMESPACE_OID, normalized.as_bytes());

// Event attribute extraction path — same issue
let slug = Uuid::new_v5(
    &Uuid::NAMESPACE_OID,
    format!("entity:{}", name).as_bytes(),
);
```

A `Person` named "Alice" and an `Organization` named "Alice" produce identical slugs. On upsert, one silently overwrites the other.

## Proof (Source Code)

The deduplication logic in `deduplicate_nodes_and_edges()` uses the UUID slug as the deduplication key. Since both entities share the same 
slug, only one survives:

```rust
// deduplication keyed on entity.base.id (the UUID5 slug)
// Person("Alice") and Organization("Alice") → same id → one is dropped
unique_nodes.entry(node.entity.base.id).or_insert(node);
```

This is mathematically deterministic — same name always produces the same UUID regardless of type.

## Steps to Reproduce

```rust
cognee.remember("Alice is a software engineer at Acme Corp").await?;
cognee.remember("Alice Corp is a technology company").await?;
// Both "Alice" entities share UUID: 969df200-a070-5bfa-bb0c-4fa2c640e90c
// One silently overwrites the other in the graph
```

## Impact

Any knowledge graph containing homonyms (same name, different type) silently loses one entity. Graph traversal and recall return wrong 
results with no error or warning. Real-world knowledge graphs commonly have this pattern — people and companies sharing names, cities and 
people, etc.

## Proposed Fix

Include entity type in the UUID5 seed at all 4 slug computation sites in `crates/cognify/src/graph_integration/deduplication.rs`.

`node_type` needs to be threaded through to each slug computation site as it is not currently available at all call sites — requires updating the function signatures:

```rust
// Ontology path
let slug = Uuid::new_v5(
    &Uuid::NAMESPACE_OID,
    format!("{}:{}", node_type, normalized).as_bytes(),
);

[...]

PR #65, opened 2026-07-05
Profile the offline pipeline and add a large-document benchmark

### LINKED ISSUE #18: Profile & optimize pure-CPU pipeline performance using the offline mock-replay harness (large document)
## Motivation

The Rust SDK already outperforms the Python `cognee` on CPU, but the pipeline has **never been deliberately profiled or optimized** — the current lead is incidental. We have a clean way to measure *pure CPU* cost: the record/replay mocks let us run the full `add → cognify → search` pipeline with **zero external API calls**, so a profile reflects our own code, not network/LLM latency.

This issue tracks: build a large-document profiling scenario on top of the existing mock harness, run a sampling profiler, identify the hot paths, and propose concrete optimizations.

## What already exists (reuse, don't rebuild)

The offline, zero-API harness is already in place:

- **Replay LLM** — `RecordingLlm` / `ReplayLlm` / content-addressed cassette in [`crates/llm/src/mock/`](https://github.com/topoteretes/cognee-rs/tree/main/crates/llm/src/mock) (`cassette.rs`, `recording.rs`, `replay.rs`). Lookups are SHA256 of `(messages [+ schema])`. Enabled via `MOCK_LLM` + `MOCK_LLM_CASSETTE` (replay) and `COGNEE_RECORD_LLM` (record); wired in [`crates/lib/src/component_manager.rs`](https://github.com/topoteretes/cognee-rs/blob/main/crates/lib/src/component_manager.rs) `init_llm_adapter`. Gated by the `mock-llm` feature.
- **Deterministic mock embeddings** — `MockEmbeddingEngine` (SHA256-derived vectors, content-stable) in [`crates/embedding/src/mock.rs`](https://github.com/topoteretes/cognee-rs/blob/main/crates/embedding/src/mock.rs), selected via `MOCK_EMBEDDING=deterministic`. (No *recording* mode — deterministic vectors are enough to keep embeddings free and offline.)
- **`bench` CLI subcommand** ([`crates/cli/src/commands/bench.rs`](https://github.com/topoteretes/cognee-rs/blob/main/crates/cli/src/commands/bench.rs), `--features bench`) with coarse per-phase timing: prune / db-setup / add / cognify / search.
- **Offline driver + fixtures** — [`scripts/perf/run_mock_bench.sh`](https://github.com/topoteretes/cognee-rs/blob/main/scripts/perf/run_mock_bench.sh), [`scripts/perf/README.md`](https://github.com/topoteretes/cognee-rs/blob/main/scripts/perf/README.md), and committed `fixtures/{memories.json,cassette.json}` (replay with no API key).
- **Criterion bench** [`crates/bench/benches/batch_add_cognify.rs`](https://github.com/topoteretes/cognee-rs/blob/main/crates/bench/benches/batch_add_cognify.rs) — but it drives a **live HTTP server + real API**, so it is not a pure-CPU measurement.

## Gaps this issue closes

1. **Corpus is tiny.** `fixtures/memories.json` is 50 short items (~21 KB). Too small to surface CPU bottlenecks or amortize per-item overhead. We need a **large document / book**.
2. **No profiler integration.** There is zero `pprof` / `flamegraph` wiring anywhere in the repo (confirmed by grep). Phase timings are coarse; we cannot see *which functions* dominate.
3. **No per-stage CPU breakdown** below the 5 bench phases.

### 1. Build a large-document scenario
- Pick a large public-domain book (e.g. a Project Gutenberg text, a few hundred KB–few MB of plain text) as the input corpus.
- Record the cassette **once** with real credentials (the only step needing an API key):
  ```sh
  COGNEE_RECORD_LLM=.../book_cassette.json MOCK_EMBEDDING=deterministic \
    cargo run --release -p cognee-cli --features bench -- bench \
      --memories .../book.json --llm-model gpt-4o-mini --output /tmp/record.json
  ```
- Commit the cassette so the large-doc profile is reproducible in CI with no key (same pattern as the existing 50-memory fixture).
- Run several representative searches across query types — at minimum a no-LLM one (`Chunks`/`Summaries`) and an LLM one (`GraphCompletion`) — so retrieval is profiled too, not just cognify.

### 2. Replay + profile
- Run the pipeline fully offline: `MOCK_LLM=true MOCK_LLM_CASSETTE=... MOCK_EMBEDDING=deterministic`.
- Add a sampling profiler. Suggested: integrate [`pprof-rs`](https://github.com/tikv/pprof-rs) behind a `profiling` feature on the `bench` subcommand to emit a flamegraph/pb per phase, and/or document a `cargo flamegraph` recipe under `scripts/perf/`. Pin CPU affinity (`taskset`) and use `--release` for stable samples; isolate storage in a tmpdir.

### 3. Investigate likely hot paths
Candidate areas to scrutinize once the flamegraph is in hand (each links the crate to inspect):
- **Chunking + tokenization** — [`crates/chunking`](https://github.com/topoteretes/cognee-rs/tree/main/crates/chunking). Token counting (HF/tiktoken) and hierarchy traversal run over the whole document; check for repeated re-tokenization and allocations.
- **Graph extraction / fact extraction** — [`crates/cognify/src/fact_extraction`](https://github.com/topoteretes/cognee-rs/tree/main/crates/cognify/src/fact_extraction). With the LLM mocked to ~0 cost, the surrounding parsing, dedup, and graph-assembly become the visible cost.
- **Embedding pipeline + vector writes** — [`crates/embedding`](https://github.com/topoteretes/cognee-rs/tree/main/crates/embedding) and the vector store (LanceDB default / `BruteForceVectorDB`). Batch sizes, vector copies, upsert/index rebuild cost.
- **Graph store** — default embedded Ladybug adapter writes/reads during cognify and graph-completion search.
- **Serialization & cloning** — `serde_json::Value` round-trips in the structured-output path, `String` vs `&str`, and `Arc`/clone churn in the pipeline task graph ([`crates/core`](https://github.com/topoteretes/cognee-rs/tree/main/crates/core)).
- **Parallelism knobs** — `RayonThreadPool` sizing and `max_parallel_extractions`; verify cores are actually saturated under replay (mocked LLM removes the natural async back-pressure that hid CPU stalls).

### 4. Deliverables
- A committed large-doc fixture + cassette and a documented profiling recipe under `scripts/perf/`.
- A short write-up (e.g. `docs/performance/`) with the flamegraph findings: top N CPU consumers with percentages, before numbers.
- A prioritized list of concrete optimization proposals (each as a follow-up issue/PR), with expected impact.

## Acceptance
- `scripts/perf/run_mock_bench.sh` (or a sibling) can profile a large-document run end-to-end with **no API key**, emitting a flamegraph.
- The findings doc identifies the top CPU bottlenecks with measured percentages and lists prioritized, actionable optimizations.

## Notes / open questions
- Which book/size to standardize on (keep the committed cassette to a reasonable size — large corpora make for large cassettes).
- Whether to gate the profiler behind a `profiling` cargo feature vs. an external `cargo flamegraph` recipe (former is reproducible in CI; latter needs no code).
- Stale-cassette guard: as the README warns, prompt/chunking changes silently degrade to empty-graph fallbacks — the large-doc fixture needs the same node/edge-count sanity check.

Akshats-git:
## Description

The mock-replay bench could already run `add → cognify → search` offline, but it had no profiler, the 50-memory corpus was too small to be CPU-bound and search ran a single query. This adds CPU profiling and per-stage telemetry to the bench, grows the corpus to a full book and documents where the CPU goes.

Closes #18.

## Changes

- **CPU profiler behind a `profiling` feature** (pprof-rs, SIGPROF, no `perf`/root). `--profile-dir` writes a per-phase flamegraph.
- **Per-stage wall-clock telemetry** that reuses the existing tracing spans and records busy (on CPU) vs idle (awaiting) time per phase.
- **Search phase** now runs the no-LLM retrievers (Chunks, Summaries) as well as GraphCompletion, so retrieval shows up in the profile.
- **Graph sanity guard**: the bench reports node/edge counts and fails on an empty or below-`--min-graph-nodes` graph, catching a stale cassette.
- **Large-document fixture**: `build_large_corpus.py` builds a 135-chapter Moby-Dick corpus, with the cassette recorded once and committed so it replays with no key.
- **Findings write-up** in `docs/performance/cpu-profiling.md`.

## Findings

At 50 memories the pipeline is await/IO bound (cognify ~11% on CPU), so the hot paths only appear at scale.

On Moby-Dick (1232 nodes, 2744 edges):

- `add` is the biggest CPU cost: **~57% on CPU**, dominated by ingestion and chunking.
- Cognify's on-CPU time is dominated by graph writes. `cognee.db.graph.query` runs **1061 times for 982 ms**, about **74%** of cognify's on-CPU time, roughly one Cypher execution per graph element.

The main optimisation lead is the graph-write Cypher path. The prioritised list is in the findings doc and will be filed as follow-up issues.

## Reproducing (offline, no key)

```bash
MOCK_LLM=true MOCK_EMBEDDING=deterministic \
  cargo run --release -p cognee-cli --features bench,profiling -- bench \
    --mock-llm --mock-memories scripts/perf/fixtures/large/cassette.json \
    --memories scripts/perf/fixtures/large/memories.json \
    --profile-dir target/perf-profiles/large --min-graph-nodes 1232
```

Recording the cassette is the only step needing a key. See `scripts/perf/README.md`.

## Testing

- `fmt` and `clippy -D warnings` clean.
- The `cognee-cli` test suite passes.
- The bench replays on both fixtures.


crates/cli/src/commands/bench.rs
```diff
@@ -540,12 +670,63 @@ async fn phase_search(
...
+    )
+    .with_session_manager(session_manager)
+    .with_dataset_resolver(Arc::clone(&database) as Arc<dyn IngestDb>)
+    .build();
+
+    let query_text = "What is in the document";
+    // Vector retrievers first (Chunks, Summaries), then the graph retrieval path
+    // (GraphCompletion in context-only mode). Labels feed the per-query logs.
+    let queries = [
+        ("chunks", SearchType::Chunks),
```
GraDKh, line 739: **`search_time` now aggregates 3 queries, breaking the cross-SDK parity comparison.** This loop times `Chunks + Summaries + GraphCompletion` as a single `search_time` (set at line 489), but `docs/performance/python-approach.md:66,81` establishes `search_time` as a percentile-compared metric measuring *one* `cognee.search(...)`. `statistics_percentile_report.py` computes p50/p75/p90/p99 over `search_time` across both SDKs, so the Rust value is now ~3× the Python value and the parity table becomes apples-to-oranges — silently defeating the harness's purpose. If the goal is retrieval coverage in the profile, please time only the representative query as `search_time` (or profile the extras without folding them into the reported metric).

crates/cli/src/commands/bench.rs
```diff
@@ -540,12 +670,63 @@ async fn phase_search(
...
+        ("summaries", SearchType::Summaries),
+        ("graph_completion", SearchType::GraphCompletion),
+    ];
+
+    for (label, search_type) in queries {
+        let request = bench_search_request(query_text, search_type, dataset_name, owner_id);
```
GraDKh, line 742: **The broadened search phase can hard-fail the bench where it previously passed.** `SummariesRetriever::get_context` returns `Err(SearchError::NotFound("missing vector collection: TextSummary_text"))` when the summaries collection is absent (`crates/search/src/retrievers/summaries_retriever.rs:52`). Because the `summaries` query runs before `graph_completion` and errors propagate via `?`, a corpus/config that produced no `TextSummary` nodes (summarization disabled, or a cassette yielding zero summaries) aborts the whole phase with `failed: summaries: …`, `GraphCompletion` never runs, and `success` flips to `false`. The old single-query path passed. Since these retrievers exist only for profiling coverage, consider making them best-effort (log and continue) rather than gating the result.

crates/cli/src/commands/bench_telemetry.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+        if let Some(span) = ctx.span(id) {
+            span.extensions_mut().insert(Timings {
+                busy: Duration::ZERO,
+                idle: Duration::ZERO,
+                last: Instant::now(),
+            });
+        }
+    }
+
```
GraDKh, line 97: **Per-span timing carries pre-phase / cross-phase time into the armed phase.** `on_enter`/`on_exit`/`on_close` accumulate into the span's `Timings` unconditionally, and `arm()` (line 138) clears only the aggregate map, not the per-span `Timings.last` living in span extensions. A span created during `prune`/`setup` (or held open across a phase boundary) whose `last` predates `arm()` will, when entered/closed during `add`, compute `idle += now - last` and fold in seconds of pre-phase time — inflating that phase's `idle_ms` and booking a prior phase's work onto whichever phase the span closes in. For a measurement tool this corrupts the numbers it reports; the module's "relative attribution" caveat covers parallel over-counting, not cross-phase bleed. Consider gating accumulation on `enabled`, or resetting `last` on arm.

crates/cli/src/commands/bench.rs
```diff
@@ -82,11 +85,119 @@ struct BenchStatus {
...
+fn finish_phase_telemetry(_profile_dir: Option<&str>, _phase: &str) {}
+
 /// Round to 3 decimals to match Python's `round(x, 3)` output.
 fn round3(value: f64) -> f64 {
     (value * 1000.0).round() / 1000.0
 }
 
+/// Read `(node_count, edge_count)` from the graph after cognify. Returns
+/// `(0, 0)` (which trips the sanity guard) if the metrics can't be read.
+async fn graph_counts(cm: &Arc<ComponentManager>) -> (i64, i64) {
```
GraDKh: **A graph-metrics read error is reported as a stale/empty-graph failure.** `graph_counts` returns `(0,0)` on any `graph_db()`/`get_graph_metrics()` error, and the sanity guard (line 453) then relabels a successful cognify as `failed: graph sanity: 0 nodes < floor 1 (stale cassette / empty-graph fallback?)` with `success=false`. If cognify built a real graph but the follow-up metrics query errored transiently, a healthy run is reported as failed with a misleading message. Please distinguish metrics-unavailable from a genuinely empty graph (e.g. skip the guard, or emit a distinct diagnostic, when the metrics read itself failed).

GraDKh (changes requested):
Thanks for the thorough profiling work and the write-up — the flamegraph + telemetry split and the committed offline cassette are really nice; a few behavioral issues to address before merge (details inline).

Akshats-git:
@GraDKh Thanks for the recommendations. I have addressed them in my recent commit.

GraDKh:
Thanks a lot for this, @Akshats-git — the profiling harness, the per-stage telemetry, the large-document benchmark, and the CPU-profiling write-up were genuinely useful work and clearly a lot of effort.

Closing this as **superseded** rather than merging, and wanted to explain why so it's not a black box:

Since this branch was opened, `main` has moved ~70 commits ahead, and equivalent profiling/benchmark infrastructure has independently landed there through a separate, newer lineage — `crates/cli/src/commands/bench_telemetry.rs`, the pprof/flamegraph wiring in `bench.rs`, the large-document corpus under `scripts/perf/`, and `docs/performance/cpu-profiling.md` all now exist on `main`. In several places `main`'s version has also had follow-up fixes that this branch predates:

- the `cognee-lib` → `cognee` umbrella-crate rename (#93),
- deterministic class-namespaced entity ids (#57), which changed the documented graph counts (1189 nodes / 2667 edges vs. the older 1232 / 2744 here), and
- a TikToken token-counter pin (#73) that keeps cassette hashing environment-independent.

Because of that, rebasing and merging this branch would mostly revert `main` back to older behavior rather than add new capability. We reviewed the branch hunk-by-hunk to make sure nothing unique was lost:

- The **single-representative-query search timing** here reports `search_time = 0.0` when a search fails; `main` reports the real elapsed-until-error, which is strictly more informative — so we kept `main`'s version.
- The **epoch-based span attribution** in `bench_telemetry.rs` is a genuinely nice idea (confining each span to the phase it was created in). We looked hard at porting just that piece, but in practice the dominant per-stage spans already open and close inside a single armed phase, so the improvement only affects millisecond-scale background-task noise in a table that's already documented as approximate. We're noting it here as a possible future refinement rather than pulling it in now.

None of this reflects on the quality of the contribution — it's purely a timing/overlap issue with work that landed in parallel. Really appreciate you digging into the offline pipeline's performance, and hope you'll send more PRs. 🙏

PR #78, opened 2026-07-13
fix(llm): litellm-parity for OpenAI-compatible adapter (custom endpoints)

GraDKh:
Makes cognee-rs's OpenAI-compatible adapter behave like Python cognee's litellm/instructor path, so cognee works against custom OpenAI-compatible endpoints (e.g. **Baseten**) whose model slug legitimately begins with `openai/`.

## Changes
- **`factory.rs`** — strip exactly one leading litellm-style provider prefix (`baseten/`, `openai/`, `ollama/`, `mistral/`, `gemini/`) instead of the adapter unconditionally stripping `openai/`; `warn!` on provider/prefix mismatch. `baseten/openai/gpt-oss-120b` → `openai/gpt-oss-120b`; `openai/gpt-4o-mini` → `gpt-4o-mini`.
- **`adapters/openai.rs`** — structured output via OpenAI **tool calling** (`tools` + forced `tool_choice`) instead of strict `response_format: json_schema`. (Strict json_schema drives server-side grammar-constrained decoding, which returns HTTP 501 on some backends incl. Baseten gpt-oss-120b.) Bounded retry with a corrective instruction on malformed/empty responses; tolerate missing tool-call `arguments`/`function`.
- Route the direct-construction integration tests + `test-utils` through `build_openai_compatible_adapter` so prefixed models are stripped like production.

## Review findings addressed
#2 (dead-code error branch), #3 (over-broad prefix strip → mismatch warning), #4 (retry-on-empty), #5 (tests bypassing the factory), #7 (missing arguments/function), #8 (identical retry request). **#1** (required-field guarantee): a strict/all-required schema reproduces the 501, so it is intentionally not used; required-field parity is approximated by the retry-with-corrective-instruction loop, and typed-validation retry against `T` is a documented follow-up.

## Verification
- `cargo build`/`clippy` clean; `cargo test -p cognee-llm` 79 lib + new `openai_structured_output.rs` (2) green.
- Mock-cassette bench replays `success:true`.
- Live Baseten smoke (gpt-oss-120b): tool-calling path succeeds end-to-end, **zero HTTP 501/404** (dense-chunk truncation is a separate `max_tokens`/`LLM_ARGS` plumbing gap, not this change).

### DIFF
--- a/crates/llm/src/adapters/openai.rs
+++ b/crates/llm/src/adapters/openai.rs
@@ -147,8 +147,21 @@ impl OpenAIAdapter {
         if self.extra_args.is_empty() {
             return;
         }
+        // Reasoning models (`gpt-5*`/`o1*`/`o3*`/`o4*` on api.openai.com) reject
+        // `max_tokens` and require `max_completion_tokens`. The request body's
+        // output cap is written by `write_max_tokens` as `max_completion_tokens`,
+        // so a bare `max_tokens` coming from `LLM_ARGS` here would land alongside
+        // it and OpenAI rejects a request carrying *both* keys with a 400. Fold a
+        // `LLM_ARGS` `max_tokens` into `max_completion_tokens` (only filling the
+        // gap) so exactly one of the two keys is ever emitted.
+        let reasoning = self.is_reasoning_model();
         if let Some(obj) = body.as_object_mut() {
             for (key, value) in &self.extra_args {
+                if reasoning && key == "max_tokens" {
+                    obj.entry("max_completion_tokens")
+                        .or_insert_with(|| value.clone());
+                    continue;
+                }
                 obj.entry(key.clone()).or_insert_with(|| value.clone());
             }
         }
@@ -224,6 +237,21 @@ impl OpenAIAdapter {
     /// - HTTP 5xx (server errors)
     ///
     /// Errors on HTTP 400 and 401 are returned immediately without retrying.
+    async fn call_api(&self, mut request_body: Value) -> LlmResult<OpenAIResponse> {
+        // Merge configured `LLM_ARGS` (Python `llm_config.llm_args`) into every
+        // chat-completion / structured-output request. Only fills keys the request
+        // does not already set, so explicit parameters win — Python's
+        // `{**self.llm_args, **kwargs}`. Scoped to the chat/structured paths: the
+        // transcription (vision) path calls `send_chat_request` directly so a
+        // graph-extraction `LLM_ARGS` (e.g. a large `max_tokens`) never leaks into
+        // an image-description request.
+        self.apply_extra_args(&mut request_body);
+        self.send_chat_request(request_body).await
+    }
+
+    /// Perform the actual chat-completions HTTP POST, retrying on transient
+    /// network/server errors. Does *not* merge [`extra_args`](Self::extra_args) —
+    /// callers that want the `LLM_ARGS` merge go through [`call_api`](Self::call_api).
     #[instrument(
         name = "llm.api_call",
         level = "info",
@@ -234,12 +262,7 @@ impl OpenAIAdapter {
             cognee.llm.provider = "openai",
         ),
     )]
-    async fn call_api(&self, mut request_body: Value) -> LlmResult<OpenAIResponse> {
-        // Merge configured `LLM_ARGS` (Python `llm_config.llm_args`) into every
-        // chat-completion request. Only fills keys the request does not already
-        // set, so explicit parameters win — Python's `{**self.llm_args, **kwargs}`.
-        self.apply_extra_args(&mut request_body);
-
+    async fn send_chat_request(&self, request_body: Value) -> LlmResult<OpenAIResponse> {
         let url = format!("{}/chat/completions", self.base_url);
         tracing::Span::current().record("url", url.as_str());
         let debug_enabled = std::env::var("COGNEE_DEBUG_LLM_REQUEST")
@@ -434,6 +457,38 @@ impl OpenAIAdapter {
     /// required field or structural constraint the previous response violated —
     /// this threads `T`'s typed validation into the repair prompt, matching how
     /// instructor feeds the validation error back to the model.
+    /// Build a schema-aware validator for the type-erased raw path (which has no
+    /// Rust type to deserialize into).
+    ///
+    /// Enforces that every property named in the schema's top-level `required`
+    /// array is present and non-null. This gives the raw path the same
+    /// required-field guarantee instructor provides for a typed model, *without*
+    /// strict/grammar-constrained decoding (`response_format: json_schema`, which
+    /// 501s on Baseten): a response omitting a required field is fed back into the
+    /// existing corrective-retry loop instead of being accepted or hard-failing.
+    fn schema_required_validator(schema: &Value) -> impl Fn(&Value) -> Result<(), String> + '_ {
+        move |value: &Value| {
+            let Some(required) = schema.get("required").and_then(Value::as_array) else {
+                return Ok(());
+            };
+            let Some(obj) = value.as_object() else {
+                return Err("expected a JSON object".to_string());
+            };
+            for field in required {
+                if let Some(name) = field.as_str() {
+                    match obj.get(name) {
+                        None => return Err(format!("missing required field `{name}`")),
+                        Some(Value::Null) => {
+                            return Err(format!("required field `{name}` is null"));
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            Ok(())
+        }
+    }
+
     fn append_corrective_instruction(request: &mut Value, reason: Option<&str>) {
         if let Some(messages) = request["messages"].as_array_mut()
             && let Some(last_msg) = messages.last_mut()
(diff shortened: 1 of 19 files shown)


crates/llm/src/adapters/openai.rs
```diff
@@ -198,7 +234,12 @@ impl OpenAIAdapter {
             cognee.llm.provider = "openai",
         ),
     )]
-    async fn call_api(&self, request_body: Value) -> LlmResult<OpenAIResponse> {
+    async fn call_api(&self, mut request_body: Value) -> LlmResult<OpenAIResponse> {
+        // Merge configured `LLM_ARGS` (Python `llm_config.llm_args`) into every
+        // chat-completion request. Only fills keys the request does not already
+        // set, so explicit parameters win — Python's `{**self.llm_args, **kwargs}`.
+        self.apply_extra_args(&mut request_body);
```
GraDKh: **[correctness · high] `LLM_ARGS` `max_tokens` collides with reasoning-model `max_completion_tokens`.**

With `provider=openai` + a reasoning model (e.g. `gpt-5-mini`) on api.openai.com and the documented `LLM_ARGS={"max_tokens":16384}`: `write_max_tokens` emits `max_completion_tokens` (~L209), then `apply_extra_args` here finds no literal `max_tokens` key and inserts one → the body carries **both** `max_completion_tokens` and `max_tokens` → OpenAI returns 400 *"Unsupported parameter: max_tokens ... use max_completion_tokens"*. Every `generate()`/structured call fails (also when `options.max_tokens` is `None`). The feature's headline path breaks reasoning models.

*Fix:* don't inject `max_tokens` from extra_args when the reasoning-model branch already wrote `max_completion_tokens` (translate/skip the key).

crates/llm/src/adapters/openai.rs
```diff
@@ -499,109 +522,290 @@ impl Llm for OpenAIAdapter {
...
                 }
             }
         }
 
-        // Try function calling first (works with OpenAI)
+        // Every tool-calling attempt returned a non-empty payload that was
+        // either unparseable or failed validation: surface it rather than
+        // silently swallowing it (the legacy fallbacks would only re-ask a
+        // server that clearly *does* speak tool calling).
+        if let Some((err, raw)) = last_invalid {
```
GraDKh: **[correctness · high] Structured-output now hard-errors on non-JSON instead of falling through to the legacy/JSON-mode fallback.**

Against an OpenAI-compatible endpoint that ignores `tools`/`tool_choice` and answers prose/markdown in `message.content`: content is non-blank and not JSON → every attempt sets `last_invalid`, and after the loop `structured_output_impl` returns `DeserializationError` here **without** ever issuing the legacy function-calling or JSON-mode (`response_format: json_object`) request. On `main` the strict path hit `break` and fell through to JSON mode, which succeeded. Net: extraction against tool-unsupporting providers that used to work now fails outright.

crates/llm/src/adapters/openai.rs
```diff
@@ -499,109 +522,290 @@ impl Llm for OpenAIAdapter {
...
-                    // function calling / JSON mode below, but make the reason
-                    // visible — a silent fallback is how required fields end up
-                    // missing from the model's output.
-                    warn!(error = %e, "strict json_schema request failed; falling back to function/JSON mode");
+                    // Tool calling is unsupported by this model/endpoint (or the
+                    // schema was rejected). Fall back to legacy function calling
+                    // / JSON mode below, but make the reason visible — a silent
+                    // fallback is how required fields end up missing.
+                    warn!(error = %e, "tool-call request failed; falling back to legacy function/JSON mode");
                     break;
```
GraDKh, line 894: **[correctness · high] A transient error on a later tool-call attempt surfaces a stale deserialization error and skips the legacy fallback.**

Attempt 0 returns unparseable arguments (`last_invalid` set); attempt 1's `call_api` returns a transient `Err` and `break`s here. After the loop, `last_invalid` is `Some`, so a `DeserializationError` is returned (L801) instead of the real API/network error, and the legacy fallbacks are never tried — masking the true cause and denying recovery for an only-intermittently-failing server.

crates/llm/src/adapters/openai.rs
```diff
@@ -499,109 +522,290 @@ impl Llm for OpenAIAdapter {
         json_schema: &Value,
         options: Option<GenerationOptions>,
     ) -> LlmResult<Value> {
-        let is_empty_or_non_json = |raw: &str| {
-            let trimmed = raw.trim();
-            trimmed.is_empty() || serde_json::from_str::<Value>(trimmed).is_err()
-        };
+        self.structured_output_impl(messages, json_schema, options, None)
```
GraDKh: **[correctness · high] Removing strict `json_schema` dropped required-field enforcement on the *non-validated* raw path, with no replacement retry.**

`create_structured_output_raw` / `_with_messages_raw` pass `validator: None`, so nothing enforces required fields there. `summarize_one` (crates/cognify/src/summarization/extractor.rs) calls the raw path with `summary_schema`; a well-formed object that omits the required `summary` key now returns `Ok(parsed)` on the first attempt (no strict coercion, no validation retry) → `summary_schema output missing string summary field`. Same loss hits the http-server infer-schema/custom-prompt endpoints. Consider threading a schema-required check + corrective retry into the raw path too.

crates/llm/src/adapters/openai.rs
```diff
@@ -499,109 +522,290 @@ impl Llm for OpenAIAdapter {
...
-                        }
+                    // Prefer a modern `tool_calls[0]`, then a legacy
+                    // `function_call`, then raw `content` (some servers echo the
+                    // JSON directly).
+                    let arguments = choice
+                        .message
+                        .tool_calls
+                        .as_ref()
+                        .and_then(|calls| calls.first())
+                        .map(|c| c.function.arguments.as_str())
```
GraDKh, line 830: **[correctness · med] Empty `tool_calls[0].function.arguments` shadows a JSON payload in `message.content`.**

When a server returns `tool_calls` with `arguments == ""` but echoes the real JSON in `message.content`, this maps arguments to `Some("")`, so the `.or(content)` branch (~L750) never runs; the valid content JSON is ignored and the call fails as blank. Treat an empty/whitespace `arguments` as absent so the content fallback engages.

crates/llm/src/adapters/openai.rs
```diff
@@ -499,109 +522,290 @@ impl Llm for OpenAIAdapter {
...
+                "role": "user",
+                "content": [
+                    { "type": "text", "text": "What's in this image?" },
+                    { "type": "image_url", "image_url": { "url": data_uri } }
+                ]
+            }],
+        });
+        self.write_max_tokens(&mut request_body, Some(max_tokens));
+
+        let response = self.call_api(request_body).await?;
```
GraDKh: **[correctness · med] `extra_args` (`LLM_ARGS`) is merged into the `transcribe_image` vision call, which shouldn't receive them.**

`transcribe_image` calls `self.call_api`, which unconditionally applies `apply_extra_args`. An `LLM_ARGS` param meant for text extraction (e.g. `top_p`, `response_format`, or a reasoning-model `max_tokens`) is now sent to the vision/transcription endpoint, which may reject or mishandle it. Scope `extra_args` to the chat/structured path, not transcription.

crates/llm/src/adapters/openai.rs
```diff
@@ -720,11 +944,23 @@ impl Llm for OpenAIAdapter {
...
+                            )));
+                        }
+                        continue;
+                    }
+                    return Ok(parsed);
+                }
                 Err(e) => {
-                    if attempt + 1 < self.structured_output_retries && is_empty_or_non_json(content)
-                    {
+                    if !last_attempt && is_blank(content) {
```
GraDKh: **[correctness · med] JSON-mode fallback retry condition narrowed from `is_empty_or_non_json` to `is_blank`.**

In the final JSON-mode fallback, a non-empty but malformed response (e.g. truncated mid-object) is now `!is_blank` → returned as an error on the first attempt instead of being retried up to `structured_output_retries` as before. Restore retry-on-invalid-JSON here.

crates/llm/src/llm_trait.rs
```diff
@@ -152,8 +198,18 @@ pub trait LlmExt: Llm {
         T: Serialize + DeserializeOwned + JsonSchema + Send,
     {
         let schema = generate_json_schema::<T>();
+        // Trial-deserialize into `T` so the adapter can retry a well-formed but
+        // incomplete response (e.g. one missing a required field) with a
+        // corrective instruction, rather than failing the whole pipeline.
+        let validator = |value: &Value| -> Result<(), String> {
```
GraDKh, line 214: **[cleanup · perf] Double deserialization into `T` on every successful structured call.**

The validator closure clones the whole parsed `Value` and runs `from_value::<T>` to check validity, then the wrapper deserializes into `T` again at the call site. On the hot cognify graph-extraction path this doubles deserialization + a full clone of a large KG value per call. Have the validator return the deserialized `T` (or validate without a full clone) and reuse it.

GraDKh:
**Code review — 2 findings that couldn't be anchored inline** (their lines aren't in the diff hunks):

**[correctness · high] `crates/llm/src/mock/recording.rs` (and `ThrottleLlm`): missing override of `create_structured_output_with_messages_raw_validated`.**
The wrappers override `create_structured_output_with_messages_raw` but not the new `_validated` variant. In record mode `registry.build_llm` wraps the real `OpenAIAdapter` in `RecordingLlm`; a typed caller hits the default trait impl → the instructor-style **validation-retry is silently bypassed** when a real adapter is wrapped (record + throttle paths). Likely why re-recording couldn't recover missing-field responses. Override `_validated` on both wrappers to delegate to the inner adapter.

**[cleanup] `crates/llm/src/adapters/openai.rs:~831` legacy function-calling retry re-issues an identical request.**
Unlike the tool-calling and JSON-mode loops (corrective instruction + temperature→0 on retry), the legacy function-calling loop `continue`s with a byte-identical `request_body.clone()`, so a deterministic endpoint just returns the same bad output — wasted attempts. Apply the same corrective + temp-0 treatment.

PR #100, opened 2026-07-23
ci(ios): link both iOS targets instead of type-check only

yonkosan:
Follow-up to #38.

`cargo check` → `cargo build` (debug, `CARGO_PROFILE_DEV_DEBUG=0`) for both iOS targets. `df -h /` and `du -sh capi/target` after each build for the disk numbers.

Pipeline assertion changes dropped — superseded by #112.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -19,7 +19,7 @@ env:
 
 jobs:
   ios-check:
-    name: iOS targets (cargo check + Swift parse)
+    name: iOS targets (cargo build + Swift typecheck)
     # macos-14 = Apple Silicon (M1).  Required for aarch64-apple-ios targets.
     runs-on: macos-14
 
@@ -50,24 +50,25 @@ jobs:
       - name: Install protoc (required by lance-encoding build script)
         run: brew install protobuf
 
-      # `cargo check` type-checks the crate for iOS without linking —
-      # avoids the 6–7 GB .a files that a full `cargo build --release` produces
-      # and keeps peak disk usage well under the runner's ~14 GB limit.
-      - name: Check — aarch64-apple-ios (device)
+      # `cargo build` (debug, CARGO_PROFILE_DEV_DEBUG=0) links both iOS targets,
+      # catching linker errors that type-checking alone misses.
+      - name: Build — aarch64-apple-ios (device)
         run: |
-          cargo check \
+          cargo build \
             --target aarch64-apple-ios \
             --manifest-path capi/cognee-capi/Cargo.toml \
             --no-default-features \
             --features sqlite,testing,mock-llm,lancedb
+          df -h /
 
-      - name: Check — aarch64-apple-ios-sim (simulator)
+      - name: Build — aarch64-apple-ios-sim (simulator)
         run: |
-          cargo check \
+          cargo build \
             --target aarch64-apple-ios-sim \
             --manifest-path capi/cognee-capi/Cargo.toml \
             --no-default-features \
             --features sqlite,testing,mock-llm,lancedb
+          df -h / && du -sh capi/target
 
       # Validate Package.swift parses correctly (does not require the xcframework
       # binary — `dump-package` only reads the manifest).
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -11,15 +11,21 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  # Incremental state is useless on CI (fresh checkout every run) and wastes
-  # 2–5 GB of the macOS runner's ~14 GB disk.
+  # Incremental state is useless on CI (fresh checkout every run), so it is
+  # pure disk cost.
   CARGO_INCREMENTAL: "0"
-  # Debug info inflates .a files by ~40 %; strip it on CI to reduce disk pressure.
+  # Strips debug info from cognee-capi itself. This is *not* what keeps the
+  # build small, despite the obvious reading: capi/Cargo.toml sets
+  # `debug = true` per package for the 24 first-party cognee-* crates, and a
+  # named package override beats `profile.dev.debug` (which is all this env var
+  # maps to), so every one of those still compiles with full DWARF. The size
+  # win comes from [profile.dev.package."*"] debug = 0 on the dependency
+  # closure, which is in the manifest and needs no help from here.
   CARGO_PROFILE_DEV_DEBUG: "0"
 
 jobs:
   ios-check:
-    name: iOS targets (cargo build + Swift typecheck)
+    name: iOS targets (link + Swift typecheck)
     # macos-14 = Apple Silicon (M1).  Required for aarch64-apple-ios targets.
     runs-on: macos-14
 
@@ -50,25 +56,57 @@ jobs:
       - name: Install protoc (required by lance-encoding build script)
         run: brew install protobuf
 
-      # `cargo build` (debug, CARGO_PROFILE_DEV_DEBUG=0) links both iOS targets,
-      # catching linker errors that type-checking alone misses.
-      - name: Build — aarch64-apple-ios (device)
+      # Link — not merely type-check — both iOS targets, so an unresolved or
+      # mismatched symbol anywhere in the Rust closure fails here instead of in
+      # a downstream Xcode build.
+      #
+      # `--crate-type cdylib` overrides the `["cdylib", "staticlib"]` list in
+      # capi/cognee-capi/Cargo.toml for this invocation only (the same trick
+      # build_xcframework.sh uses in the other direction, documented at
+      # capi/cognee-capi/Cargo.toml:20-25). The cdylib is the slice worth
+      # building: rustc emits a staticlib by archiving objects *without*
+      # resolving symbols, so a staticlib build runs no linker at all and would
+      # add nothing this job does not already get from `cargo check`. The
+      # shipped staticlib is built separately, in release, by
+      # capi/scripts/build_xcframework.sh.
+      #
+      # The `test -f` guard is deliberate: rustc treats an unsupported crate
+      # type as a warning, not an error, so without it a future toolchain could
+      # quietly downgrade this step to a compile-only check.
+      - name: Link — aarch64-apple-ios (device)
         run: |
-          cargo build \
+          cargo rustc --crate-type cdylib \
             --target aarch64-apple-ios \
             --manifest-path capi/cognee-capi/Cargo.toml \
             --no-default-features \
             --features sqlite,testing,mock-llm,lancedb
-          df -h /
+          test -f capi/target/aarch64-apple-ios/debug/libcognee_capi.dylib
 
-      - name: Build — aarch64-apple-ios-sim (simulator)
+      - name: Link — aarch64-apple-ios-sim (simulator)
         run: |
-          cargo build \
+          cargo rustc --crate-type cdylib \
             --target aarch64-apple-ios-sim \
             --manifest-path capi/cognee-capi/Cargo.toml \
             --no-default-features \
             --features sqlite,testing,mock-llm,lancedb
-          df -h / && du -sh capi/target
+          test -f capi/target/aarch64-apple-ios-sim/debug/libcognee_capi.dylib
+
+      # Its own step with `if: always()`, both deliberate. ENOSPC aborts the
+      # step it occurs in, so a `df` appended to a build step never runs on the
+      # one failure it exists to diagnose.
+      #
+      # `/System/Volumes/Data` is the writable volume the build lands on. `/` is
+      # the read-only system snapshot: across both builds of run 31305066951 its
+      # `Used` column sat unchanged at 9.6Gi while capi/target grew to 5.8G, so
+      # only the container-shared `Avail` there carried any signal (36Gi ->
+      # 34Gi of a 320Gi disk). Peak usage is nowhere near the runner's limit —
+      # the "full builds do not fit" premise this job was originally written
+      # against no longer holds.
+      - name: Disk usage after linking
+        if: always()
+        run: |
+          df -h /System/Volumes/Data
+          du -sh capi/target || true
 
       # Validate Package.swift parses correctly (does not require the xcframework
       # binary — `dump-package` only reads the manifest).
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -73,6 +73,15 @@ jobs:
       # The `test -f` guard is deliberate: rustc treats an unsupported crate
       # type as a warning, not an error, so without it a future toolchain could
       # quietly downgrade this step to a compile-only check.
+      # Baseline for the post-link probe below. Without a starting figure the
+      # probe can report what is left but not what this run consumed, and if
+      # the first link hits ENOSPC the only reading ever taken is the one from
+      # after the failure.
+      - name: Disk usage before linking
+        run: |
+          df -h /System/Volumes/Data || true
+          du -sh capi/target 2>/dev/null || true
+
       - name: Link — aarch64-apple-ios (device)
         run: |
           cargo rustc --crate-type cdylib \
@@ -98,14 +107,16 @@ jobs:
       # `/System/Volumes/Data` is the writable volume the build lands on. `/` is
       # the read-only system snapshot: across both builds of run 31305066951 its
       # `Used` column sat unchanged at 9.6Gi while capi/target grew to 5.8G, so
-      # only the container-shared `Avail` there carried any signal (36Gi ->
-      # 34Gi of a 320Gi disk). Peak usage is nowhere near the runner's limit —
+      # only the container-shared `Avail` there carried any signal. Note those
+      # figures were taken *between* the two builds, so they are mid-run
+      # readings rather than starting headroom — which is what the baseline
+      # step above now records. Peak usage is nowhere near the runner's limit —
       # the "full builds do not fit" premise this job was originally written
       # against no longer holds.
       - name: Disk usage after linking
         if: always()
         run: |
-          df -h /System/Volumes/Data
+          df -h /System/Volumes/Data || true
           du -sh capi/target || true
 
       # Validate Package.swift parses correctly (does not require the xcframework
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -56,23 +56,6 @@ jobs:
       - name: Install protoc (required by lance-encoding build script)
         run: brew install protobuf
 
-      # Link — not merely type-check — both iOS targets, so an unresolved or
-      # mismatched symbol anywhere in the Rust closure fails here instead of in
-      # a downstream Xcode build.
-      #
-      # `--crate-type cdylib` overrides the `["cdylib", "staticlib"]` list in
-      # capi/cognee-capi/Cargo.toml for this invocation only (the same trick
-      # build_xcframework.sh uses in the other direction, documented at
-      # capi/cognee-capi/Cargo.toml:20-25). The cdylib is the slice worth
-      # building: rustc emits a staticlib by archiving objects *without*
-      # resolving symbols, so a staticlib build runs no linker at all and would
-      # add nothing this job does not already get from `cargo check`. The
-      # shipped staticlib is built separately, in release, by
-      # capi/scripts/build_xcframework.sh.
-      #
-      # The `test -f` guard is deliberate: rustc treats an unsupported crate
-      # type as a warning, not an error, so without it a future toolchain could
-      # quietly downgrade this step to a compile-only check.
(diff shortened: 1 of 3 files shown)


ios/Tests/CogneeSDKTests/PipelineTests.swift
```diff
@@ -93,22 +111,53 @@ final class PipelineTests: XCTestCase {
...
-        XCTAssertFalse(searchResult.isEmpty,    "search() must return a non-empty JSON string")
-        XCTAssertNotEqual(searchResult, "null",  "search() must not return null")
+        // The mock graph store is a no-op, so only brute-force vector search
+        // runs — no LLM completion call is made.  The returned chunks must
+        // contain content from the Alan Turing text that was added.
+        XCTAssertFalse(searchResult.isEmpty,
+            "search() must return a non-empty JSON string")
+        XCTAssertNotEqual(searchResult, "null",
+            "search() must not return the JSON null literal")
+        XCTAssertTrue(
```
GraDKh, line 218: **Blocking: this assertion can never pass under the test's mock configuration.**

Traced end to end:

- `search(query:)` passes `optsJSON = nil` (`ios/Sources/CogneeSDK/Cognee.swift:169-174`) → `Value::Null` (`capi/cognee-capi/src/sdk_retrieval.rs:100-105`) → `None => SearchType::default()` (`crates/bindings-common/src/ops/retrieval.rs:77-80`) → `#[default] GraphCompletion` (`crates/search/src/types/search_type.rs:11-12`).
- `demo_cassette.json` has exactly two entries, both `"method": "structured_output"`. `ReplayLlm` matches on a sha256 over messages **plus** the canonicalized schema (`crates/llm/src/mock/cassette.rs:106-118`), so a structured entry can never answer a `generate` call.
- The mock is built by `ReplayLlm::from_path` with no policy override (`crates/components/src/builtins/llm.rs:218-220`), which hardcodes `miss: MissPolicy::EmptyGraph` (`crates/llm/src/mock/replay.rs:53-60`) and returns `content: String::new()` on a miss (`replay.rs:178-183`).
- `prepare_search_result` discards `context` and `graphs` when `!(verbose || only_context)` (`crates/search/src/orchestration/prepare_search_result.rs:63-68`).

So the JSON the test receives is roughly `{"search_type":"GRAPH_COMPLETION","result":{"kind":"Text","data":""},"context":null,"graphs":null,...}`. No query echo, no dataset names, no node payloads — nothing that could contain `turing`/`alan`/`mathematician`. The two assertions above (lines 150-153) still pass, which is why this looks fine at a glance.

Also worth fixing: the comments at lines 143-144 and 147-148 are inaccurate. GRAPH_COMPLETION **always** calls `generate` (it just replays an empty miss here), and the raw chunk hits are *not* returned — the result is `SearchOutput::Text` and context is stripped in non-verbose mode.

Suggested fix — pass explicit opts so the retrieved content actually reaches the response, e.g. `{"searchType":"CHUNKS"}`, or `{"verbose":true}` / `{"onlyContext":true}` if you want to keep GRAPH_COMPLETION and assert on the context.

ios/Tests/CogneeSDKTests/PipelineTests.swift
```diff
@@ -93,22 +111,53 @@ final class PipelineTests: XCTestCase {
...
+        // Decode CogneeAddResult and assert on numeric fields.
+        // On a fresh in-memory store the first add() must ingest ≥ 1 chunk with
+        // nothing deduplicated (the store was empty before warm()).
+        let addParsed = try JSONDecoder().decode(
+            AddResult.self,
+            from: try XCTUnwrap(addResult.data(using: .utf8), "add() returned non-UTF-8 data")
+        )
+        XCTAssertGreaterThan(addParsed.addedCount, 0,
+            "add() must ingest ≥ 1 chunk; got addedCount=\(addParsed.addedCount)")
+        XCTAssertEqual(addParsed.deduplicatedCount, 0,
```
GraDKh: **Blocking: these assertions only hold on the very first run against a given simulator.**

The comment on line 115-116 says "on a fresh in-memory store … the store was empty before `warm()`", but the relational store is neither in-memory nor fresh:

- `configureMockMode` (`ios/Sources/CogneeSDK/Cognee.swift:235-240`) sets exactly five keys — `llm_mock`, `llm_cassette`, `embedding_provider`, `vector_db_provider`, `graph_database_provider`. It never touches `relational_db_url`, `system_root_directory`, or `data_root_directory`.
- The default is `sqlite:./cognee.db?mode=rwc` (`crates/lib/src/config.rs:1103`), resolved CWD-relative by `sqlite_fs_path` (`crates/components/src/builtins/database.rs:29-51`), plus persistent `./.cognee_system` and `./.data_storage`.
- The test builds `Cognee()` with `settingsJSON: nil` and uses the fixed literal `dataset: "demo"`.

On run 2, `existing_data_ids` / `partition_added` (`crates/bindings-common/src/ops/pipeline.rs:111-142`) classify the content-addressed UUID5 item as a duplicate → `addedCount=0, deduplicatedCount=1`, failing line 121 and this line. Cognify then reads the persisted `pipeline_runs` row (`crates/cognify/src/tasks.rs:2087-2097`) and returns `CognifyResult::already_completed` with empty payload vectors (`crates/cognify/src/pipeline.rs:83-89`) → `chunks`/`entities`/`edges` all `0`, failing lines 135-140 as well.

`capi/examples/example_sdk_add.c:151-154` documents this exact hazard and mitigates it two ways: a settings override pointing `relational_db_url` and the roots at a tempdir, **and** a PID-unique dataset name. Doing the same here (or at least one of the two) makes the test order-independent. This matters because `ios/README.md` instructs running `xcodebuild test` before each push — i.e. repeatedly, against the same persistent device.

.github/workflows/ios.yml
```diff
@@ -19,7 +19,7 @@ env:
 
 jobs:
   ios-check:
-    name: iOS targets (cargo check + Swift parse)
+    name: iOS targets (cargo build + Swift typecheck)
```
GraDKh: The job name now advertises "Swift typecheck", which is accurate for the Sources step — but worth being explicit that **nothing in CI typechecks or executes this PR's new test code**.

`swiftc -typecheck` (lines 84-93) covers only the three `ios/Sources/CogneeSDK/*.swift` files. `PipelineTests.swift` gets only `swiftc -parse` (lines 97-100), which builds the AST and diagnoses syntax errors but performs no import resolution, name binding, or type checking. It would not catch a wrong field type, a misspelled property, or a wrong-typed `XCTAssert` argument — and of course does not run the assertions.

There is no other gate: no `xcodebuild test` or `swift test` anywhere in `.github/workflows/`, `scripts/`, or `capi/scripts/`, and `scripts/check_all.sh` has zero Swift references.

To be clear, this is **not** something the PR introduced — the `-typecheck`/`-parse` split predates it and the limitation is documented at lines 95-96 and in `ios/README.md:106-108`. Flagging it because it's the reason the two blocking issues above shipped green: the correctness of this PR rests entirely on a local `xcodebuild test` run. Please confirm the suite passes locally after the fixes (and ideally note it in the PR description).

ios/Tests/CogneeSDKTests/PipelineTests.swift
```diff
@@ -65,6 +65,24 @@ final class PipelineTests: XCTestCase {
...
+        let deduplicatedCount: Int
+    }
+
+    /// Typed mirror of `CogneeCognifyResult` (cognee_sdk.h §D3):
+    /// `{"chunks":N,"entities":N,"edges":N,"summaries":N}`
+    private struct CognifyResult: Decodable {
+        let chunks: Int
+        let entities: Int
+        let edges: Int
+        let summaries: Int
```
GraDKh, line 139: `summaries` is decoded but never asserted — the only declared-but-unasserted field in either mirror. The sibling `AddResult` declares exactly the two fields it asserts and omits the wire's other three keys, so the file's convention is "declare what you assert"; this is the one deviation. (If the intent were to pin the wire contract, the struct would need all seven keys — see the doc-comment note above.)

The comment on line 130 even claims "all counts ≥ 1", and `summaries` is the one count not checked against it.

Asserting it is the better of the two fixes and does hold: the cassette's `SummarizedContent` entry, `enable_summarization` defaulting to `true` (`crates/cognify/src/config.rs:216`), and unconditional per-chunk summary emission (`crates/cognify/src/summarization/extractor.rs:159-200`) guarantee `summaries == 1` — an empty-string summary is not filtered out.

ios/Tests/CogneeSDKTests/PipelineTests.swift
```diff
@@ -65,6 +65,24 @@ final class PipelineTests: XCTestCase {
...
+    // MARK: – Result types
+
+    /// Typed mirror of `CogneeAddResult` (cognee_sdk.h §D3):
+    /// `{"datasetName":"…","added":[…],"addedCount":N,"deduplicated":[…],"deduplicatedCount":M}`
+    private struct AddResult: Decodable {
+        let addedCount: Int
+        let deduplicatedCount: Int
+    }
+
+    /// Typed mirror of `CogneeCognifyResult` (cognee_sdk.h §D3):
```
GraDKh, line 133: This spells the wire shape as a closed four-key object, but `CogneeCognifyResult` has seven keys — `cognee_sdk.h:332-334` and `cognify_result_json` (`crates/bindings-common/src/wire.rs:30-40`) both list `chunks`, `entities`, `edges`, `summaries`, `embeddings`, `alreadyCompleted`, `priorPipelineRunId`.

All seven are unconditionally present: the JSON is hand-built with `json!`, there is no `skip_serializing_if`, and `priorPipelineRunId` serializes as `null` when `None` — so this isn't an honest "stable subset" listing.

The struct decoding only four fields is fine and idiomatic; it's just the comment. Note the `AddResult` comment five lines above lists its complete five-key shape while decoding only two fields, so this also breaks the file's own convention.

Worth fixing precisely because `alreadyCompleted` is the field that explains an all-zero cognify result — directly relevant to the rerun issue below.

ios/Tests/CogneeSDKTests/PipelineTests.swift
```diff
@@ -93,22 +111,53 @@ final class PipelineTests: XCTestCase {
...
-        XCTAssertFalse(addResult.isEmpty,   "add() must return a non-empty JSON string")
-        XCTAssertNotEqual(addResult, "null", "add() must not return null")
+        // Decode CogneeAddResult and assert on numeric fields.
+        // On a fresh in-memory store the first add() must ingest ≥ 1 chunk with
+        // nothing deduplicated (the store was empty before warm()).
+        let addParsed = try JSONDecoder().decode(
+            AddResult.self,
+            from: try XCTUnwrap(addResult.data(using: .utf8), "add() returned non-UTF-8 data")
+        )
+        XCTAssertGreaterThan(addParsed.addedCount, 0,
```
GraDKh: `addedCount` counts ingested `Data` records, not chunks: `add_result_json` emits `"addedCount": newly_added.len()` over a `&[Data]` slice, one per input (`crates/bindings-common/src/ops/pipeline.rs:90-105`), and `AddPipeline::add` returns `Vec<Data>` (`crates/ingestion/src/pipeline.rs:1112`). Chunking happens in cognify — `DocumentChunk` production lives in the cognify crate, and ingestion's `process_by_chunks` is unrelated I/O buffering.

The `cognify() must process ≥ 1 chunk` message 14 lines below *is* about genuine `DocumentChunk`s (`result.chunks.len()`), so the same word means two different things nearby. A failing `addedCount` means ingestion/dedup misbehaved, but the message points at the chunker.

Nit-level, zero behavioral impact — suggest `"add() must ingest ≥ 1 data item; got addedCount=…"`, and the same wording fix in the comment on lines 115-116.

GraDKh (changes requested):
Thanks @yonkosan — this is the right direction. Swapping the `isEmpty` / `!= "null"` smoke checks for typed decoding plus numeric assertions is a real improvement, and the `cargo check` → `cargo build` upgrade genuinely closes a gap: CI now links both iOS targets instead of only type-checking them.

Requesting changes because two of the new assertion blocks cannot pass as written. I traced both through the Rust side:

1. **The search substring assertion has no reachable path to passing** under this test's mock configuration — default `GRAPH_COMPLETION`, a cassette with no `generate` entry, and `context`/`graphs` stripped when neither `verbose` nor `onlyContext` is set.
2. **The add/cognify count assertions assume a fresh store**, but `configureMockMode` never overrides `relational_db_url`, so they only hold on the very first run against a given simulator.

Neither surfaces in CI, because the test file is only `swiftc -parse`d — so the green check on this PR does not cover any of the new code. Details and suggested fixes inline.

Four things I initially suspected turned out to be non-issues once I checked them properly, so **no action needed** on any of these — recording them so they don't get re-raised:

- *Runner disk / deleted 6–7 GB comment* — the ~14 GB budget is still documented at `ios.yml:14-17` (and in `http-parity.yml` / `publish-dry-run.yml`), and the deleted figure described `--release`, which this PR doesn't run. The job passed in 10.6 min on a cold runner.
- *No `capi/target` cache* — real delta is only +2.5–4 min (332s+266s vs 218s+139s on main); the registry-only cache is byte-identical to main, and the repo is public so macOS minutes aren't billed.
- *No `SDKROOT` export / debug instead of `--release`* — `cc-rs` and rustc both fall back to `xcrun --show-sdk-path`, and the release profile configures no LTO/strip (`panic = "abort"` is in `[profile.dev]` too, so it is exercised).
- *Test-private `Decodable` mirrors instead of shipped typed results* — the wire keys are already pinned by `ts/__tests__/add.test.ts`, `python/tests/test_core_ops.py`, and `CogneeAddTest.java`, all of which run in CI. A public Swift result type would be a nice follow-up, but it's out of scope here.

The remaining inline notes are small cleanups.

GraDKh:
@yonkosan — an important heads-up before you spend any time on my review comments: **`main` moved under this PR, and most of what's here has been overtaken.** That's on me, not on you, and I want to be transparent about exactly what happened.

## What landed

#112 (merged today) rewrote `ios/Tests/CogneeSDKTests/PipelineTests.swift` and independently implemented the same direction this PR was taking:

- **Typed decoding** — a `decodeOrFail<T: Decodable>` helper plus `Decodable` result mirrors, replacing the `isEmpty` / `!= "null"` smoke checks. Same idea as your change, with the decode failure reported non-fatally so a regression in a later stage still gets exercised.
- **Store isolation** — `makeIsolatedRoot()` / `isolateStores()` redirect `data_root_directory`, `system_root_directory` and `relational_db_url` at a per-test temp dir before `warm()`. That is precisely the fix for the blocking issue I raised on line 123: `configureMockMode` never mocked the *relational* DB, so the fresh-store count assertions only held on the first run against a given simulator.

The practical consequence: **both blocking findings from my 07-29 review are moot against current `main` — please don't work on them.** The Swift half of this diff is now redundant, and it conflicts with `main` on exactly that file.

I reviewed this PR two days before landing a change to the same file, and I didn't tell you that was coming. That's a process failure on my side and it cost you effort — sorry. If you'd like, I'm glad to add a note crediting this PR in the changelog entry, since it's what put the typed-decoding direction on my radar.

## What is still genuinely unlanded, and worth having

The CI change. `main` still runs `cargo check` for both iOS targets, so nothing links them — your `cargo check` → `cargo build` swap closes a real gap that the Swift-side work doesn't touch. That's about `+7/-8` lines in `.github/workflows/ios.yml` and I'd merge it on its own.

One thing to settle first, and it's the reason `main` is on `cargo check` rather than an oversight: **the disk budget.** The comment you removed records that a full build produces 6–7 GB of `.a` files and that `cargo check` keeps peak usage under the runner's ~14 GB limit. Your version is debug with `CARGO_PROFILE_DEV_DEBUG=0`, which is plausibly much smaller — but the two target builds accumulate in the same `target/` dir, so it needs a number rather than an argument. Could you add a `df -h /` (or `du -sh capi/cognee-capi/target`) after each build step and paste the peak from a real run on your fork? If it doesn't fit, the fallbacks are linking only the simulator target, or cleaning between the two.

## Concretely, if you want to carry it forward

1. Rebase onto current `main`.
2. Drop the `PipelineTests.swift` changes entirely — that resolves the conflict and the review backlog with it.
3. Keep the `ios.yml` change, with the disk numbers from a real run.
4. While you're in there: the title is garbled (`Test/ios strongetest(ios)/ci(ios): … r assertions`) — worth a quick fix.

That leaves a small, single-purpose PR that I'll review the same day you push.

## One thing you may find more interesting

My third review point still stands and now applies to `main`'s own code: **CI never typechecks or executes any of this test file.** It's `swiftc -parse` only — syntax, nothing more — because `@testable import CogneeSDK` needs a built module, which needs the xcframework that CI doesn't have. So #112 just added ~120 lines of test code that no CI job can prove even compiles. Making CI actually typecheck (or run) the iOS tests would be a substantially more valuable contribution than either half of this PR, and you clearly have the context for it. Happy to sketch out what that would take if you're interested.

And no obligation on any of this — if you'd rather not continue, just say so and I'll close it with thanks. Your read on both problems here was correct.

yonkosan:
@GraDKh Disk numbers from CI run #31040981705: after the device build, `/` had 36 GB free; after the sim build, 34 GB free with `capi/target` at 5.8 GB total. I see that the runner is now 320 GB, i suppose there has been an upgrade to the github CI runners— the 14 GB limit the original `cargo check` comment was written against no longer applies.

With that headroom, the debug sim-only xcframework + `xcodebuild test` path is straightforward. Planning to take that on as the next PR.

GraDKh:
Thank you for sticking with this one, @yonkosan — and for going and getting the actual number instead of arguing the point. `df`/`du` from a real run is what unblocked it: **~36 GiB free, 5.8 GB of `capi/target`**, against a comment that asserted a ~14 GB ceiling. That comment had been the reason this job was stuck on `cargo check` since it was written, and it was wrong. You also took the harder path when I told you #112 had overtaken your Swift work — dropping half your own diff and rebasing down to a single-purpose change is not the fun option, and it is the right one.

CI is green on `1f96f541`. I pushed four review follow-ups onto your branch rather than sending you another round-trip, since they all live in the same nine lines:

- **`cargo build` → `cargo rustc --crate-type cdylib`.** `cargo build` emitted both crate types from the `[lib]` list, and neither extra paid for itself: rustc archives a staticlib *without* resolving symbols, so that half ran no linker at all, and the cdylib it did link is not the shipped artifact either (`build_xcframework.sh` builds the shipped staticlib separately, in release). Same link coverage, less output — and a bigger win than I expected: **20m40s, down from 29m13s**, with `capi/target` at 4.0G instead of 5.8G.
- **A `test -f …libcognee_capi.dylib` guard per slice.** rustc reports an unsupported crate type as a *warning*, not an error, so without this a future toolchain could silently reduce the whole step to a compile-only check — exactly the gap this PR set out to close.
- **The disk probe** is now its own `if: always()` step pointed at `/System/Volumes/Data`. It was measuring `/`, the read-only system snapshot — which is why its `Used` column sat at 9.6Gi before *and* after both builds in your run while `capi/target` grew to 5.8G; only the container-shared `Avail` there meant anything. And sitting at the end of a `bash -e` build step, it would never have run on the one failure it exists to diagnose, since ENOSPC aborts the step first. It now reports `/dev/disk3s5`, 269Gi used / 34Gi avail.
- **Two comments corrected**, including one of mine: `CARGO_PROFILE_DEV_DEBUG=0` is close to inert here, because `capi/Cargo.toml` pins `debug = true` per package for the 24 first-party `cognee-*` crates and a named override beats `profile.dev.debug`. The real size win is `[profile.dev.package."*"] debug = 0` on the dependency closure. Also updated `ios/README.md`, which still documented the `cargo check` steps and justified them with the premise you just disproved.

One thing I deliberately left alone: the job is still ~20 min of 10x-billed macOS time on every push with no `capi/target` cache. Every fix for that reaches outside this PR — a 4 GB cache churns against the 10 GB repo budget and the cache-lifecycle work in 02be121d, and a `paths:` filter would be the first in the repo. That is mine to sort out, not a condition on your change.

And yes — please do take the sim-only xcframework + `xcodebuild test` path if you still want it. It is the most valuable iOS work available right now: after this merges, CI links the Rust side properly, but `PipelineTests.swift` is still only `swiftc -parse`d, so ~120 lines of test code that no job can prove even compiles. Happy to sketch the shape if useful.

GraDKh (approved):
Re-reviewing to clear my stale `CHANGES_REQUESTED` from 2026-07-29. Both blocking comments in that review were against `ios/Tests/CogneeSDKTests/PipelineTests.swift`, which #112 overtook and you subsequently dropped from this diff — the change is now scoped to `.github/workflows/ios.yml`, `ios/README.md` and `docs/RELEASE.md`. All 23 check runs are green on `424d38e9`.

Thanks for the two extra passes on the disk probe and the comment accuracy. Approving and merging.

PR #129, opened 2026-08-06
perf(ci): cut average PR wall-clock from ~81 min to ~39 min

GraDKh:
Cuts the wall-clock of an average PR from **~81 min to ~39 min**, measured rather than estimated. Every number below comes from a real run on this branch; run IDs are in the commit messages.

## Where the time was going

A PR fires 5 workflows in parallel, so wall-clock is set by the slowest — HTTP Parity, which took 76–83 min on **every** run across 10 runs and 6 branches. That flatness was the tell: zero cache reuse.

| Workflow | Before | Notes |
|---|---|---|
| **HTTP Parity** | **76–83 min** | the binding constraint |
| CI | 40–65 min | critical path = `lint` → `capi-check` |
| iOS / Community / publish-dry-run | 2.6–7.6 min | never critical |

Two jobs accounted for almost all of it, and in both cases the actual *testing* was nearly free:

- **HTTP Parity**: the `Phase-1a` step took 4572s while every subsequent test phase took 14–28s. That step is really the Docker build. Inside it, three `cargo build --release` invocations cost 1689s + 1350s + 1309s = **72.5 min**.
- **capi-check**: four cold compilations of one dependency graph (475s + 420s + 704s + 704s), while **all 20+ C smoke tests combined take 13.5s**.

### 1. One cargo invocation instead of three (`e2e-cross-sdk/Dockerfile`)

Each `cargo build -p X` resolves features for only its own subgraph, so shared dependencies land on different feature sets and cargo builds an *additional* variant of each rather than reusing what's already in the shared target mount. The third step was rebuilding `proc-macro2`, `syn`, `serde`, `tokio` for a few-hundred-line binary.

Verified with resolve-only `cargo tree -f "{p} {f}"`: `proc-macro2` resolves to `default,proc-macro,span-locations` under `cognee-cli` but `default,proc-macro` under `cognee-telemetry-emit`, and that cascades through `syn` into every proc-macro consumer.

> **`Compiling` lines 1167 → 741. Build 4348s → 2034s. Job 80.8 → 39.0 min.**

Note `resolver = "3"` unifies features across selected packages, so `cognee-cli` is now built with `cognee-vector/testing` + `cognee-components/testing` (via http-server's `dev-mock`). That only *registers* a mock provider; it stays inert unless `VECTOR_DB_PROVIDER=mock`. All LLM, telemetry, provenance and logging parity phases pass, so this is empirically harmless.

### 2. Drop capi's two redundant `cargo check` passes (`capi/scripts/check.sh`)

They duplicated the two CMake `cargo build`s that follow at the same feature sets, and a build subsumes a check.

> **Measured in simultaneous paired runs on two branches, capi cold in both: 2395s → 1704s (−691s, −11.5 min).**

### 3. Cache saves scoped to main + capi cache re-enabled (`ci.yml`)

`capi-check` is CI's critical path and had been deliberately running cold every single run, because saving its cache blew the 10 GB quota and evicted everything else. That tradeoff is no longer necessary:

- all six writers (lint, msrv, wasm, test, ts, java) now save on main only;
- the `capi-check-v2` cache is **1.18 GB**, not the ~3 GB the old comment assumed — dropping the `check-slim` target dir shrank it.

> **capi-check 1704s cold → 1195s warm (−509s, −8.5 min).**

### 4. Drop `github.sha` from the CI concurrency group (`ci.yml`)

With the sha in the key, every push formed a *new* group, so `cancel-in-progress` never had anything to cancel and superseded 40–65 min runs kept occupying runners to completion.

> **Evidence: `ci.yml` has 0 cancelled runs in its last 60; the ref-only-keyed `http-parity.yml` cancels routinely.**

### 5. Remove the inert buildx layer-cache step (`http-parity.yml`)

Nothing ever wrote `/tmp/.buildx-cache` — the compose build declares no `cache-from`/`cache-to`. It restored in 0s and saved in 0s on every run, and no `buildx-http-parity-*` key has ever existed.

## Tried and rejected (documented inline so nobody retries them)

**A ghcr.io registry layer cache.** Measured across two runs at an *identical commit* — the best case a layer cache can get: **2085s cold vs 2034s warm**, against **+179s** of export cost. Net negative. Why zero layers were reused is still unexplained; the likely suspects are the regenerated `Cargo.lock` in the build context and the cache-mount RUN steps.

**`cargo test --no-run` to preserve capi's inline-test coverage.** Cost **819s** — as much as the checks it replaced, because `cargo test` pulls dev-dependencies and rebuilds the lib as a test harness. Worth knowing: that coverage was compile-only anyway, since nothing in CI ever *runs* the capi workspace's unit tests (the root `cargo test --workspace` doesn't reach into `capi/`).

## Caveats and follow-ups (not in this PR)

**Main-only saves interact badly with toolchain churn.** Mid-measurement the rust-cache key's env component moved `adf1a63c` → `b33fdd40` while the dependency hash held identical — `dtolnay/rust-toolchain@stable` picked up a new stable Rust, invalidating every cache repo-wide. With main-only saves, each Rust release now strands all PRs cold until main runs once. **Worth pairing this PR with a pinned toolchain or a scheduled main warm-up.**

**Merged-PR caches are never reclaimed.** Hours after #127 merged, its `refs/pull/127/merge` entries still held **4.61 GB of a 10.35 GB quota** (GitHub only expires them after 7 days). A PR-close cleanup job would reclaim more than the save-scoping in this PR does. I deleted #127's manually while working.

**Other findings, deliberately out of scope:** `community.yml` runs on all PRs with no same-repo skip, duplicating fmt+check+clippy on every in-repo PR (~7.6 min of compute, but it is *not* on the critical path, so zero wall-clock saving); no path filters anywhere, so a docs-only change still triggers the full build; HTTP Parity's remaining 2034s is genuine cold dependency compilation, which a cargo-chef-style prebuilt-dependency base image would attack directly.

## Reviewer notes

- **Shared-runner variance here is large** — identical CMake work ranged 704s→895s, and lint spent 40s vs 234s at the *same* cache key with a confirmed hit. Only paired or replicated measurements are quoted above; single-run CI wall-clock comparisons were too noisy to trust.
- **`Java Bindings Check` has a pre-existing flake** unrelated to this PR and may redden it: `Failed to delete temp directory /tmp/junit…: cognee.db-shm, cognee.db-wal` with `Failures: 0, Errors: 2` — assertions pass, JUnit's `@TempDir` teardown races SQLite's WAL cleanup. Worth its own fix.
- This PR's own CI may run cold, since main's caches predate the toolchain bump and nothing but main writes them now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,7 @@ on:
 # created a brand-new group, so cancel-in-progress had nothing to cancel and a
 # superseded 40-65 min run kept occupying runners until it finished on its own.
 # ios.yml, community.yml and http-parity.yml already key on the ref alone
-# (and do show cancelled runs); CI has 0 cancelled runs in its last 60, which
-# is how this surfaced.
+# (and do show cancelled runs); CI showed none, which is how this surfaced.
 concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
@@ -117,6 +116,14 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: workspace-lint-v6
+          # Save only on main. Branch-scoped copies are what pushed the repo
+          # over GitHub's 10 GB cache quota and kept the LRU rotating (see the
+          # capi-check note below). Branches still *restore* main's caches;
+          # they just stop writing a second copy. Note this only bites on PRs
+          # that change the dependency graph — rust-cache skips the save
+          # entirely on an exact key hit, so dependency-neutral PRs already
+          # wrote nothing.
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       # ort-sys downloads the ONNX Runtime static library to ORT_CACHE_DIR
       # (target/ort-cache) during its build script. We cache it separately so
@@ -246,6 +253,8 @@ jobs:
         with:
           shared-key: workspace-msrv-v1
           cache-on-failure: true
+          # main-only save — see the note in the lint job.
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Cache ORT binary
         uses: actions/cache@v4
@@ -299,6 +308,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           shared-key: wasm-v1
+          # main-only save — see the note in the lint job.
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       # Build first so Cargo.lock exists (it is gitignored) before we pin the
       # runner to the resolved wasm-bindgen version below.
@@ -413,6 +424,8 @@ jobs:
         with:
           shared-key: workspace-test-v6
           cache-on-failure: true
+          # main-only save — see the note in the lint job.
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Cache ORT binary
         uses: actions/cache@v4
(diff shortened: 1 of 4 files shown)


GraDKh:
## Validation evidence (PR checks are not attaching — see below)

GitHub is not creating a `pull_request` check-suite for this PR. `opened`, `reopened` and a real `synchronize` (new SHA, byte-identical content) all failed to produce one, while `vercel` and `delve-auditor` sat at `queued` too. Actions is enabled, there are no rulesets, and `workflow_dispatch` on this branch works fine — so this looks like check-suite processing on GitHub's side rather than anything about these changes.

Both workflows were therefore validated by dispatch against the exact PR tree (`97b0912`):

**CI — run `31128889790`: success, 11/11 jobs green**

| Job | Result |
|---|---|
| Lint, MSRV, wasm32 | ✅ |
| C API Check | ✅ |
| Test (stable), Test (Postgres span lane) | ✅ |
| Python / JS+TS / Java Bindings Check | ✅ |
| Docs + feature lanes | ✅ |
| Test Suite Status | ✅ |

**HTTP Parity — run `31127507503`: success** (all parity phases, including the LLM-gated ones).

### Structural result

`capi-check` is no longer CI's critical path. It was the longest job at **2436s**; on this run it placed **3rd at 1106s**, with `Test (stable)` (1511s) now the longest.

```
before:  lint 344s -> capi-check 2436s   (capi = critical path, ~46 min)
after:   capi-check 1106s                (3rd; Test (stable) 1511s now longest)
```

Note the 41.4 min wall-clock of that run is *not* the steady-state figure — cache state was mixed (lint ran at 954s vs 287s when warm). The durable changes are the capi reduction above and HTTP Parity's 80.8 → 39.0 min.

### Two earlier runs that reported `failure` — both infrastructure, neither a defect

- Run `31127452002`: 8 jobs green, 3 cancelled. I force-pushed an amended commit (trying to nudge PR checks into firing) while it was in flight, orphaning its SHA.
- Run `31128300733`: `Lint` and `MSRV` sat queued for 15 min with **zero steps recorded** and were cancelled at the same instant — they never got a runner, with nothing else competing for capacity.

PR #130, opened 2026-08-07
ci: cache lifecycle follow-ups (scheduled warm-up, PR-close cache reclaim)

GraDKh:
Follow-ups to #129. All checks green on `0b9f934`.

> **Scope note:** an earlier revision of this PR also carried a fix for the Java `@TempDir`/SQLite-WAL flake. That was **reverted** — see [§4](#4-reverted-the-java-tempdirsqlite-wal-fix) — so the Java tree here is byte-identical to `main`. The underlying defect is tracked in #132.

## 1. Warm main's caches on a schedule (Mon+Thu)

#129 made cache saves main-only, so branches can only restore what a main run has written — anything that invalidates main's caches leaves **every** PR cold until main next runs. Two things do, and neither is rare:

- **Toolchain churn.** The jobs install `dtolnay/rust-toolchain@stable`, which sets `RUSTUP_TOOLCHAIN` and therefore overrides `rust-toolchain.toml`'s `channel = "1.91"` — CI really runs current stable (the MSRV job is what covers 1.91). Every stable release rotates the rustc component of every rust-cache key at once; observed going `adf1a63c` → `b33fdd40` while the dependency hash held at `c4307474`.
- **The 7-day idle eviction.** GitHub deletes any cache not *read* for a week, so a quiet week cools the repo by itself.

Pinning CI to a fixed stable was considered and rejected: it would delete the early-warning value that having a separate MSRV lane implies is wanted.

**Mon+Thu rather than daily** — case 1 is already covered on an active repo, since main is pushed most days; case 2 is what needs a floor, and a max 4-day gap sits well under 7. Daily would have run the LLM-backed test lane every night for no additional cache benefit.

Scheduled runs also get their **own concurrency group**. A cron fires with `github.ref == refs/heads/main`, so sharing `ci-${{ github.ref }}` with `cancel-in-progress: true` meant the 03:11 warm-up cancelled any still-running post-merge run — leaving the merge commit with a cancelled check (`notify` is gated on `!cancelled()`, so the aggregator did not even report) and skipping that run's main-only cache saves, the exact opposite of the cron's purpose.

## 2. Reclaim a PR's caches when it closes

GitHub keeps `refs/pull/<n>/merge` caches until they go 7 days unread — merging reclaims nothing. Hours after #127 merged, its entries still held **4.61 GB of a 10.35 GB quota**, and the resulting LRU eviction is what kept the working caches cold. This is the other half of the main-only `save-if` policy: that stops branches *writing* duplicates, this reclaims what PR runs already wrote.

Uses **`pull_request_target`, not `pull_request`**: a fork PR's `pull_request` run gets a read-only token regardless of the permissions block, so the delete would silently fail for exactly the PRs that need it — `community.yml` (the fork lane) saves lint+test caches unrestricted, so a fork writes ~2 GB of its own. The usual `pull_request_target` hazard does not apply: there is no checkout and nothing from the PR is executed, only its number is read.

Hardened after review — three of these failed **silently green**, the worst mode for a janitor job nobody watches:

- `pr_number` (a free-form dispatch string) was interpolated straight into a jq `select(.ref == "…")` predicate, so a quote rewrote the filter instead of being matched by it; `x" or true or .ref=="` would have matched every cache in the repo and deleted main's warm caches wholesale. Now validated `^[0-9]+$`, and nothing is interpolated into jq at all.
- The listing ran in a process substitution, whose exit status bash never checks and `pipefail` does not cover — so a 403, rate limit, 5xx or jq error produced zero lines and reported "no caches found — nothing to do" on a green run. Now written to a file with the status checked.
- Every DELETE error went to `/dev/null` and printed "already gone?", so a blanket 403 read as a successful cleanup of 0 MiB. Now only a 404 is benign; anything else is reported and fails the step.
- `per_page=100` bounded the **repo-wide** listing, not the PR's, with the ref filter applied client-side. The endpoint sorts by `last_accessed_at` desc, so the oldest page out first — exactly the dispatch-an-old-PR case this workflow advertises. Now filters server-side via `?ref=` with `--paginate`.

Verified locally against a stubbed `gh`: happy path, a 404 mid-run, a 403 mid-run, a failed list, an empty list, and six malformed `pr_number` inputs.

**⚠️ Cannot be exercised by its own PR** — it only fires on `pull_request_target: closed`. After merge, please run it once via `workflow_dispatch` against an already-closed PR and confirm the step summary reports a real reclaim. The stub is a simulation, not the live API.

## 3. Stop copying `capi/` into the e2e image

`capi/` is its own cargo workspace, is not a member of the root workspace, is not a path dependency of anything in it, has no `[patch]` pointing at it, and is not referenced by the three binaries the image builds. Copying it did nothing except widen the cache-invalidation surface of the ~34-minute cargo build: editing anything under `capi/` — including `capi/scripts/check.sh`, which is not even Rust — busted the COPY layer and everything behind it. (This bit me during #129.)

Suggestive but single-sample: HTTP Parity ran **31.7 min** on `0b9f934` vs **37.4 min** on the previous SHA, with this as the only relevant difference.

## 4. Reverted: the Java @TempDir/SQLite-WAL fix

An earlier revision added a `SqliteSettle` helper that waited for SQLite's `-wal`/`-shm` sidecars after `Cognee.close()`. It was reverted, because measurement showed **the premise was wrong**.

The design assumed sqlx's pool `Drop` unlinks the sidecars shortly after close, so a brief wait would sequence teardown behind it. It does not: with the ceiling raised to 60 s, three of four `SessionsAdminTest` cases waited the **full 60 s** with the sidecars still present (180.8 s for the class). The connection is held until JVM exit, not released late.

So the helper never removed the condition it existed for in any test that opens the DB — including `SessionsAdminTest`, the one that was actually flaking — while costing ~25 s per suite run (`SessionsAdminTest` 15.69 s with it, 0.550 s without). The javadoc claim that "every other `@TempDir` test releases in single-digit milliseconds" was false; only the tests that never open a relational DB were fast.

The real defect is a **leaked SQLite connection after `Cognee.close()`**, tracked in **#132** with reproduction, measurements, ruled-out causes and ranked candidate mechanisms. It needs a Rust-side fix; the flake should be closed by fixing it, not worked around in the tests.

## Not in this PR

**cargo-chef for HTTP Parity's remaining ~2034 s of cold dependency compilation.** Chef's entire benefit is Docker layer caching — `chef cook` only pays off if that layer is reused — and #129 measured layer caching here producing **zero reuse at an identical commit**, with even trivial apt/rustup layers missing, for reasons still unexplained. Building chef on top of that would buy nothing at real complexity cost.

The alternative — extending `buildkit-cache-dance` to the cargo target dir — uses machinery proven to work here (the ccache hit is confirmed), but a release target dir for 741 crates is multi-GB against a 10 GB quota, and the current `run_id`-based key mints a fresh entry every run. That would recreate the exact quota problem #129 and this PR fix. Diagnose the layer-cache failure first, then choose.

**Deleting the orphaned `cognee-rs-e2e-cache` ghcr package** (left by #129's rejected experiment) needs `delete:packages` scope. Nothing in the repo references it; delete via Packages → `cognee-rs-e2e-cache` → Settings → Delete.


GraDKh:
The reverted `SqliteSettle` change is tracked as the underlying defect in #132 — the SQLite connection is leaked after `Cognee.close()` and the `-wal`/`-shm` sidecars are never released (measured still present 60s after close), so no test-side wait can sequence teardown behind it. The `@TempDir` flake should be closed by fixing that leak rather than worked around here.

GraDKh:
Good catch on the stale metadata — that was accurate. The Java fix was reverted after measurement showed its premise was wrong (sidecars never release: three of four `SessionsAdminTest` cases waited a full 60s with them still present, so no wait can sequence teardown behind it). Title and description now updated to match the diff, with the revert documented in §4 and the underlying connection leak tracked in #132.

PR #133, opened 2026-08-07
perf(ci): build e2e binaries on the runner — HTTP Parity 39 → 9 min

GraDKh:
Moves the e2e Rust build out of the Docker image and onto the runner, where `Swatinem/rust-cache` can actually persist the target dir. **HTTP Parity: 39.0 → 9.1 min warm.**

## Measured

Every number from a real run on this branch; cold and warm measured as separate dispatches.

| | wall clock | build step | cache |
|---|---|---|---|
| Baseline (build inside the image) | 39.0 min | 2034 s | — |
| Runner build, `debug = "line-tables-only"` — cold | 54.1 min | 2837 s | 2.61 GB |
| Runner build, `debug = "line-tables-only"` — warm | 11.3 min | 249 s | 2.61 GB |
| **Runner build, `debug = 0` — cold** | **42.6 min** | 2161 s | **0.92 GB** |
| **Runner build, `debug = 0` — warm** | **9.1 min** | **211 s** | **0.92 GB** |

- **Warm is the common case** — main writes the cache, branches restore it: **39.0 → 9.1 min (−77%)**.
- **Cold is 42.6 min**, ~3.6 min worse than today. It happens after a Rust toolchain bump or a cache eviction, and #130's Mon+Thu warm-up bounds that window to under a day.

## Why the build moved out of the image

The `rust-builder` stage recompiled ~741 crates on **every** run — roughly 34 min of a 39 min job — and nothing carried over, by either available mechanism:

- its cargo target dir lived in a `--mount=type=cache`, which never survives a fresh runner;
- BuildKit layer caching was measured in #129 producing **zero reuse at an identical commit** (2085 s cold vs 2034 s warm, with even trivial apt/rustup layers missing), for reasons still unexplained.

Building on the runner puts the target dir behind `Swatinem/rust-cache` instead — the one caching mechanism this repo has demonstrated works. What remains in the image is the Python venv, the ONNX model fetch and harness assembly: Phase-1a is now **111 s** covering image build, container start *and* the tests.

This is deliberately **not** cargo-chef. Chef's entire benefit is Docker layer caching, which is the thing measured broken here.

## Portability, verified not assumed

The runner builds on `ubuntu-24.04` (glibc 2.39, clearing ort-sys's ≥ 2.38) and the harness runs Trixie (glibc 2.41). Old → new is the compatible direction for glibc and libstdc++ alike.

`ldd` on the built server (printed by the build step, so a surprise dep fails on the runner rather than at container start):

```
linux-vdso.so.1 · libstdc++.so.6 · libgcc_s.so.1 · libm.so.6 · libc.so.6
```

Nothing exotic, and no OpenSSL to satisfy — `reqwest` is `default-features = false` with `rustls-tls`, and the resolved graph for these three binaries contains no `openssl-sys` or `native-tls`. All four runs passed every parity phase, so the ubuntu-built binaries demonstrably execute in the Trixie harness.

## Why `debug = 0`

Scoped to these three harness binaries only; the workspace release profile keeps `debug = "line-tables-only"` for everything shipped.

It was added to fix a real objection: at 2.61 GB the cache was the largest single addition to a 10 GB repo quota already at its ceiling, and **an evicted cache costs a 42-min cold run instead of a 9-min warm one** — so cache size directly buys reliability of the speedup. Stripping debug info cut it to **0.92 GB (−65%)**, dropped the cold build by 11 min, and made the cache save 3.5× faster (50 s → 14 s).

The trade is that harness backtraces lose line numbers. That never affects a released binary, and a harness failure is reproduced by re-running the phase.

Cache key is `v2` deliberately: had it stayed `v1`, a `debug=0` build could have restored the `line-tables-only` tree and thrashed against it, producing a meaningless measurement.

## Also removed

Everything that existed only for the deleted stage: the `Cargo.lock` materialisation step, the buildx setup, the `buildkit-cache-dance`, and the `COPY`s of `crates/`, `examples/`, `python/` and the manifests. That shrinks the build context and removes most of what used to invalidate the image build.

## Known and not addressed here

**The runner build is slower than the identical in-image build was** — 2161 s vs 2034 s (2837 s before `debug=0`). Close enough not to block this, but unexplained. The plausible candidate is a lower ccache hit rate: the Dockerfile tuned `CCACHE_NOHASHDIR`/`CCACHE_BASEDIR` for lbug's shifting `OUT_DIR`, and the runner relies on the repo's launcher config instead. Untested.

**`e2e-cross-sdk/.dockerignore` is inert.** Docker resolves `.dockerignore` at the *context root* (the monorepo root), not next to the Dockerfile — confirmed by `docker buildx build --check` reporting `load .dockerignore → transferring context: 2B`, i.e. an empty one. So its `cognee-rust/target/` exclusion has never applied and local builds ship the whole target dir as context. The fix is renaming it to `Dockerfile.dockerignore`, which BuildKit picks up per-Dockerfile. Kept out of this PR on purpose: changing context size would confound the measurements above.

### DIFF
--- a/.github/workflows/http-parity.yml
+++ b/.github/workflows/http-parity.yml
@@ -125,7 +125,7 @@ jobs:
         with:
           workspaces: cognee-rust -> target
           shared-key: http-parity-release-v2
-          save-if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/perf/e2e-prebuilt-binaries' }}  # TEMP: measurement only, revert before merge
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Cache ORT binary
         uses: actions/cache@v4
--- a/.github/workflows/http-parity.yml
+++ b/.github/workflows/http-parity.yml
@@ -125,7 +125,7 @@ jobs:
         with:
           workspaces: cognee-rust -> target
           shared-key: http-parity-release-v2
-          save-if: ${{ github.ref == 'refs/heads/main' }}
+          save-if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/perf/e2e-prebuilt-binaries' }}  # TEMP: measurement only, revert before merge
 
       - name: Cache ORT binary
         uses: actions/cache@v4
--- a/.github/workflows/http-parity.yml
+++ b/.github/workflows/http-parity.yml
@@ -68,16 +68,8 @@ jobs:
       - name: Stub cognee-rust/.env (required by docker-compose env_file)
         run: touch cognee-rust/.env
 
-      # Cargo.lock is gitignored (workspace policy: see .gitignore:6) but the
-      # rust-builder stage does `COPY cognee-rust/Cargo.lock ./`, so we
-      # materialize it on the host before docker build. `generate-lockfile`
-      # only resolves the dependency graph; it does not compile anything.
-      - name: Generate Cargo.lock (gitignored, required by Dockerfile COPY)
-        run: cargo generate-lockfile --manifest-path cognee-rust/Cargo.toml
-
-      # ubuntu-latest ships with ~14 GB free; lbug's C++ link step needs more
-      # headroom than that once cargo registry, target cache, and Docker
-      # layers are factored in. Mirror ci.yml's Test job cleanup.
+      # ubuntu-latest ships with ~14 GB free; the release build plus the cargo
+      # registry and target cache needs more headroom. Mirror ci.yml.
       - name: Free disk space
         uses: jlumbroso/free-disk-space@main
         with:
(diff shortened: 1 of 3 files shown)


GraDKh:
Addressed in 5c50458 — three of the four were real:

1. **Stale timings** ✅ fixed. The comment still quoted the pre-`debug=0` numbers (54/11); corrected to the measured 42.6/9.1, keeping both sets since the comparison is what motivates the setting.
2. **`ldd … || true` vs "fail loudly"** ✅ fixed, and the best catch here — the comment and the code genuinely contradicted each other. Replaced with a real gate (executable check on all three binaries, which catches a copy that silently didn't happen), and `ldd` is now labelled as what it actually is: informational. It runs on the runner, so it *cannot* prove the harness image provides those libraries — only the parity phases can. The old comment overclaimed.
3. **Local-build instructions** ✅ clarified, though the premise was wrong: run from the cognee-rust repo root and `mkdir -p dist` produces `cognee-rust/dist/`, which is exactly what `COPY cognee-rust/dist/…` resolves to against the monorepo-root context. The docs just never said *where* to run them, so I've made that explicit.
4. **curl retries** ✅ added — sensible for a build-time network fetch. Note the stated justification isn't accurate: there is no `curl --retry` anywhere else in this repo (`grep -rE 'curl .*--retry' scripts/ .github/workflows/ e2e-cross-sdk/` is empty), so it's a new convention rather than matching an existing one.

PR #134, opened 2026-08-10
ci: make the e2e dockerignore effective and reclaim branch-scoped caches

GraDKh:
The two known gaps recorded in #130 and #133, plus the fixes from a high-effort review pass. CI hygiene only; no product code.

## 1. `e2e-cross-sdk/.dockerignore` was inert

Docker resolves `.dockerignore` relative to the **context root**, and the context here is the monorepo root one level above this repo (`context: ../..`). A file next to the Dockerfile was therefore never read. Confirmed, not assumed — `docker buildx build --check` reports:

```
#4 [internal] load .dockerignore
#4 transferring context: 2B done      ← an empty ignore list
```

**Fix:** renamed to `Dockerfile.dockerignore`, which BuildKit honours per-Dockerfile, and inverted from a blacklist to **deny-all + allow-list** (one `!` entry per `COPY`). The context is a whole multi-repo dev directory, so a blacklist has to name every artifact dir and sibling checkout that ever appears there — and the old one had already fallen behind. The Dockerfile's `COPY` set is small and fixed.

> **Correction to an earlier version of this description.** It claimed this "became load-bearing once #133 moved the Rust build to the runner", because a multi-GB `cognee-rust/target/` now sits in the context on CI. **That was wrong.** BuildKit derives its context include-patterns from the `COPY` sources, so paths no `COPY` references are never transferred, ignored or not — measured directly: the allow-list build transfers **1.14 kB**, and an excluded `target/` contributed nothing either way. The real benefit is the metadata walk over a large local tree, plus the guarantee that a future `COPY` cannot silently start pulling in an artifact dir. Smaller than advertised, still correct to fix.

**Verified** against a synthetic context in the real layout: all ten `COPY`s resolve (build exits 0), transfer drops to 1.14 kB, `cognee-rust/target/` fails to `COPY` with `"/cognee-rust/target": not found`, and `__pycache__` inside an allowed tree is absent from the image.

**Plus a CI path contract** (in the spirit of `ci/assert-bindings-layout.sh`): a step asserting `Dockerfile.dockerignore` exists and no inert `.dockerignore` sits beside it. The original bug was invisible for months because its only symptom was a slower build. Tested in a simulated checkout layout in all three states.

## 2. `cache-cleanup.yml` only reclaimed PR caches

It handled `refs/pull/<n>/merge` only. Caches under `refs/heads/<branch>` outlived the branch until 7-day expiry — during this work they were deleted by hand twice, most recently a **0.92 GB** release tree after #133.

**Fix:** a `delete` trigger, covering **branches and tags**.

### Review findings addressed

**Tags do have caches, and my first version skipped them.** I had written "Tags never have caches of their own" and not checked it. `capi-release.yml`, `ts-prebuild.yml` and `java-prebuild.yml` all run on `push: tags:` with unrestricted `rust-cache` across a platform matrix, so `refs/tags/<tag>` holds GB-scale entries. Re-pushing a release tag left several GB pinned to the quota while the run exited green claiming nothing to reclaim — the exact LRU eviction this workflow exists to prevent. Tags are now reclaimed like branches.

**A replay path.** `delete` fires once per ref and cannot be re-run, so anything it missed was unreachable: a post-job cache save landing *after* the cleanup listed the ref, a ref deleted by an action using `GITHUB_TOKEN` (no workflow run), or a ref deleted before this existed. `workflow_dispatch` now takes a full `ref` as well as `pr_number`, validated to a `refs/` shape so a typo fails loudly instead of matching nothing against an exact-match API filter.

**Transient DELETE failures retry.** A batch of branch deletions fans out into concurrent runs that can trip the secondary rate limit, and a 403 previously failed the job with no retry path. 403/429/rate-limit responses now back off and retry; a genuine failure prints the exact `gh workflow run … -f ref=…` to replay.

The concurrency group stays **per-ref on purpose** — a shared group looks like the fix, but GitHub keeps only one running plus one pending run per group and *cancels the rest*, so a ten-branch cleanup would silently skip most of it.

**The safety argument is now stated per ref shape.** "Once a PR is closed nothing can restore from its ref" holds for `refs/pull/<n>/merge` but not for `refs/heads`/`refs/tags`, which come back via Restore branch, a re-push, or a stacked PR's base. In those cases the cost is a cold rebuild, not lost data — named explicitly as a deliberate tradeoff.

**Renamed** the workflow, job and step, which still said "closed PRs" while branch and tag deletions ran under them; a failing branch cleanup pointed maintainers at a PR that did not exist.

## Verification

The cleanup script was exercised against a stubbed `gh` across 12 paths:

| case | result |
|---|---|
| tag deletion | reclaims `refs/tags/v0.2.0`, exit 0 |
| branch deletion | reclaims `refs/heads/<branch>`, exit 0 |
| PR close | unchanged, exit 0 |
| dispatch with `ref=` | reclaims any ref, exit 0 |
| dispatch with a bad ref shape | exit 1, clear message |
| dispatch with neither input | exit 1 |
| unexpected `ref_type` | exit 1 |
| transient 403 ×2 then success | retries, exit 0 |
| hard 500 | exit 1, prints the replay command |
| 404 mid-run | benign, exit 0 |
| failing list | exit 1 (no silent green) |
| empty result | exit 0 |

An earlier attempt had a bug — a bad `.replace()` in a patch script meant the event-dispatch block never landed, so branch deletions fell through to PR validation and the summary hit `LABEL: unbound variable`. This matrix is what caught it.

## Note

Neither change can be exercised by its own PR: `delete` fires on branch deletion and `pull_request_target` on close. Deleting this branch after merge is itself the live test — the step summary should report a real reclaim.


GraDKh:
Two of three were real and are fixed in `037ce71`:

1. **Concurrency group collision** ✅ real, and I'd missed it. `github.event.ref` is the unqualified name on a delete event, so a branch and a tag both named `v0.2.0` shared one group — and since GitHub keeps only one running plus one pending run per group, a simultaneous branch+tag deletion could silently drop a cleanup. Now keyed on `ref_type/ref`.

   I did **not** take the suggested `github.ref`, though: on a `delete` event that is not the deleted ref (it no longer exists), so it would collapse *every* deletion into one shared group and reintroduce exactly the batch-cancellation hazard the comment above it warns about. `format('{0}/{1}', ref_type, ref)` disambiguates without collapsing.

2. **Unquoted `${REF}`** in the printed retry command ✅ fixed. Note the stated justification isn't accurate — git refnames cannot contain spaces (nor `~ ^ : ? * [ \\`), so the copy/paste break described can't occur. Quoting is free and correct for the characters that *are* legal, so I applied it anyway.

3. **Deny-all + deep `!` exceptions** ❌ does not hold for BuildKit. Verified twice:
   - minimal case: `*` plus `!a/b/c/wanted.txt` three levels deep → `COPY` resolves, 121 B transferred, while sibling `a/b/junk.txt` still fails `"/a/b/junk.txt": not found`;
   - full synthetic layout mirroring the real tree → all ten real `COPY` sources resolve, build exits 0, 1.14 kB transferred, and `cognee-rust/target/` correctly fails.

   The directory-pruning behaviour you describe is real for some ignore-matcher implementations, but BuildKit's continues walking a pruned directory when a child exception exists. Left as-is.

PR #139, opened 2026-08-11
test(ci): run the pgvector and single-database Postgres suites for the first time

GraDKh:
## The problem

Three instances of the same failure mode: a Postgres-gated suite that reports `ok` forever without executing.

| | Cases | Why it never ran |
|---|---|---|
| `pgvector_span_instrumentation.rs` | 3 | The `pgvector-spans` lane ran `cargo test -p cognee-vector --test pgvector_span_instrumentation` with **no `--features`**. The target is `#![cfg(feature = "pgvector")]` and `cognee-vector` declares no `default` features, so the file compiled away: `running 0 tests`, exit 0, green — since the day the lane was added, while paying for an idle `pgvector/pgvector:pg16` container. |
| `pgvector_integration.rs` + 2 inline adapter cases | 26 | No job ever set `PGVECTOR_TEST_URL`. |
| `provenance_batch.rs` (2 of 3), `pg_shared_db_single_stack.rs`, `pg_full_stack_e2e.rs` | 4 | No job ever set `TEST_POSTGRES_URL`. |

Verified on `4db8bbf3`, before any change:

```
$ cargo test -p cognee-vector --test pgvector_span_instrumentation -- --nocapture
running 0 tests
test result: ok. 0 passed; 0 failed; ...
```

`git log -S '#![cfg(feature = "pgvector")]'` confirms the gate has been there since the initial commit and `pgvector` has never been a default feature — broken from the start, not a regression.

## Real defects the newly-running tests found

**1. The three span cases were missing a precondition (test bug).** All three failed on their first ever execution:

```
thread 'upsert_emits_pgvector_span' panicked at crates/vector/tests/pgvector_span_instrumentation.rs:51:10:
upsert: StorageError("Query Error: error returned from database: relation \"DocumentChunk_text\" does not exist")
test result: FAILED. 0 passed; 3 failed;
```

They call `index_points` without creating the collection. **Concluded: the test is wrong, not the adapter.** Every `VectorDB` impl in the workspace requires the collection to exist first — `BruteForceVectorDB` returns `CollectionNotFound`, LanceDB fails to `open_table`, pgvector's INSERT hits a missing relation — the entire shared `tests/common` conformance suite calls `create_collection` first, and the production callers in `cognee-cognify` all do `has_collection` → `create_collection` before indexing (`tasks.rs:2870`). Python's `create_data_points` auto-creates instead (`PGVectorAdapter.py:278`), but that is a difference in *where the responsibility sits*, not in observable pipeline behaviour. Fixed by establishing the precondition in `make_adapter()`, with `expect` rather than `.ok()?` so a create failure fails the test instead of degrading into the skip path this lane exists to detect.

**2. `reset()` orphaned collection tables and poisoned the next run (test-hygiene bug).** The two inline `shared_db_migration_tests` cases shared the `PGVECTOR_TEST_URL` database and called a `reset()` that dropped `_vector_collections` and the `seaql_migrations*` tables — but not the collection tables it did not know about. `UUID_f`, `Meta_f` etc. survived, invisible to `list_collections`, so the integration suite's cleanup could not delete them:

```
thread 'test_uuid_round_trip' panicked at crates/vector/tests/common/mod.rs:197:48:
StorageError("Execution Error: error returned from database: relation \"UUID_f\" already exists")
```

Only reachable once both targets run in one lane against one server — i.e. it was latent until this PR. Fixed by giving each case its own throwaway database via `create_temp_postgres_db` (the #136 pattern), which also removes the `#[serial]` attributes. Proven idempotent by running the whole lane sequence twice back-to-back.

**3. `pg_full_stack_e2e` required a model no CI job downloads.** It demanded `COGNEE_E2E_EMBED_MODEL_PATH` + `COGNEE_E2E_TOKENIZER_PATH` (BGE-Small ONNX). Nothing in `scripts/` or `.github/` fetches it, and the workspace moved to OpenAI embeddings by default. The ONNX pair is now optional; without it embeddings come from the same OpenAI-compatible account as the LLM (`Settings::resolve_embedding_inputs` already falls endpoint/key back to `llm_*`). **No assertion was weakened** — the test asserts ComponentManager wiring plus non-empty graph/vector output, and nothing depends on which backend produced the vectors. It also now uses a throwaway database: its "non-empty" assertions were satisfiable by leftovers from an earlier run, and its cleanup ran `delete_graph()` against whatever database `TEST_POSTGRES_URL` named — destructive on a developer's own instance.

Two smaller finds: `cargo test -p cognee-database --features postgres` fails the file's third (SQLite) case with `has no supporting driver`, because that crate declares no default features — the lane passes `--features sqlite,postgres`. And the span tests' skip message (`DB_PROVIDER!=postgres — skipping …`) did not match the guard's variable-agnostic `[A-Z_]+ not set .* skipping` glob, so it is reworded.

## Proof the suites execute

All six targets, locally against `pgvector/pgvector:pg16`:

```
--test pgvector_integration          -> running 24 tests   test result: ok. 24 passed; 0 failed;
--lib                                -> running 34 tests   test result: ok. 34 passed; 0 failed;
--test pgvector_span_instrumentation -> running 3 tests    test result: ok. 3 passed; 0 failed;
--test provenance_batch              -> running 3 tests    test result: ok. 3 passed; 0 failed;
--test pg_shared_db_single_stack     -> running 1 test     test result: ok. 1 passed; 0 failed;
--test pg_full_stack_e2e             -> running 1 test     test result: ok. 1 passed; 0 failed;
```

Named cases, zero `... not set — skipping` lines:

```
test upsert_emits_pgvector_span ... ok
test search_emits_pgvector_span ... ok
test delete_emits_pgvector_span ... ok
test pgvector_adapter::shared_db_migration_tests::pgvector_coexists_with_relational_migrator_in_shared_db ... ok
test pgvector_adapter::shared_db_migration_tests::pgvector_purges_legacy_row_from_default_table_on_upgrade ... ok
test postgres::upsert_batch_with_duplicate_ids_dedups_keeping_last ... ok
test postgres::temp_database_supports_back_to_back_initialize ... ok
test relational_pgvector_pggraph_coexist_in_one_database ... ok
test pg_full_stack_add_and_cognify ... ok
```

**33 cases that had never executed once now do.** CI job logs are linked in the first review comment.

## The guard: generalized, not copied

`assert_pggraph_suite_ran.sh` → `scripts/ci/assert_pg_suite_ran.sh`, with `--label` / `--log … --min-tests N` (repeatable) / `--require-test NAME` (repeatable). **Generalized rather than copied** because its skip-marker matching was already variable-agnostic — only the count floor and the inline test names were hard-coded, so parameterising two things was a smaller change than a third 130-line copy, and one script cannot drift across six YAML call sites. All three lanes call it; the pggraph invocation is behaviour-preserving.

Every check was exercised against captured logs, positive and negative:

| Scenario | Result |
|---|---|
| pggraph real run (**nextest** format, 32 + 24) | passes — no regression for the existing lane |
| pgvector real run (**libtest** format, 24 + 34 + 3) | passes |
| single-database real run (3 + 1 + 1) | passes |
| `--features` dropped → `running 0 tests`, cargo exits **0** | `::error::only 0 tests passed in zero.log, expected at least 3` |
| URL unset → 24 skips, cargo exits **0** | `::error::PgVector tests skipped — the suite never reached a live Postgres` |
| `DB_PROVIDER` unset (interleaved mid-line: `test delete_… ... DB_PROVIDER not set to postgres — skipping …`) | caught — confirms the unanchored match is load-bearing |
| log missing / unreadable | `::error::… is missing or empty` (grep's exit 2 must not read as "clean") |
| required test name absent | `::error::required test … did not run (or did not pass)` |
| `--min-tests` not following its `--log`, or a `--log` with no floor | usage error, so a misordered invocation cannot leave a log unchecked |
| `--require-test 'a.*b'` | rejected — a bare name, never a pattern |

## Lane structure

**`pgvector-spans` → `pgvector-postgres`** ("Test (Postgres pgvector lane)"): runs all three `cognee-vector` targets with `--features pgvector,testing` against the container it already had, and exports `PGVECTOR_TEST_URL` alongside the existing `DB_*`. Folding these in does **not** falsify #128's reason for keeping the lanes apart — that was specifically about per-method *span* instrumentation being deferred for `pg_graph_adapter`, a statement about the graph adapter, not about test lanes. The note is preserved on the graph lane, where it still applies, and the "span lane" name is no longer a misnomer.

**New `pg-single-db-postgres`** ("Test (Postgres single-database lane)") for the `TEST_POSTGRES_URL` suites. A sibling rather than more steps in the pgvector lane, because it builds `cognee` with full default features (onnx/ort, lancedb, pdfium) — the long pole — and sequencing it behind the small `cognee-vector` build would add that compile to the stage's wall clock instead of overlapping it. A second service container is free (public image, no secrets).

**Not folded into the `test` lane**, which was tempting — one container, no extra build, and it would sweep in the `DB_*` suites too. Rejected: that lane runs `cargo nextest --workspace`, where per-test processes defeat the `#[serial]` the 24 shared-database pgvector cases depend on (the exact 33-of-59 failure #136 hit), and a job-wide `DB_PROVIDER=postgres` would redirect every config-reading test in the workspace.

`cargo test`, not nextest, in the pgvector lane, for the same `#[serial]` reason. `--nocapture` is mandatory throughout: a skipped case still prints `ok`.

## Merge gate

Both keyless lanes are mirrored in `community.yml` for fork PRs, gated to forks with `if: …head.repo.full_name != github.repository` like `pggraph-postgres`, and its aggregator accepts `skipped` for them. All three lanes are in **both** aggregators' `needs:` list **and** their `needs.*.result` expression — verified by parsing the YAML and diffing the two sets. `pg_full_stack_e2e` is deliberately absent from the community mirror rather than allowed to skip there: its skip line is exactly what the guard treats as a failed lane, and "the key is missing" is indistinguishable from "the URL never arrived" in a log.

## Load-bearing check

Two deliberate breakages, both caught, both reverted:

```
# 1. dropped the collection attribute from the upsert span
assertion `left == right` failed
  left: None
 right: Some("DocumentChunk_text")
test upsert_emits_pgvector_span ... FAILED

# 2. made PgVectorAdapter::delete_points a silent no-op
assertion `left == right` failed
  left: 2
 right: 1
test test_delete_points ... FAILED
```

## Left undone, on purpose

The same never-executed problem exists for the **other** `pg_test_url()` consumers — `crates/ingestion/tests/{integration_deduplication,tenant_isolation,image_audio_add}.rs` (13 cases) and `crates/database/tests/migration_compat.rs` (3). They are outside this PR's two named variables and would need their own lane and their own triage of first-run failures. Recorded in `.claude/CLAUDE.md` so the next person finds them rather than rediscovering them.

## Verification

`bash scripts/check_all.sh` → `All checks passed!` (fmt, `check --all-targets`, `clippy -D warnings`, C API, Python, TS; Java skipped locally for want of a JDK, as it does on this machine before the change).

https://claude.ai/code/session_01F2Re5GvvGmBLW89J7HhKj6


GraDKh:
## CI job logs — the suites executed on the runner

All 20 checks green. Quoted from the job logs of run [31483518000](https://github.com/topoteretes/cognee-rs/actions/runs/31483518000), not from a local reproduction. **Zero `... not set — skipping` lines across all three lanes** (`grep -cE '[A-Z_]+ not set .* skipping'` → `0`, `0`, `0`).

### [Test (Postgres pgvector lane)](https://github.com/topoteretes/cognee-rs/actions/runs/31483518000/job/93755034282) — 3m30s

The lane that reported green while running zero tests, before and after:

```
# before (4db8bbf3): cargo test -p cognee-vector --test pgvector_span_instrumentation
running 0 tests
test result: ok. 0 passed; 0 failed; ...
```

```
running 24 tests
test result: ok. 24 passed;

running 34 tests
pgvector_coexists_with_relational_migrator_in_shared_db ... ok
pgvector_purges_legacy_row_from_default_table_on_upgrade ... ok
test result: ok. 34 passed;

running 3 tests
test upsert_emits_pgvector_span ... ok
test search_emits_pgvector_span ... ok
test delete_emits_pgvector_span ... ok
test result: ok. 3 passed;

PgVector suite verified: 61 tests passed against a live Postgres across 3 log(s), 5 named case(s) confirmed, no skip markers.
```

### [Test (Postgres single-database lane)](https://github.com/topoteretes/cognee-rs/actions/runs/31483518000/job/93755034307) — 16m26s

```
running 3 tests
test postgres::upsert_batch_with_duplicate_ids_dedups_keeping_last ... ok
test postgres::temp_database_supports_back_to_back_initialize ... ok
test result: ok. 3 passed;

running 1 test
test relational_pgvector_pggraph_coexist_in_one_database ... ok
test result: ok. 1 passed;

running 1 test
test pg_full_stack_add_and_cognify ... ok
test result: ok. 1 passed;

Postgres single-database suite verified: 5 tests passed against a live Postgres across 3 log(s), 4 named case(s) confirmed, no skip markers.
```

`pg_full_stack_add_and_cognify` is the case that needed a BGE-Small ONNX model no job downloads. It ran a real `add` + `cognify` through `ComponentManager` against relational + pgvector + Postgres-graph in one throwaway database, with embeddings from the `OPENAI_KEY` account.

### [Test (Postgres graph lane)](https://github.com/topoteretes/cognee-rs/actions/runs/31483518000/job/93755034309) — 3m46s

No regression from generalizing the guard — same lane, same nextest logs, now via `assert_pg_suite_ran.sh`:

```
Summary [   1.506s] 32 tests run: 32 passed, 0 skipped
Summary [   0.158s] 24 tests run: 24 passed, 0 skipped
PgGraph suite verified: 56 tests passed against a live Postgres across 2 log(s), 2 named case(s) confirmed, no skip markers.
```

### Merge gate

`Test Suite Status` (keyed aggregator) and `Community Checks Status` both pass, with all three lanes in their `needs:` list and `needs.*.result` expression. The three `community.yml` mirrors report `skipping`, which is the intended result on an in-repo PR and is what the community aggregator accepts.

**Net: 33 cases that had never executed once now run on every PR, and cannot go green while skipping or while compiling to zero tests.**

Issue #132, opened 2026-08-07
SQLite connection leaked after Cognee.close(): -wal/-shm sidecars never released


GraDKh:
## Summary

After `Cognee.close()`, the SQLite relational connection is not released. Its `-wal`/`-shm` sidecars remain on disk indefinitely — measured still present **60 s** after close, and apparently held until process exit.

This was found while chasing an intermittent Java test failure. The flake turned out to be a symptom; the leak is the actual defect. Because the handle machinery lives in `crates/bindings-common`, this is likely **not** Java-specific.

## Measured evidence

Reproduced locally on macOS (aarch64), `java-check` test suite, using the branch's own harness. With a diagnostic wait inserted after `close()`:

| Ceiling on the wait | Result |
|---|---|
| 5 s | sidecars still present in `CogneeAddTest`, `DatasetsTest`, `CogneeAsyncTest`, and 3 of 4 `SessionsAdminTest` cases |
| 10 s | still present |
| **60 s** | **still present** — `SessionsAdminTest` took 180.8 s for 4 tests, 3 of them waiting the full ceiling |

So this is not slow asynchronous teardown that a wait can sequence behind. The files never go away while the JVM lives.

Two useful negatives that narrow it:

- **Constructing a handle does not open the DB.** `CogneeLifecycleTest.constructCloseRoundTrips` (construct → `close()`, no op) and `CogneeConfigTest` finish in ~3–100 ms with no sidecars. The leak requires an op to have actually opened the relational DB (e.g. `warm()`, `users()`, `notebooks()`).
- **It is not universal within a class.** 3 of 4 `SessionsAdminTest` cases leak, so it depends on which ops ran, not merely on the class.

## Reproduction

```bash
# macOS note: Homebrew's openjdk is keg-only, so `which java` / `java_home`
# report no JDK even when one is installed. Point JAVA_HOME at it explicitly.
export JAVA_HOME=/opt/homebrew/opt/openjdk
export PATH="$JAVA_HOME/bin:$PATH"

cargo build --manifest-path java/cognee-java-jni/Cargo.toml
export COGNEE_JAVA_LIB_PATH="$PWD/bindings/target/debug/libcognee_java.dylib"

cd java && mvn -Dtest=SessionsAdminTest test
```

[...]

PR #140, opened 2026-08-11
fix(teardown): close the graph, vector, ONNX and HTTP-client resources, not just the relational pool

GraDKh:
Follow-on to #135, which fixed the **relational** pool only. `ComponentManager::close()` released the relational connection and left every other component in its cache — on the documented premise that they "release everything on drop". They do, but **the cache is the last strong reference, so a slot that is never emptied is a destructor that never runs.**

`Refs #132` (that issue is already closed by #135).

## How this was scoped: four backend families audited, one found clean

Each family was audited independently with real measurements (`lsof -p`, `ps -M`, on-disk file counts) rather than code reading.

| family | leaks? | finding |
|---|---|---|
| **Postgres pools** | **yes** | `PgGraphAdapter` and `PgVectorAdapter` each own a **private** sqlx pool with no close of any kind — measured against a real PG 16. Three untuned pools per process. |
| **Embedded graph** (ladybug/kuzu) | **yes** | `LadybugAdapter` had no `close()`, and `GraphDBTrait` had **no teardown member at all**, so no caller could close a graph store polymorphically. Holds a write lock plus an un-checkpointed `-wal`. |
| **Clients / misc** | **yes** | ONNX session leaves **7 OS threads + ~32 MB** past close; embedding/LLM keep-alive sockets survive (incl. a second, easily-missed LLM pool inside the transcriber); an in-flight telemetry POST is silently dropped at shutdown (**0 of 1 delivered**). |
| **Embedded vector** (LanceDB) | **no** | Measured *not* to leak fds/threads per adapter. Left alone deliberately, with a comment saying so — Lance installs a process-lifetime CPU thread pool that no `close()` can reclaim, and that is not a per-adapter leak. |

The audit also caught that the doc comment in `component_manager.rs` asserted the **opposite of reality** about non-relational components; it is corrected here.

## Verification

Everything below I ran myself on this branch, not taken on trust:

| check | result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean |
| `ladybug_close` (embedded graph) | **4 passed** |
| `pgvector_close` (live PG 16 on :5433) | **2 passed** |
| `pg_graph_close` (live PG 16 on :5433) | **3 passed** |
| **mutation test** — `LadybugAdapter::close` stubbed to `Ok(())` | **3 of 4 tests go red**; restored → 4 green, no diff |

That last row is the one that matters. A green suite proves nothing here — every suite in #135 was green *while* leaking 18 files. Neutering the close turns `close_checkpoints_and_unlinks_the_wal`, `close_is_idempotent` and `operations_after_close_fail_with_a_closed_error` red, so the tests genuinely guard the behaviour.

## Contract changes — read before merging

Recorded in `CHANGELOG.md`. These are user-visible:

1. **`close()` is no longer a cheap relational reset.** It now takes every cache slot out and closes what is closable.
2. **A closed graph/vector store rejects later operations**, rather than silently reconnecting. The closed state lives inside the shared inner handle (mirroring where sqlx keeps its flag), so surviving `Arc` clones fail their next op instead of reopening.
3. New trait members `GraphDBTrait::close` and `VectorDB::close`, both defaulting to a **no-op** — meaning "this backend owns nothing that needs releasing". Out-of-tree implementors are unaffected.

## Relies on CI rather than local verification

- The full `Test (stable)` / binding-check lanes.
- `scripts/ci/assert_pggraph_suite_ran.sh` is extended so a silently-skipped Postgres suite fails the lane instead of passing green — worth a look, since a skipping test that reports success is exactly how the original leak stayed invisible.
- The HTTP-server shutdown and telemetry-flush paths have unit tests here, but `http-parity` is what exercises them end-to-end.

## Provenance

Produced by a 5-phase agent workflow (scope → 4 parallel audits → plan → implement → verify). The verify phase stalled on rebuild timeouts, so I did that phase by hand — the table above is mine. Two earlier "failures" during my verification were my own errors, not defects: wrong Postgres credentials, then a wrong feature flag (`pggraph` vs `postgres`). Notably the tests failed loudly on a wrong-but-present URL rather than skipping.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -773,17 +773,27 @@ jobs:
       # same conclusion ("the lib target holds 20-odd unrelated unit tests, so a
       # count floor would not notice these two disappearing") and the
       # `pggraph-postgres` lane below still passes `--min-tests 1` for it.
+      # Pool-teardown regression tests (topoteretes/cognee-rs#132) — see the
+      # PgGraph lane for why a named `--test` invocation is required.
+      - name: Run PgVectorAdapter pool-close tests
+        run: |
+          set -o pipefail
+          cargo test -p cognee-vector --features pgvector,testing \
+            --test pgvector_close -- --nocapture 2>&1 | tee pgvector-close.log
+
       - name: Assert the suite really executed
         run: |
           bash scripts/ci/assert_pg_suite_ran.sh --label PgVector \
             --log pgvector-integration.log --min-tests 20 \
             --log pgvector-lib.log         --min-tests 1 \
             --log pgvector-spans.log       --min-tests 1 \
+            --log pgvector-close.log --min-tests 2 \
             --require-test pgvector_coexists_with_relational_migrator_in_shared_db \
             --require-test pgvector_purges_legacy_row_from_default_table_on_upgrade \
             --require-test upsert_emits_pgvector_span \
             --require-test search_emits_pgvector_span \
-            --require-test delete_emits_pgvector_span
+            --require-test delete_emits_pgvector_span \
+            --require-test close_releases_the_pool_while_the_adapter_is_still_held
 
   # ── Stage 2.5: single-database Postgres stack (depends on lint, parallel with `test`) ──
   # The `TEST_POSTGRES_URL` suites, which no CI job ever supplied that variable
@@ -1060,13 +1070,25 @@ jobs:
       # this lane reporting green while every case inside it skips. The logic
       # lives in a script rather than inline so this lane and its community.yml
       # mirror cannot drift, and so it can be tested against captured logs.
+      # Pool-teardown regression tests (topoteretes/cognee-rs#132). This lane
+      # invokes individual `--test` targets, so a new file is otherwise compiled
+      # by nobody and run nowhere — which for a leak test means the leak comes
+      # back silently.
+      - name: Run PgGraphAdapter pool-close tests
+        run: |
+          set -o pipefail
+          cargo test -p cognee-graph --features postgres,testing \
+            --test pg_graph_close -- --nocapture 2>&1 | tee pggraph-close.log
+
       - name: Assert the suite really executed
         run: |
           bash scripts/ci/assert_pg_suite_ran.sh --label PgGraph \
             --log pggraph-integration.log --min-tests 30 \
             --log pggraph-lib.log         --min-tests 1 \
+            --log pggraph-close.log --min-tests 3 \
             --require-test pggraph_coexists_with_relational_migrator_in_shared_db \
-            --require-test pggraph_purges_legacy_row_from_default_table_on_upgrade
+            --require-test pggraph_purges_legacy_row_from_default_table_on_upgrade \
+            --require-test close_releases_the_pool_while_the_adapter_is_still_held
 
   # ── Stage 3: Language binding checks (depend on lint, parallel with `test`) ──
   # Gated on `lint` (not `test`) so the binding builds run concurrently with the
--- a/.github/workflows/community.yml
+++ b/.github/workflows/community.yml
@@ -275,13 +275,25 @@ jobs:
             --lib --success-output immediate --color never \
             2>&1 | tee pggraph-lib.log
 
+      # Pool-teardown regression tests (topoteretes/cognee-rs#132). This lane
+      # invokes individual `--test` targets, so a new file is otherwise compiled
+      # by nobody and run nowhere — which for a leak test means the leak comes
+      # back silently.
+      - name: Run PgGraphAdapter pool-close tests
+        run: |
+          set -o pipefail
+          cargo test -p cognee-graph --features postgres,testing \
+            --test pg_graph_close -- --nocapture 2>&1 | tee pggraph-close.log
+
       - name: Assert the suite really executed
         run: |
           bash scripts/ci/assert_pg_suite_ran.sh --label PgGraph \
             --log pggraph-integration.log --min-tests 30 \
             --log pggraph-lib.log         --min-tests 1 \
+            --log pggraph-close.log --min-tests 3 \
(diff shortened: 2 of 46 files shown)


GraDKh:
All four comments were legitimate and are fixed in `d68ff1b` — the cleanest review pass this branch has had.

**1. Lock ordering defeated the relational-first guarantee** ✅ real, and the best catch. `teardown()` documented that the relational close runs first so a bounded teardown still reaps the SQLite sidecars — but it acquired the database write lock **last**, after six other `write().await` calls that can each block on a concurrent accessor. A caller with a timeout (the CLI) could spend its whole budget waiting on the graph lock and never start the one close with an on-disk consequence. Database slot goes first now.

This is the previous review's finding 6 in a subtler form: I fixed the *close* order and left the *lock acquisition* order defeating it. Neither my own audit nor a 29-agent review pass caught that.

**2. `bench.rs` reverted main's #137** ✅ real, and it exposed a pattern. It had dropped `dataset_delete_time_s`, `dataset_delete` and the delete phase, while main's `cli_bench.rs` — which this branch never included, and so never ran — still asserts those keys.

That prompted auditing every file both this branch and main touched, which found **two more silent reverts**: `pg_graph_adapter.rs` and `pgvector_adapter.rs` had reverted main's #136 `with_temp_db` helpers (87 and 99 lines). All three rebuilt from main with only this branch's deltas re-applied.

**3. `.wal` vs `-wal`** ✅ real. Ladybug's sidecar is `.wal`; `-wal` is what the SQLite relational pool leaves. Confirmed against the test matchers, which key on `extension() == "wal"`.

**4. Unconditional "post-close operations fail"** ✅ real. It cannot hold for the no-op default, and the unit test for that default asserts operations still succeed. Now scoped to backends that actually close something.

---

Also removed, found while verifying: this branch's own `close_right_after_a_query_still_releases_the_sidecars`. It was written against the `settle_sqlite_pool` this branch no longer carries (wait *before* the close); under main's `drain_sqlite_pool` (wait *after*, with a 400 ms stall window) its assertion only holds if the straggler's return lands inside that window — **3 failures in 5 back-to-back runs**. Main's `close_waits_for_a_straggling_connection` and `close_reaps_a_connection_stranded_after_the_pool_was_closed` already cover the mechanism deterministically, so it was a flaky duplicate. A comment records why it is gone.

Verification widened to the targets that would have caught the reverts — `cli_bench` and both adapters' `--lib` targets, none of which this branch owns. fmt, clippy `-D warnings`, **109 tests across 12 suites**, pgvector/pggraph against a live PG 16, and `connection_pool` stable over 6 consecutive runs.

GraDKh:
Second pass: **no new comments** on the four fixes from `d68ff1b` — all confirmed.

It did surface one previously-missed item in unchanged code, and it is a fair catch: `run_command` snapshotted `DEFERRED` *before* `block_on`, fixing the release decision before the command had even run. If a `DeferGuard` outlives the load but not the command, that snapshot is stale and the release gets skipped even though teardown is no longer deferred — precisely the silent never-torn-down case this module exists to remove.

I checked reachability before changing anything: the only caller (`run_sequence.rs:176`) holds its guard across the whole dispatch, so both readings agree today and there is no live bug. Fixed anyway in `cf154b4` — the failure direction is a *skipped release*, the fix is moving one load inside the closure, and not depending on a single caller's scoping is worth more than the line saved.

CLI suites green after the change (teardown 2, cli_bench 3, lib 25, plus the rest).

GraDKh:
All four review findings are fixed in `fbb5505`. Two of them were regressions this branch introduced while fixing earlier findings, which is worth stating plainly.

**1. The relational-first guarantee still was not delivered — and my previous attempt made it worse.** `cf154b4` moved the database slot's `take()` to the front and left its `close()` behind six more `write().await` calls. Since `versioned_accessor!` holds a slot's write lock across the whole `init_*().await`, a concurrent warm can block the next acquisition for seconds; under the CLI's 5 s bound the budget expires while blocked on (say) the embedding lock, the future is dropped, `cognee_database::close` never runs, and the sidecars survive. Worse than the original, because the slot was already emptied — so no later `close()` could reach that pool at all.

I had also written a comment asserting the guarantee was delivered. Fixed properly: the relational close is now fully sequenced — take, close, drop — before any other lock is requested.

**2. `defer_teardown` suppressed the flush as well as the close**, so `run-sequence` delivered **zero** telemetry events for an entire sequence file — the exact defect this module's header claims to have fixed, reintroduced by my `run_command` wrapper on the deferred path only. Split out as `flush_telemetry()` and run on both paths.

The gap that hid it was coverage: the existing test only exercised the non-deferred path. Added `a_deferred_teardown_still_flushes_the_step_it_dispatched` and mutation-tested it — **fails with the flush skipped, passes with it restored**.

**3. `Notified` was not registered before the counter read.** `notified()` only builds the future; registration happens on first poll, which was inside the `timeout`. A POST completing in between called `notify_waiters()` with nobody enqueued (it stores no permit, by design), so the flush slept out its whole budget before returning success — up to 500 ms on every CLI exit, every binding `close()`, and the SIGTERM handler. `tokio::pin!` + `enable()` makes the existing comment true.

**4. The `services` mutex was held across the flush.** Holding it across `cm.close()` is deliberate and documented; extending it over the flush is not, and it stalled any other thread's `services()` for up to the flush timeout. Dropped before the flush.

Also noted from the review's non-findings, not addressed here: `release`'s `strong_count == 1` rule means a GC finalizer racing an in-flight op still leaks the relational sidecars — the documented tier trade-off, but the one case where `release` provably cannot recover the resource. And `a_store_that_never_closes_does_not_hang_the_shutdown` burns a real 10 s per CI run; `tokio::time::pause` would keep the assertion and drop the wall time. Both are follow-ups rather than blockers.

Verification: fmt, clippy `-D warnings`, **131 tests across 13 suites**, pgvector/pggraph against a live PG 16.

PR #142, opened 2026-08-19
feat(mcp): add serverless cognee-agent lifecycle adapter

netbrah:
## Summary

- establish the `cognee-agent` multicall command surface
- add explicit private-state and model configuration boundaries
- normalize, bound, and redact official agent lifecycle hook events
- keep the runtime serverless: stdio MCP and transient workers only

## Current verification

- `CARGO_BUILD_JOBS=2 cargo test -p cognee-mcp --no-default-features --test protocol --test cli --test events --test redaction`
- `CARGO_BUILD_JOBS=2 cargo test -p cognee-mcp --features runtime --test config`
- 31 focused tests passed; one subprocess helper is intentionally ignored

## Draft scope still landing

Durable spool and generation fencing, cross-process lease and ledger, idempotent engine integration, memory tools over stdio MCP, recovery controls, and the Linux release workflow will be added as reviewable commits. No HTTP service, Python runtime, resident daemon, or listening socket is introduced.


GraDKh:
Hi @netbrah !
Thanks for taking time to contribute to the project!
Can you explain your usage scenario where to you need the serverless agent?

GraDKh:
Hi @netbrah, gentle nudge on this one — no rush at all, just making sure it hasn't fallen off your radar.

I'd still really like to hear about the usage scenario behind the serverless agent. This is a large change (~28k lines, and it reaches into 15 existing crates plus two migrations), so understanding the motivating use case first lets us scope the review properly and give you useful feedback instead of a pile of nitpicks.

Two mechanical notes for whenever you pick it back up:

- The branch now conflicts with `main` in 9 files (it's ~30 commits behind), so a rebase will be needed before we can evaluate it properly. One thing to watch for while rebasing: your new migrations are numbered `m20260914_000002_external_event_id`, which sorts *before* `m20260915_000001_pipeline_run_claims` that has since landed on `main` — they'll need renumbering.
- CI hasn't reported anything here yet because workflow runs from first-time contributors need manual approval on our side — that's on us, not you. For what it's worth, I checked the branch locally: it builds, `fmt` is clean, and the `cognee-mcp` tests pass (79 with default features, 206 with `runtime`). The only gate failure is `cargo clippy -- -D warnings`, on a single dead-code item — `validate_diagnostic_snapshot` in `crates/mcp/src/reference/delta.rs`.

Thanks again for the effort you've put into this!

PR #151, opened 2026-08-26
perf(search,graph): stop FEEDBACK/CODING_RULES full scans, fix the neighborhood default

GraDKh:
Follow-up to #148, which left two items out deliberately.

## 1. `FEEDBACK` and `CODING_RULES` still full-scanned the graph

`FeedbackRetriever::recent_interaction_ids` and `CodingRulesRetriever::load_rules`
called `get_graph_data()`, destructured the edge half into `_`, and filtered the
nodes by type in Rust — every node row *and* every edge row parsed to keep a
handful of typed nodes.

**`get_filtered_graph_data` cannot express what these two test, so this does not
use it.** Both predicates are a case-insensitive *substring* match over five keys
(`type`, `node_type`, `kind`, `label`, `labels`), over JSON strings and string
arrays. Only `type` is a column; the other four arrive inside the `properties`
blob. `get_filtered_graph_data` is an exact `IN (…)` on `id`/`name`/`type`, and
`PgGraphAdapter` rejects any other attribute name outright — so pushing this
predicate through it would either drop every node whose label lives under
`kind`/`labels`, or fail the query. It is not a superset you can post-filter; it
is a subset.

Instead: a new `GraphDBTrait::get_candidate_nodes_by_label(needle)`, specified as
a loose *candidate* read — implementations may return a superset and callers must
re-apply the exact predicate, but nothing matching may be missing and **no edge
row may be read**.

| Backend | Behaviour |
|---|---|
| `PgGraphAdapter` | one `SELECT … FROM graph_node`, filtered by a `position($1 IN lower(…)) > 0` clause per label key — the `type` column plus `properties->>'<key>'` for each of `NODE_LABEL_KEYS`, built from that constant so it cannot drift. A documented superset of the Rust predicate; still a seq scan, but a handful of rows instead of two whole tables. (Revised in `5491e2c` — the first version matched `properties::text` wholesale, which swept in every chunk of prose containing the word "rule". See the review-response comment.) |
| `LadybugAdapter` | node query only, exact predicate on the parsed rows; the edge traversal never runs |
| trait default | `get_graph_data().0` — no worse than the status quo, keeps out-of-tree backends compiling |

The Rust predicate stays as a post-filter in both retrievers and is now one
shared `cognee_graph::node_label_contains` instead of two copies, so the needle
pushed into the query cannot drift from the one re-applied to the rows.

`MockGraphDB` now overrides `get_candidate_nodes_by_label` and logs it — and also
logs `get_filtered_graph_data`, which ignores its filters and delegates to
`get_graph_data`, so without its own entry no test could tell a pushed-down read
from a full scan.

## 2. The default `get_neighborhood` violated its own contract

The doc promises "every edge whose endpoints are both in that resolved set"; the
default collected edges *during* the BFS, keeping only those incident to the
current frontier, and so omitted every edge between two nodes discovered at the
same depth — the whole outermost layer. Same bug `MockGraphDB` had before #148,
fixed the same way: resolve the id set by undirected BFS, then return the induced
subgraph. Phase 2 re-reads the edges incident to every resolved node (reusing what
the walk already fetched) because the default can only see the graph one node at a
time.

Two behaviours change, both **towards** the real backends:

- `depth == 0` now returns seed-to-seed edges instead of none, matching all three
  overrides. The doc note claiming otherwise is gone.
- Edges are read via `get_edges`, not `get_connections`, which drops
  `relationship_name` — previously every edge came back with an empty relationship
  name and parallel edges collapsed under the `EdgeKey` dedup.

Direction is still passed through untouched, so a backend whose `get_edges`
rewrites direction (Ladybug) is still wrong on the default path — it overrides the
method, and the doc comment now says plainly what the default can and cannot
guarantee. `PgGraphAdapter`, `LadybugAdapter` and `MockGraphDB` are untouched.

## Tests

Ten new cases; **each guard verified to fail when its fix is reverted**. Two are
equivalence pins that pass either way and are labelled as such in place.

Observed failures on revert:

```
default_neighborhood_returns_the_induced_subgraph
  left:  [("a","b",""), ("a","c","")]
  right: [("a","b","r1"), ("a","c","r2"), ("b","c","r3")]
default_neighborhood_includes_seed_to_seed_edges_at_depth_zero
  left: []   right: [("a","b","r1")]
default_neighborhood_preserves_relationship_names_and_direction
  left:  [("b","a","")]
  right: [("b","a","authored"), ("b","a","reviewed")]
coding_rules_retriever_does_not_scan_the_whole_graph
  rule loading must go through the narrowed read, got ["get_graph_data"]
feedback_retriever_does_not_scan_the_whole_graph
  interaction lookup must go through the narrowed read, got ["get_graph_data"]
```

## Verification

- `cargo fmt --check`, `cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo test -p cognee-graph --features ladybug,postgres,testing --all-targets` — 155 passed.
- `cargo test -p cognee-search` — only `temporal_retriever_single_month_query` fails, the known non-deterministic LLM interval-extraction case that fails identically on a clean baseline.
- `pg_graph_integration` run against a live PostgreSQL 17 — **34/34**, including the new `test_get_candidate_nodes_by_label`, so the raw SQL is exercised rather than assumed.
- `cognee-http-server`, `cognee-cognify`, `cognee-delete`, `cognee-visualization` lib suites green (they hold `GraphDBTrait` doubles that use the changed default).

**Reviewer note:** `get_candidate_nodes_by_label` adds public API surface to
`GraphDBTrait`. It is non-breaking (it carries a full-scan default, following the
same convention as `get_degree_one_nodes` / `get_id_filtered_graph_data`), but the
shape of the method is the one judgement call in this PR and is worth a look.
It exists because `get_filtered_graph_data` could not express the predicate:
that method is an exact `IN` match on `id`/`name`/`type` only, and the real
predicate is a case-insensitive substring across five keys, four of which live in
the `properties` blob.

https://claude.ai/code/session_01Y2KsXSUu7FfnJdqMF3jppu

### DIFF
--- a/crates/graph/src/ladybug.rs
+++ b/crates/graph/src/ladybug.rs
@@ -1302,6 +1302,48 @@ impl GraphDBTrait for LadybugAdapter {
         Ok(metrics)
     }
 
+    /// Read only the node half of the store and apply
+    /// [`node_label_contains`](crate::node_label_contains) in Rust.
+    ///
+    /// Ladybug keeps a node's extra properties as a serialised JSON *string*
+    /// column, so the four non-`type` label keys are not addressable from
+    /// Cypher here the way they are from Postgres' jsonb. The win over the
+    /// default is therefore narrower but still real: `MATCH (a:Node)-[r:EDGE]->(b:Node)`
+    /// is never run, so the whole edge table stays unread. Because the filter
+    /// runs on fully parsed rows it is the exact predicate, not a superset —
+    /// callers re-applying it is harmless either way.
+    async fn get_candidate_nodes_by_label(&self, needle: &str) -> GraphDBResult<Vec<GraphNode>> {
+        let nodes_query = "MATCH (n:Node) RETURN n.id AS id, n.name AS name, n.type AS type, n.properties AS properties";
+        let node_results = self.execute_query(nodes_query)?;
+
+        let mut nodes = Vec::new();
+        for row in node_results {
+            if row.len() >= Self::NODE_QUERY_COLUMN_COUNT {
+                let mut node_data = NodeData::new();
+                if let Some(id_str) = row[0].as_str() {
+                    node_data.insert(Cow::Borrowed("id"), json!(id_str));
+                }
+                if let Some(name_str) = row[1].as_str() {
+                    node_data.insert(Cow::Borrowed("name"), json!(name_str));
+                }
+                if let Some(type_str) = row[2].as_str() {
+                    node_data.insert(Cow::Borrowed("type"), json!(type_str));
+                }
+                if let Some(props_str) = row[3].as_str() {
+                    node_data.insert(Cow::Borrowed("properties"), json!(props_str));
+                }
+                let parsed_node = self.parse_node_data(node_data)?;
+                if !crate::node_label_contains(&parsed_node, needle) {
+                    continue;
+                }
+                if let Some(id_str) = parsed_node.get("id").and_then(|v| v.as_str()) {
+                    nodes.push((id_str.to_string(), parsed_node));
+                }
+            }
+        }
+        Ok(nodes)
+    }
+
     async fn get_filtered_graph_data(
         &self,
         attribute_filters: &HashMap<Cow<'static, str>, Vec<serde_json::Value>>,
--- a/crates/graph/src/lib.rs
+++ b/crates/graph/src/lib.rs
@@ -38,7 +38,9 @@ pub mod mock;
 
 pub use error::{GraphDBError, GraphDBResult};
 pub use formatted::get_formatted_graph_data;
-pub use traits::{EdgeKey, GraphDBTrait, GraphDBTraitExt, NodeTruthState};
+pub use traits::{
+    EdgeKey, GraphDBTrait, GraphDBTraitExt, NODE_LABEL_KEYS, NodeTruthState, node_label_contains,
+};
 pub use types::{EdgeData, GraphEdge, GraphNode, NodeData};
 
 #[cfg(feature = "ladybug")]
--- a/crates/graph/src/mock.rs
+++ b/crates/graph/src/mock.rs
@@ -87,7 +87,8 @@ impl MockGraphDB {
     /// Get a snapshot of the call log — the names of methods invoked on
     /// this mock in invocation order.
     ///
-    /// Currently records `"get_graph_data"`, `"get_nodeset_subgraph"`,
+    /// Currently records `"get_graph_data"`, `"get_filtered_graph_data"`,
+    /// `"get_candidate_nodes_by_label"`, `"get_nodeset_subgraph"`,
     /// `"get_neighborhood"`, `"get_node_truth_state"`, and
     /// `"set_node_truth_state"`.
     pub fn get_call_log(&self) -> Vec<String> {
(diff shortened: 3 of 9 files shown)


GraDKh:
## Review round 1 — addressed in `5491e2c`

One finding was a real defect in the optimisation itself; the other three were
contract/ergonomics problems on brand-new public API. All four fixed.

**1. The PG filter matched the whole property blob (medium — the important one).**
`position($1 IN lower(properties::text))` is a valid superset of
`node_label_contains`, which is all the contract demanded, but a useless one:
the needles are ordinary English words. `"rule"` matched every `DocumentChunk`
whose `text` mentions "rule"/"rules", so on the 171k-paper corpus this PR cites
the query would drag thousands of full-text rows over the wire for
`is_rule_node` to discard — a full scan replaced by a slightly smaller full
scan, with a "handful of rows" claim in the PR body that only held for needles
that never occur in prose. **The claim was wrong and I've corrected it above.**

Now one `properties->>'<key>'` extraction per label key, built from
`NODE_LABEL_KEYS` so the pushed-down predicate cannot drift from the Rust one.
`type` is tested both as a column and as a JSON key, because `parse_node_row`
merges `properties` *over* the columns — so a writer that put `type` in the blob
is what the Rust predicate sees, and the column-only clause would have missed it.

**2. `get_neighborhood`'s direction caveat understated what changed (low).**
Phase 2 reads the incident edges of *both* endpoints of every internal edge, so
a backend whose `get_edges` reports the queried node as the source — Ladybug's
does — now sees each internal edge twice with opposite directions, and the
`EdgeKey` dedup cannot collapse them (for a well-behaved backend those two
tuples are two genuinely different edges that both have to survive). The old
frontier-only walk emitted it once with one arbitrary direction. Such a backend
**MUST** override, which Ladybug does, so nothing in tree regresses — but the
doc now says that in those words instead of "the emitted direction would be
wrong". Not repairable in the default: direction is information only the backend
has.

**3. A doc guarantee contradicted the code four lines below it (low).**
`get_candidate_nodes_by_label` claimed "no edge row is read at all" of *every*
implementation, immediately above a default that calls `get_graph_data()`. Not
reading edges is a property of the overrides; the trait only promises that no
matching node goes missing. The two retriever comments repeated the false claim
and are corrected the same way.

**4. The lowercase-needle precondition was an unenforced footgun (low).**
`PgGraphAdapter` lowercased the needle for its SQL and returned rows, which the
Rust post-filter then discarded every one of; Ladybug returned an empty vec
outright. An uppercase needle produced a *silently empty result* on a new public
method rather than a failure. `node_label_contains` now normalises its own
needle, so all four implementations agree and the precondition is gone.

### New guards, both verified to fail on revert

```
test_get_candidate_nodes_by_label   (live PostgreSQL 17, blob-wide SQL restored)
  a node that merely mentions the needle in its text must not be a candidate —
  the filter is on labels, not on content: ["cl_prose", "cl_r"]

node_label_contains_normalises_the_needles_own_case   (traits.rs:1200)
  needle "Interaction" must match regardless of case
```

Re-verified: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`
clean; `cognee-graph` 156 passed; `cognee-search --lib` 323 passed;
`pg_graph_integration` 34/34 against a live PostgreSQL 17.

PR #165, opened 2026-08-31
fix(llm): bound the total time one structured-output call can take

GraDKh:
Closes the "hung request beyond timeout" finding: a single extraction could run for **45 minutes** while every individual HTTP request finished well inside its timeout, because nothing bounded the aggregate.

## The problem

The reqwest timeout was hardcoded at 600s per *request*, and structured extraction composes those. `structured_output_impl` cascades through three request shapes (tool calls → legacy functions → JSON mode), each `structured_output_retries` deep, each attempt honouring the `min_retry_elapsed` time floor from the retry work in #144. Multiplied out, the designed worst case exceeds an hour.

Two related gaps in the same area:

- **No `connect_timeout` anywhere.** `reqwest` sets none by default, so a black-holed connect — a stopped local Ollama, a wedged proxy — burned the full 600s without sending a byte.
- **No `tokio::time::timeout` anywhere** in the `llm`, `cognify` or `lib` crates.

## The change

Three knobs, each at a different scope, because the existing one composed into no aggregate:

| Knob | Bounds | Status |
|---|---|---|
| `LLM_CONNECT_TIMEOUT_SECONDS` | TCP handshake | new, default `10` |
| `LLM_REQUEST_TIMEOUT_SECONDS` | one HTTP request | was hardcoded `600` |
| `LLM_REQUEST_DEADLINE_SECONDS` | one *logical* call | new, default `1800` |

The deadline is checked at the head of **every cascade attempt**, not only between modes, so a long retry ladder inside a single mode cannot overrun it either. It gates *starting* new work rather than cancelling in flight, so one already-dispatched request can overrun by at most the request timeout — the honest ceiling is `deadline + request_timeout`, and both are configurable. `0` disables the deadline.

**On the 1800s default.** It sits above the legitimate retry envelope and below the pathological one: three modes each honouring the default 240s floor account for ~24 minutes of *deliberate* waiting that a call surviving a rate-limit window genuinely needs. The factory warns when the configured deadline falls below the ladder it has to contain (`3 × LLM_MAX_RETRIES × LLM_MIN_RETRY_SECONDS`), since a tighter budget would quietly undo #144's storm survival.

## Two adjacent fixes

- **The transcriber factory never installed the pacer.** A transcriber-only process — audio ingestion with no cognify — ran with no rate limiting and no in-flight ceiling at all. Transcription is an LLM HTTP call like any other; it now installs both gates.
- **The http-server carries the three settings itself** rather than inheriting defaults, for the same reason `llm_max_parallel_requests` does: its cognify routers build `CognifyConfig::default()`, so a knob the server does not lower is a knob its operators cannot reach. This is exactly the failure mode #155 just fixed for the parallelism knob.

## Second commit: self-review fixes

`1f0100d` fixes two real defects found reviewing the first commit:

1. **`0` meant opposite things across the three new knobs.** The deadline documents `0` as "disabled", but the two timeouts passed `Duration::ZERO` straight to reqwest. Verified against the previous code rather than assumed: **every request aborts after 0.0s** with `error sending request`. An operator generalizing "0 disables it" from one knob to its neighbours would have stopped all LLM traffic instead of lifting a bound. Both now mean "no limit". Pinned by a test confirmed to fail against the first commit.
2. **The guard-rail warning under-reported the ladder by the retry multiplier** — it counted the three cascade modes but not the `LLM_MAX_RETRIES` attempts inside each. Real floor sum is 1440s at defaults, not 720s. With `LLM_MIN_RETRY_SECONDS=600` the true ladder is 3600s against an 1800s deadline and the old check evaluated `1800 < 1800` and stayed **silent** — precisely the misconfiguration the warning exists to catch.

## Verification

`cargo fmt --check`, `cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings` — all exit 0.

New `openai_request_deadline` suite (4 cases), including an **exact** assertion that a spent budget issues 1 request rather than the cascade's 6. `openai_truncation_retry` (6), `openai_structured_output` (7), `retry_parity` (12), `cognee-components` and the `cognee` config unit tests (77) all pass.

Two pre-existing failures, both confirmed unrelated by reverting the adapter and reproducing them:
- `integration_openai::test_simple_text_generation` — the live endpoint returns 10 completion tokens with **empty content**. That is the reasoning-token-only symptom described below, observable today.
- `http-server::post_cognify_blocking` — HTTP 401 `invalid_api_key` from the embedding endpoint. Credentials, not code.

## Scope limits, stated rather than left to be discovered

- The aggregate bound covers the **OpenAI adapter's cascade**, where the reported symptom lives. Anthropic runs a single mode and is bounded by its own retry budget.
- `LLM_REQUEST_TIMEOUT_SECONDS` / `LLM_CONNECT_TIMEOUT_SECONDS` bind the OpenAI-compatible adapter (including Azure). The Anthropic, Responses and Bedrock clients still use a fixed 600s/10s pair — documented in `docs/configuration.md`.
- **Reasoning-token-only responses are still not detected.** `completion_tokens_details` is modelled nowhere, so a response with large `reasoning_tokens` and empty content reads as blank output and is retried through the cascade.
- `LlmConfig::timeout_seconds` remains dead code, read by nothing but its own setter. Wiring or removing it is a separate decision.
- The ladder fix is asserted **by inspection, not by a test**: capturing a `tracing` warning needs a subscriber harness this crate lacks, and I judged that not worth building for a two-line formula. Flagging so a reviewer can disagree.

https://claude.ai/code/session_01HDhq4uC3jVtKCDGpGGhUEt

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -83,18 +83,21 @@ LLM_API_KEY="your_api_key"
 #LLM_MAX_PARALLEL_REQUESTS=1000
 # Timeout for a single LLM HTTP request. Was hardcoded at 600 and is now
 # configurable, so a deployment behind a slow gateway can fail faster.
+# 0 = no limit. Applies to the OpenAI-compatible adapter (incl. Azure); the
+# Anthropic, Responses and Bedrock clients use a fixed 600s / 10s.
 #LLM_REQUEST_TIMEOUT_SECONDS=600
 # TCP connect timeout. reqwest applies none by default, so before this a
 # black-holed connect (a stopped local Ollama, a wedged proxy) consumed the
-# whole LLM_REQUEST_TIMEOUT_SECONDS without sending a byte.
+# whole LLM_REQUEST_TIMEOUT_SECONDS without sending a byte. 0 = no limit.
 #LLM_CONNECT_TIMEOUT_SECONDS=10
 # Wall-clock ceiling on ONE structured-extraction call, across the whole
 # tools -> legacy-functions -> JSON-mode cascade and every retry inside it.
 # LLM_REQUEST_TIMEOUT_SECONDS bounds a single request and composes into no
 # aggregate: three modes, each LLM_MAX_RETRIES deep, each honouring the
 # LLM_MIN_RETRY_SECONDS floor, multiply out past an hour. Keep it above
-# 3 x LLM_MIN_RETRY_SECONDS or legitimate rate-limit-window waits get cut
-# (a warning is logged if it is not). Set to 0 to disable.
+# 3 x LLM_MAX_RETRIES x LLM_MIN_RETRY_SECONDS (1440s at the defaults) or
+# legitimate rate-limit-window waits get cut; a warning is logged if it does
+# not fit. Set to 0 to disable the deadline entirely.
 #LLM_REQUEST_DEADLINE_SECONDS=1800
 
 ################################################################################
--- a/.env.example
+++ b/.env.example
@@ -81,6 +81,21 @@ LLM_API_KEY="your_api_key"
 # number. Default is 128 on Android/iOS, where 1000 sockets would exhaust the
 # typical descriptor budget.
 #LLM_MAX_PARALLEL_REQUESTS=1000
+# Timeout for a single LLM HTTP request. Was hardcoded at 600 and is now
+# configurable, so a deployment behind a slow gateway can fail faster.
+#LLM_REQUEST_TIMEOUT_SECONDS=600
+# TCP connect timeout. reqwest applies none by default, so before this a
+# black-holed connect (a stopped local Ollama, a wedged proxy) consumed the
+# whole LLM_REQUEST_TIMEOUT_SECONDS without sending a byte.
+#LLM_CONNECT_TIMEOUT_SECONDS=10
+# Wall-clock ceiling on ONE structured-extraction call, across the whole
+# tools -> legacy-functions -> JSON-mode cascade and every retry inside it.
+# LLM_REQUEST_TIMEOUT_SECONDS bounds a single request and composes into no
+# aggregate: three modes, each LLM_MAX_RETRIES deep, each honouring the
+# LLM_MIN_RETRY_SECONDS floor, multiply out past an hour. Keep it above
+# 3 x LLM_MIN_RETRY_SECONDS or legitimate rate-limit-window waits get cut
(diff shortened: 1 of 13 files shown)


GraDKh:
Addressed all three review findings in 067e008 — all were correct.

**1. The deadline didn't reach the transport retry ladder.** Correct, and worse than the comment implied. The field doc claimed the budget spanned "every transport retry inside" the call, but the guards sat only at cascade attempt heads; `send_chat_request`'s loop never consulted it. Measured with the guards absent (`network_retries = 10`, `min_retry_elapsed = 30s`): **the new test does not finish inside 420 seconds. With the fix it completes in 0.21s.** So the documented `deadline + request_timeout` ceiling wasn't optimistic, it was wrong by orders of magnitude under persistent transient failure — precisely the case this PR set out to close.

`call_api` / `send_chat_request` now take an absolute `Instant`, threaded from all three cascade call sites. The ladder won't start a further attempt once the budget is spent, and clamps its backoff to what remains — a 128s backoff against 5s of budget would blow the ceiling on its own. New test `spent_budget_stops_the_transport_retry_ladder_too` pins it.

**2. `DEFAULT_REQUEST_DEADLINE`'s rationale undercounted the envelope** (the suppressed comment). Correct — same multiplier bug I'd already fixed in the guard-rail one commit earlier, left behind in the constant's doc. `3 × 240s` = 12 min was wrong; the real envelope includes the retry count, `3 × 2 × 240s` = 24 min. Code, docs and PR description now agree.

**3. The http-server's `llm_max_retries: 3`.** Correct and a real bug: `Settings` and `docs/configuration.md` both say 2, so the server's *own default config* tripped the new guard-rail (3×3×240 = 2160s vs the 1800s deadline) and warned at every startup. Aligned to 2.

⚠️ **One thing to confirm:** finding 3 is a behaviour change — the HTTP server now retries twice rather than three times by default. Nothing in the file justified the 3, and the alternative (raising only the server's deadline so the mismatch fits) would have preserved an undocumented divergence to silence a warning that was telling the truth. Revertable in isolation if that default was deliberate.

Verified: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` exit 0; `openai_request_deadline` (5), `openai_truncation_retry` (6), `openai_structured_output` (7), `retry_parity` (12), http-server lib (270) all pass.

PR #166, opened 2026-08-31
feat(java,bindings): make the JNI binding work on Android, and expose COGX export

GraDKh:
Makes the Java binding usable on Android, and gives the bindings a COGX export op.

Targets `main`. #161 (the COGX migration crate this builds on) is already merged, so this is the same work rebased onto current `main` — supersedes #163, which was merged into #161's branch after that branch had already been squash-merged, leaving these commits off `main`.

All of this came out of embedding the SDK in an Android app and hitting each wall in turn; everything below was verified on a physical device (Android 16, arm64-v8a) unless noted.

## Bug fixes

**The binding could not load on Android at all.** `JNI_OnLoad` returned `JNI_VERSION_1_8`; ART accepts nothing above 1.6 and fails the load outright with `Bad JNI version returned from JNI_OnLoad: 65544`. The declared version is now `cfg`'d per target — every entry point here is already inside the 1.6 surface.

**Nothing with a structured result could be deserialized on Android.** D8 desugars `record` declarations for minSdk < 34, stripping the metadata Jackson's record support needs, so every op failed with `Cannot construct instance of ai.cognee.CogneeDataset (no Creators, like default constructor, exist)`. It hid at first because an *empty* dataset list never constructs an element — the failure only appeared once a dataset existed.

The result records now carry `@JsonProperty` on every component. The annotation propagates to the canonical constructor's parameters and survives desugaring, so the constructor is a properties-based creator by name alone: no module registration, no javac `-parameters`, no runtime platform check. `Json` keeps a plain `ObjectMapper` on every platform, so JVM marshalling is unchanged and cannot be perturbed by whatever modules a host has on its classpath. This also fixes *serialization* of desugared records.

**`pdfium-render` 0.9.2 does not compile for `aarch64-linux-android`** — it hardcodes `*const i8` where `c_char` is unsigned there. Floor raised to 0.9.3 in the workspace manifest rather than a lockfile pin, since consumers that gitignore `Cargo.lock` resolve versions themselves.

## New capability

**`set_handle_factory` + `rlib`.** The Java binding was the only one without a cloud-flavored counterpart, so it always built handles with the OSS registry, ruling out the closed adapters on the platform that needs them most. Rather than fork ~2k lines of glue the way the other cloud cdylibs do, the crate now also builds as an `rlib` and exposes a handle-constructor seam. Unset, behaviour is identical to before.

**`set_extra_migrations`.** sea-orm rejects a database whose `seaql_migrations` records a migration the running migrator does not know (`Migration file of version '…' is missing`), so a downstream crate that adds tables cannot just run its own migrator afterwards — the next schema init fails and the database is stuck until someone deletes it. This lets one migrator own the whole applied set. OSS-only builds never call it.

**COGX export through the bindings.** The migration crate from #161 has a Rust API and a CLI command, so only a process with a shell could produce an archive — which rules out the case the format exists for on mobile. `ops::migration::export_cogx` runs `export_graph` then `pack_archive` (the two halves must agree on the `.cogx.tar.gz` suffix and on `manifest.json` at the tarball root, which is all Python's `find_archive_root` accepts), plus `Cognee.exportCogx` / `ExportOptions` / `ExportResult`.

Verified end to end: 105 nodes / 310 edges from a phone into a 141 KB tarball, imported by a cognee Cloud tenant via `POST /api/v1/remember` with `content_type=cogx-archive`.

**PDF and EML loaders exposed** (`pdf-pdfium`, `pdf-pure-rust`, `unstructured-eml`) — only `html-loader` was forwarded before. PDFium is the one to use: `pdf-extract 0.10.0` returns `Ok` with **empty text** for real PDFs (measured on a 3 MB Alice in Wonderland edition — 105 pages, 0 characters, where poppler read 179,173 from the same file), and an empty document then ingests as a *success*, surfacing three steps later as a search failing on a missing vector collection. PDFium reads it correctly: 181,398 bytes, 77,782 tokens, 10 chunks, 64 entities.

Consumers supply `libpdfium` themselves — note `pdfium-auto` has no Android entry in its target table at all, so an Android consumer must ship the `.so` and set `PDFIUM_LIB_PATH` (plus `PDFIUM_NO_AUTO_DOWNLOAD=1`, or a phone tries to fetch a native library at runtime).

## Review hardening

The last two commits address findings from three review passes over this branch (a general review, an adversarial validation of it, and Copilot on #163):

- **`export_cogx` could ship files it never wrote.** `pack_archive` tars every top-level file of its directory, so a caller reading "the archive directory to write" as "somewhere I already keep things" would upload those things. The destination must now be absent or empty. It fails **closed** rather than filtering by name — only 3 of the archive's 9 filenames are public, so a name filter would drift from the writer's list, and the failure mode of drift is leaking a file.
- **Destinations with no final component produced a useless archive.** `.`, `..`, `/` all fell back to `cognee.cogx.tar.gz` in the CWD; for `.` that is inside the directory being packed, and since `pack_archive` creates the file before enumerating, the half-written tarball was appended to itself. Downstream, Python's `tarfile` silently yields only that junk member — no manifest. Now rejected before the graph is read.
- **A failed pack left a truncated tarball** at exactly the path a successful call returns, so anything keying off the file existing would ship a stream the importer rejects opaquely. Removed on failure.

## Verification

`cargo check --all-targets` on both workspaces (the `cognee-java-jni` crate is its own workspace), `cargo fmt --check`, `cargo clippy -- -D warnings`, and 9 unit tests in the new module (13 in the crate). The new tests cover the destination guard, the rejected path shapes, and the eight result keys **plus their values** — the bindings deserialize that object by name with `ignoreUnknown = true` and `int` counts, so a rename or a mis-wired field silently yields `0` rather than failing.

Two things a reviewer should know rather than discover:

- **The Java sources are uncompiled here** — no Maven in the environment where this was built. They were reviewed by reading.
- **The `@JsonProperty` path has not been exercised at runtime**, because the test device was disconnected before that run. The evidence is `d8 --min-api 26` output showing the annotations landing on the constructor parameters after desugaring, plus a desugared-record simulation deserializing with a plain mapper and a negative control reproducing the original `no Creators` error. It wants one on-device check against a populated store before anyone relies on it.
- The 8 failing `handle_close.rs` tests are pre-existing (they come from #161 and need backends that `--no-default-features` disables) — confirmed by stashing this branch's changes and reproducing them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,9 @@ pbkdf2 = { version = "0.12", default-features = false }
 pdf-extract = "0.10"
 quick-xml = "0.39"
 pdfium-auto = "0.3"
-pdfium-render = { version = "0.9", features = ["thread_safe"] }
+# 0.9.3 is the floor for Android: 0.9.2 hardcodes `*const i8` where c_char is
+# unsigned on aarch64-linux-android, so it fails to compile there (E0308).
+pdfium-render = { version = "0.9.3", features = ["thread_safe"] }
 predicates = "3.1"
 rand = "0.8"
 rayon = "1.10"
--- a/crates/bindings-common/src/ops/migration.rs
+++ b/crates/bindings-common/src/ops/migration.rs
@@ -116,9 +116,20 @@ fn check_destination(destination: &Path) -> Result<(), SdkError> {
             "destination must be a non-empty path".to_string(),
         ));
     }
-    let Ok(mut entries) = std::fs::read_dir(destination) else {
-        // Absent (or unreadable — the export itself will report that better).
-        return Ok(());
+    let mut entries = match std::fs::read_dir(destination) {
+        Ok(entries) => entries,
+        // Genuinely absent: the export creates it.
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => return Ok(()),
+        // Anything else means emptiness could not be verified — a file where a
+        // directory was expected, or a directory that cannot be read. A guard
+        // whose purpose is to keep unrelated files out of an uploaded tarball
+        // must not treat "cannot tell" as "nothing there".
+        Err(error) => {
+            return Err(SdkError::Validation(format!(
+                "destination '{}' cannot be checked for emptiness: {error}",
+                destination.display()
+            )));
+        }
     };
     if entries.next().is_some() {
         return Err(SdkError::Validation(format!(
@@ -211,6 +222,24 @@ mod tests {
                 "{path} should be rejected as invalid, got {error:?}"
             );
         }
+    }
+
+    #[test]
+    fn a_destination_whose_emptiness_cannot_be_verified_is_rejected() {
+        // A path that exists but is a *file* cannot be read as a directory. The
+        // guard used to treat any read_dir error as "absent" and wave it
+        // through, which is the opposite of what a guard against packing
+        // unrelated files should do when it cannot tell what is there.
+        let file = std::env::temp_dir().join(format!("cogx-not-a-dir-{}", std::process::id()));
+        std::fs::write(&file, b"x").expect("write");
+
+        let error = check_destination(&file).expect_err("a file is not a usable destination");
+        assert!(
+            matches!(error, SdkError::Validation(_)),
+            "expected a validation error, got {error:?}"
+        );
+
+        let _ = std::fs::remove_file(&file);
     }
 
     #[test]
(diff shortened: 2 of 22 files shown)


GraDKh:
Supersedes #163. That PR was merged into `feat/cogx-export` (#161's branch), but #161 had already been squash-merged to `main`, so those commits never reached `main` — and `ci.yml` is filtered to PRs targeting `main`/`master`, so the full suite never ran on them either. Same 8 commits, cherry-picked onto current `main` with no conflicts, so this PR gets real CI coverage.

PR #178, opened 2026-09-03
fix(llm): stop probing tool calls on an endpoint that has no parser

GraDKh:
Closes finding 9 of the corpus-run triage — the last open item in the WS D chain after #144 (pacing), the truncation fix and #174 (aggregate deadline) landed.

## Problem

The structured-output cascade (tool calls → legacy `functions` → JSON mode) is stateless per call: nothing remembers that an endpoint has never once produced a `tool_calls` entry. An OpenAI-compatible server started without a tool-call parser — vLLM without `--enable-auto-tool-choice --tool-call-parser` — answers HTTP 200 with the text in `message.content` and never populates `tool_calls`. Mode 1 can therefore never succeed, and every structured call re-pays the whole ladder.

Measured on one such deployment: **zero tool calls in 11,890 responses**, 4.14 API calls per extraction against 1.09 on an endpoint with a parser, with **~71% of output tokens and cost producing nothing**.

## Approach

`ToolSupportProbe` gives the adapter a per-endpoint memory. After `MISS_THRESHOLD` (3) consecutive structured calls complete without a native tool call, mode 1 is skipped and the request goes straight to the fallbacks.

Three rather than one because a single miss is far more likely to be a bad response than a missing parser, and being wrong in the cautious direction costs only two extra cascades.

The memory is deliberately **not permanent**: a tripped probe still lets one call in every `RE_PROBE_INTERVAL` (64) attempt tool calling, so an endpoint that gains a parser behind a stable URL recovers without a restart, and a transient run of bad responses cannot disable mode 1 for the life of the process. Residual waste is ~1.6% of calls.

Detection keys on **capability, not failure**: only `tool_calls` and a legacy `function_call` count, never JSON echoed in `content`, and a transport error records nothing. The signal is split out of the existing `arguments` precedence chain, which is otherwise unchanged.

Mode 1 is gated by a zero-length attempt range rather than a wrapping conditional, so the loop body, the `ValidationMiss` check after it and the budget variables all keep their existing shape — with no attempts the outcome stays `NoUsableOutput` and control falls through to legacy mode. This keeps the diff to the cascade small and reviewable.

## Notes for review

- **No new setting.** This is a property of the endpoint, observed rather than declared. An explicit `auto | tools | functions | json` override is a reasonable follow-up but needs plumbing across six crates; the automatic path fixes the problem without the operator having to know.
- **Not a parity divergence.** Python has no cascade to protect — it pins one instructor mode per provider from a static table (`instructor_modes.py`) and stays there, overridable by `llm_instructor_mode`. The cascade is Rust-only, so bounding it is bounding our own mechanism.
- **Concurrency.** The counters are `Relaxed` — they are a heuristic, and a race costs at most one extra probe — and live behind an `Arc` so the state follows the endpoint rather than resetting on every `clone()` of the adapter (which `#[derive(Clone)]` would otherwise do).

## Tests

Three unit tests pin the state machine: the threshold, the reset on a native tool call, and the re-probe interval (including that the re-probe is one call, not a permanent reset).

Two httpmock tests cover the ends:
- a parser-less endpoint stops being sent tool calls after three calls, while still being answered via JSON mode;
- a working endpoint is sent exactly one tools request per call, guarding against the probe tripping on a healthy provider.

The first **fails with the probe defeated** — 4 tools requests against 3 — verified by temporarily forcing `try_tools = true`.

`cargo fmt`, `cargo clippy -p cognee-llm --all-targets -- -D warnings` and `cargo check --workspace --all-targets` all clean. Full `cognee-llm` suite passes; the one failure (`integration_openai::test_simple_text_generation`) is pre-existing and environmental — it fails identically with these changes stashed, because the configured live endpoint returns empty content.

https://claude.ai/code/session_016MjPQ455oLeo7coh7Lbxmh

### DIFF
--- a/crates/llm/src/adapters/openai.rs
+++ b/crates/llm/src/adapters/openai.rs
@@ -49,43 +49,49 @@ use crate::types::{GenerationOptions, GenerationResponse, Message, MessageRole, 
 ///     None,
 /// ).await?;
 /// ```
-/// Consecutive-miss memory for one endpoint's tool-calling support.
+/// Per-mode memory of whether a structured-output cascade mode is worth sending
+/// to one endpoint.
 ///
-/// The structured-output cascade below (tool calling -> legacy
-/// `functions`/`function_call` -> JSON mode) is stateless per call: nothing
-/// remembers that an endpoint has never once produced a `tool_calls` entry.
-/// Against an OpenAI-compatible server started without a tool-call parser --
-/// vLLM without `--enable-auto-tool-choice --tool-call-parser` is the case this
-/// exists for -- mode 1 can never succeed, so *every* structured call pays the
-/// full tools -> corrective -> legacy -> JSON ladder before reaching an answer
-/// the third mode could have given immediately. Measured on one such
+/// The cascade (tool calling -> legacy `functions`/`function_call` -> JSON mode)
+/// exists because OpenAI-compatible servers differ in what they accept, but it
+/// was stateless per call: nothing remembered that an endpoint had never once
+/// answered a given mode. A server started without a tool-call parser -- vLLM
+/// without `--enable-auto-tool-choice --tool-call-parser` -- answers 200 with
+/// the text in `message.content` and populates neither `tool_calls` nor
+/// `function_call`, so the first two modes can never produce anything and every
+/// structured call re-paid their whole retry ladder. Measured on one such
 /// deployment: zero tool calls in 11,890 responses, 4.14 API calls per
 /// extraction against 1.09 on an endpoint with a parser, with ~71% of output
 /// tokens and cost producing nothing.
 ///
-/// Python has no cascade to protect: it pins one instructor mode per provider
-/// from a static table (`instructor_modes.py`) and stays there, overridable by
-/// `llm_instructor_mode`. The cascade is a Rust-only mechanism, so bounding it
-/// is not a parity divergence.
+/// The counter tracks "is this mode worth sending", not "does the server
+/// implement it". The distinction matters: an endpoint with no parser that
+/// echoes usable JSON in `content` is answered by mode 1 on its first attempt,
+/// so mode 1 is the *cheapest* path there and must not be skipped. Only a mode
+/// that ran out of attempts having produced nothing usable counts against
+/// itself.
+///
+/// Python needs no equivalent: it pins one instructor mode per provider from a
+/// static table (`instructor_modes.py`) and stays there, overridable by
+/// `llm_instructor_mode`. The cascade is Rust-only, so bounding it is bounding
+/// our own mechanism rather than diverging from Python.
 ///
 /// The memory is deliberately **not** permanent. Once tripped it still lets one
-/// call in every [`Self::RE_PROBE_INTERVAL`] attempt tool calling, so an
-/// endpoint that gains a parser (a redeploy behind a stable URL) recovers on its
-/// own, and a transient run of bad responses cannot disable mode 1 for the life
-/// of the process. Held behind an `Arc` so the state follows the endpoint rather
-/// than resetting on every `clone()` of the adapter.
+/// call in every [`Self::RE_PROBE_INTERVAL`] try the mode, so an endpoint that
+/// gains a parser (a redeploy behind a stable URL) recovers on its own and a
+/// run of bad responses cannot disable a mode for the life of the process.
 #[derive(Debug)]
-struct ToolSupportProbe {
-    /// Structured calls in a row whose tool-calling mode completed without a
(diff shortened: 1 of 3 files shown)


GraDKh:
Pushed `72eddfb` — fixes for five review findings. The first was a genuine correctness bug that contradicted this PR's own description, so the summary above is now partly wrong; the corrected behaviour is below.

**A miss was recorded on paths carrying no capability information.** `native_tool_call` also stays false when the tools request *errored* (the `Err(e) => break` arm — HTTP 4xx/5xx, connection reset, schema rejection) or when a truncation consumed the last attempt. So three concurrent chunks in a cognify fan-out hitting a brief gateway blip were enough to disable tool calling on a healthy endpoint for the next 64 calls — and the fallbacks are materially worse, since JSON mode sends only a `schema_to_example` template rather than the real schema. Misses are now gated on a response that actually *arrived* carrying no native call. Truncation is excluded too: it says only that the output budget was too small, which `LLM_ARGS` and the truncation retry already handle.

**`consecutive_misses` was not consecutive.** The success path returns from inside the loop before any probe update, and only a native tool call reset the count, so misses accumulated across arbitrarily many intervening successful calls. On an endpoint that echoes usable JSON in `content` — where mode 1 succeeds on its first attempt and is the *cheapest* path — that eventually skipped the one mode that was working. Any call the mode answers now clears the count, whichever field the payload came from.

**The `ValidationMiss` short-circuit was not gated on the new signal.** Its comment justifies skipping the fallbacks with "the server clearly speaks tool calling" — a claim this diff could finally test but didn't. On a parser-less endpoint echoing an incomplete object it returned `DeserializationError` *and* recorded a miss, so the same request failed three times and then succeeded once the probe tripped. Byte-identical input, two outcomes. Now gated on `native_tool_call`.

**Only one of the two useless modes was removed.** Legacy `functions` needs a server-side parser just as tool calling does. `ToolSupportProbe` is now `ModeProbe` plus a `CascadeProbe` holding one per mode, tracked **independently** — a shared counter would skip legacy mode on a server that supports only legacy, which is the case the cascade exists for. JSON mode has no probe, being the terminal fallback.

**The skip log reported the constant, not what happened**, at `debug`, for all 64 skipped calls. `record_useless` now returns the new count, and the tool-calling transition logs once at `warn` naming the vLLM flags, so a downgrade is visible rather than silent.

### On the tests

Every fix now has a test verified to fail with that fix individually reverted. Worth flagging that this caught a weak test of my own: `a_content_echoing_endpoint_keeps_using_tool_calling_mode` passed with the reset removed, because the success path returns before the miss is recorded — so it never guarded the reset. Replaced with `a_call_the_mode_answers_resets_the_consecutive_miss_count`, which walks one adapter through an interleaved unusable/answered/unusable sequence keyed on the input text: two misses, a reset, two more. With the reset that is a run of two, below the threshold; without it, four.

Now 5 unit + 5 httpmock tests. `cargo fmt`, `clippy -D warnings`, `check --workspace --all-targets` clean; 25 `cognee-llm` suites green, with the same pre-existing environmental failure in `integration_openai::test_simple_text_generation`.

Not addressed: an explicit `auto | tools | functions | json` override, still a follow-up.

GraDKh:
Addressed Copilot's review in `0013c56`. All four findings were valid; none were behavioural bugs.

**The semantics mismatch was the substantive one, and Copilot read the code correctly.** `record_useful()` fires the moment a mode's native field is *observed* — even if that payload then fails to parse or validate — while my comments and the docs claimed the count clears only when the mode "produced something usable". Those are different claims, and the code's is the deliberate one: the field's presence proves a server-side parser exists, which is the only question the probe asks. A badly-formed tool call is a model problem, already handled by the corrective-retry ladder, not a capability one. Both rules are now stated explicitly in the `ModeProbe` doc, on the `consecutive_misses` field, on `record_useful` itself, and in `docs/configuration.md`, which now names the native field as the actual criterion.

**The `>=` assertions are a real strengthening, so thanks for pushing on it.** With `with_structured_output_retries(1)` and `with_network_retries(0)` the counts are deterministic, and a loose bound would let a regression in request *volume* — the entire point of this change — pass unnoticed. Now `assert_eq!` on exactly 3 tools, 3 legacy and 4 JSON-mode requests across the four calls, which doubles as documentation of the per-call budget being protected. Same tightening applied to the transport-error test (3, then exactly one more).

**The "then starts working" comment was wrong** — left over from an earlier draft of that test. The mock 400s on every call; the point is that an error is not evidence about the parser, not that the endpoint recovers.

All four guard tests still fail with their own fix individually reverted, re-checked after tightening the assertions. `clippy -D warnings` and the 124 lib tests plus 12 structured-output tests are clean.

PR #182, opened 2026-09-07
fix(cli): wire --llm-max-retries so the flag reaches the LLM client (SDK-511)

GraDKh:
Closes SDK-511.

## Problem

`--llm-max-retries` is declared on `cognify`, `add-and-cognify` and `search`, and read nowhere. Passing `--llm-max-retries 10` parsed cleanly and changed nothing — behaviour stayed at the config default of 2, while `LLM_MAX_RETRIES` worked.

It is the **only** unread flag in the CLI, not a systemic plumbing problem: the sibling `--llm-max-parallel-requests` is correctly consumed at `commands/cognify.rs:20`.

## What changed

**`Commands::apply_overrides(&mut Settings)`**, called from `run()` before `ConfigManager::new`.

The ordering is load-bearing. `ComponentManager` builds the LLM adapter lazily from the settings snapshot it is handed, and `build_openai_compatible_adapter` bakes the retry counts in at construction — so applying the override inside each command would race an adapter that may already exist. Applying it once at the top makes the constraint structural instead of a rule every new command has to remember.

The fold lives in the library rather than in `main` **so it is reachable from tests**. That matters for this bug specifically: the flag always parsed correctly and then reached nothing, so a parse-only assertion would have passed against it.

**Validator relaxed** from `range(1..)` to `range(0..)`. `LLM_MAX_RETRIES=0` parses and is stored today, so rejecting `--llm-max-retries 0` was an asymmetry between two surfaces for one setting.

> Worth being precise, because the ticket is not: `0` is **not** a fail-fast switch. `build_openai_compatible_adapter` floors it at 1, so `0` and `1` behave identically. The documented escape hatch is `LLM_MIN_RETRY_SECONDS=0`, per the doc comment on `Settings::llm_min_retry_seconds`. The flag help now says this.

## Tests

Six unit tests in `crates/cli/src/cli.rs`. The load-bearing one is `apply_overrides_writes_the_flag_into_settings` — **verified by mutation**: stubbing the fold to a no-op makes it fail, restoring it makes it pass. The other five cover reachability from all three subcommands, absent-flag no-clobber, a subcommand without the flag, and `0`.

Not covered: that the value reaches a live adapter. That needs a built LLM client, and the `Settings` -> `LlmInputs` -> factory lowering is exercised elsewhere.

## On the stated blocker

SDK-511 is marked blocked by SDK-508 (tool-call-parser detection). **That shipped in #178**, which bounds the cascade — so the ticket's concern, that a high retry count multiplies calls against a parser-less endpoint, is now mitigated upstream. SDK-508 is still `Todo` in Linear despite the code being merged; flagging that separately.

## Verification

`cargo fmt --check`, `cargo check -p cognee-cli --all-targets`, `cargo clippy -p cognee-cli --all-targets -- -D warnings`, and the full `cargo test -p cognee-cli` suite all exit 0. Confirmed against the real binary that `--llm-max-retries 0` and `10` are both accepted.

https://claude.ai/code/session_01Pcm8xaVZpTq8eVthcu5tVf

### DIFF
--- a/crates/cli/src/cli.rs
+++ b/crates/cli/src/cli.rs
@@ -44,22 +44,38 @@ impl Commands {
     /// silently did nothing while `LLM_MAX_RETRIES` worked — the sibling
     /// `--llm-max-parallel-requests` was wired, this one was missed (SDK-511).
     ///
-    /// Returned rather than applied here so the caller can fold it into
-    /// `Settings` *before* `ConfigManager` is built: `ComponentManager`
-    /// constructs the LLM adapter lazily from the settings snapshot it was
-    /// handed, so a mutation made after that point cannot reach an adapter that
-    /// has already been built.
+    /// Returned rather than applied, so each entry point can fold it in at the
+    /// point that suits it: `main::run` writes it into `Settings` before
+    /// `ConfigManager` exists, while `run_sequence` — whose steps never reach
+    /// `main::run` — goes through `ConfigManager::set_llm_max_retries` per step.
+    ///
+    /// Folding it in up front is preferred where possible. Not because a later
+    /// change could not take effect (it can: the setter bumps the config version
+    /// and `ComponentManager` rebuilds affected components on next access), but
+    /// because that rebuild discards every warm component, which is wasteful
+    /// when the value was known before anything was built.
     pub fn llm_max_retries_override(&self) -> Option<u32> {
         match self {
             Commands::Cognify(args) => args.llm_max_retries,
             Commands::AddAndCognify(args) => args.llm_max_retries,
             Commands::Search(args) => args.llm_max_retries,
+            // `memify` / `remember` / `recall` / `improve` also drive LLM work
+            // but do not declare the flag. Adding it to them is a separate
+            // change — this one wires the flag that already exists. `RunSequence`
+            // is `None` by design: its steps carry their own flags and are
+            // folded in per step by `run_sequence::run`.
             _ => None,
         }
     }
 
-    /// Fold this invocation's flags over `settings`, so a flag outranks config
-    /// and env for the run it was passed on.
+    /// Fold this invocation's settings-valued flags over `settings`, so a flag
+    /// outranks config and env for the run it was passed on.
+    ///
+    /// Currently one field. `--llm-max-parallel-requests` is deliberately *not*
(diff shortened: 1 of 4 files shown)


GraDKh:
Adversarial review pass found one functional gap and three wrong claims — all fixed in 508a71b. Correcting the PR body above, since two of the wrong claims are in it.

### 1. `run-sequence` was a half-fix

Sequence steps are parsed by `run_sequence::run` and dispatched straight to the command, never passing through `main::run` — so a step's `--llm-max-retries` was **still dead** after the first commit. `demo/sequences/demo_pipeline.json:14` passes it, so the repo's own demo was hitting this.

Now folded per step via `ConfigManager::set_llm_max_retries`, guarded on the value actually changing (a sequence that overrides nothing keeps its warm components), and baselined before the loop so a step *without* the flag falls back to config rather than inheriting the previous step's value.

### 2. The `0` claim in the PR body above is wrong

I wrote that `0` and `1` behave identically because `build_openai_compatible_adapter` floors the value. **That is false for Bedrock**, which is in the CLI's default features: `with_network_retries` applies no floor and the loop is `for attempt in 0..=self.network_retries`, so `0` gives one attempt and `1` gives two. The floor is real only for the OpenAI-compatible, Azure and Anthropic adapters. And `LLM_MIN_RETRY_SECONDS` — which I offered as "the real escape hatch" — is never read by Bedrock at all.

Corrected in the flag help and in `docs/tools/cli.md`, which was separately stale: it still claimed a minimum of `1`, and described the setting as governing structured-output calls when it drives the transport retry loop too.

### 3. The placement rationale was wrong

I claimed a later change "cannot reach an adapter that has already been built." It can — `ConfigManager`'s setters bump a version and `ComponentManager` rebuilds affected components on next access. `bench.rs` already relies on this, and so does the run-sequence fix above.

The placement is still correct, for a different reason: the value is known before anything is built, so folding it in up front avoids discarding every warm component. Fixed in the code comments rather than left to mislead someone into avoiding a legitimate setter.

### Verification

`cargo fmt --check`, `cargo clippy -p cognee-cli --all-targets -- -D warnings`, the full `cargo test -p cognee-cli` suite, and `cargo check --workspace --all-targets` all exit 0. One test added for the sequence-step argv shape, copied from the demo file rather than invented — the first version I wrote used a flag name that does not exist and failed.

PR #185, opened 2026-09-07
feat(llm): let an operator pin the structured-output mode (SDK-508)

GraDKh:
Closes the remaining half of **SDK-508**.

The ticket asked for two things. #178 shipped the second (a sticky negative cache, so a mode an endpoint never answers stops being re-probed). This ships the first: an explicit operator override.

## Why both

They solve the same problem from opposite ends. The probe **learns** an endpoint's shape — at the cost of spending its miss threshold in every fresh process, and re-probing periodically forever. The pin is simply **told**, so it costs nothing and never re-probes.

An operator running vLLM without `--enable-auto-tool-choice --tool-call-parser` already knows the answer. The case that motivated the ticket saw **0 tool calls in 11,890 responses**, with 72.5% of calls and 71% of cost producing nothing. The probe bounds that; a pin avoids it entirely.

Keeping the probe matters too — it is the only thing that helps an operator who *doesn't* know what their gateway speaks.

## Parity

Python has no cascade at all: it picks one instructor mode per provider from a static table (`instructor_modes.py`) and exposes `llm_instructor_mode` to override it. So this knob closes a parity gap rather than opening one. The cascade remains the Rust default because it is a Rust-only mechanism and removing it would be the divergence.

## Behaviour

| Value | Sends |
|---|---|
| `auto` (default) | cascade: `tools` → `functions` → `json`, probe-bounded |
| `tools` | native tool-calling only |
| `functions` | legacy `functions`/`function_call` only |
| `json` | `response_format: {"type": "json_object"}` only |

- **`auto` is byte-for-byte the previous behaviour.** The change is opt-in: an adapter built without config — tests, embedders, downstream users of the published crate — keeps the cascade. Two regression tests guard this.
- **A pin excludes, it does not reorder.** The other two shapes never reach the wire.
- **A pin bypasses the miss probe** rather than composing with it. Composing would let a tripped probe disable the *only* permitted mode, at which point calls send no request at all and fail until the re-probe interval elapses. Attempting the one mode the operator asked for is strictly better. This was the one real design decision here.
- **Exhausting a pinned mode names the pin** and the env var, rather than reporting generic cascade exhaustion — otherwise a mode the endpoint cannot speak reads as a provider fault.
- **An unrecognised value falls back to `auto`**, matching `parse_reasoning_override`. A hard failure over a misspelled optional knob would be worse, and the cascade is the safe default in a way that no particular pin is.

## Wiring

Follows the sibling `LLM_REASONING` knob exactly: `Settings` field, env read, `ConfigManager` setter plus string-keyed get/set, lowered by a new `parse_structured_output_mode` into `LlmInputs`, applied by both OpenAI-compatible factories. The standalone HTTP server reads the env var on its own path too. Documented in `docs/configuration.md` and `.env.example`.

## Tests

7 new cases in `crates/llm/tests/openai_structured_output_mode.rs` — each pin in isolation, no fall-through on exhaustion, the probe bypass held across six consecutive calls, and the two `auto` regression guards. Plus unit coverage of token parsing (including alias and typo handling) and the per-mode predicates.

`cargo fmt`, `cargo clippy -D warnings` and the `cognee-llm` / `cognee-components` / `cognee` / `cognee-http-server` suites are clean. Two pre-existing failures in this environment are unrelated and confirmed against pristine `origin/main`: `integration_openai::test_simple_text_generation` (a live model spends all 10 `max_tokens` on reasoning and returns empty content) and `test_cognify_blocking` (HTTP 401 `invalid_api_key` on an embedding call).

## Incidental finding

Encoded in the fixtures: legacy `functions` mode requires a **native** `function_call` and will not accept JSON echoed in `content`, which the tool-calling path does accept. Worth knowing if anyone writes mocks against that mode.

https://claude.ai/code/session_01Cf5zk52eebyLhd24Hxu5u4

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,6 +66,10 @@ LLM_API_KEY="your_api_key"
 #LLM_REASONING="auto"            # auto | always | never — override OpenAI reasoning-model (o1/o3/o4/gpt-5)
                                  # parameter detection; use "never" for a remote gateway that serves a
                                  # reasoning-named model but only accepts legacy max_tokens+temperature
+#LLM_STRUCTURED_OUTPUT_MODE="auto"  # auto | tools | functions | json — pin the structured-output request
+                                 # shape instead of cascading tools -> functions -> json. Set "json" for a
+                                 # vLLM/gateway deployment with no --tool-call-parser: the cascade would
+                                 # otherwise pay all three modes before finding the one that works.
 #LLM_TEMPERATURE=0.0
 #LLM_MAX_RETRIES=2
 # Minimum seconds a transient LLM failure is retried for. Paired with
--- a/crates/components/src/builtins/database.rs
+++ b/crates/components/src/builtins/database.rs
@@ -159,6 +159,7 @@ mod tests {
                 llm_args: serde_json::Map::new(),
                 api_version: String::new(),
                 reasoning_override: None,
+                structured_output_mode: cognee_llm::StructuredOutputMode::Auto,
                 mock: false,
                 cassette: String::new(),
                 record_path: String::new(),
--- a/crates/components/src/builtins/llm.rs
+++ b/crates/components/src/builtins/llm.rs
@@ -133,6 +133,7 @@ impl LlmFactory for OpenAiCompatibleLlmFactory {
         .with_extra_args(ctx.llm.llm_args.clone())
         .with_default_max_tokens(Some(ctx.llm.max_completion_tokens))
         .with_reasoning_override(ctx.llm.reasoning_override)
+        .with_structured_output_mode(ctx.llm.structured_output_mode)
         .with_request_deadline(request_deadline(ctx));
         let (request_timeout, connect_timeout) = http_timeouts(ctx);
         let adapter = adapter.with_http_timeouts(request_timeout, connect_timeout);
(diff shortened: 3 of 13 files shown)


GraDKh:
Pushed `26c6acb`, addressing all four review findings.

**1 (medium) — a pin was swallowing the real failure. Fixed.** This was the important one: the pin was hiding the diagnostic it exists to provide, and the review is right that it made the mis-pin case *less* diagnosable than no pin at all. Both paths now surface the real error under a pin — the tool loop's `Err` arm returns instead of breaking toward modes that will not run, and the `ValidationMiss` short-circuit treats a pin as satisfying its `native_tool_call` gate, since that gate exists to keep a fall-through available and a pin has none.

**2 (low) — the misleading probe warning. Fixed.** `record_useless` is now skipped under a pin at both call sites. The review was also right that my comment overclaimed: it said the counters were untouched when only the read side was. The comment now says a pin neither consults them nor records misses, and notes that a success still resets them — which is inert either way.

**3 (low) — `json` oversold in the docs. Fixed.** Now distinguishes a server that *rejects* `tools`/`functions` from one that merely *ignores* them, and states the fidelity cost: JSON mode sends a prose example derived from the schema, never the schema itself, so pinning it on an ignoring endpoint trades one wasted request for permanently weaker schema pressure.

**4 (low) — silent fallback. Fixed.** An unrecognised value now warns. Not hard-failing was deliberate; staying silent was not, and the review's framing of that distinction is the right one.

Two regression tests added, and I verified both fail with the fixes reverted rather than assuming they were load-bearing.

Also confirming the review's read on scope: `crates/cli/src/config_store.rs::known_keys()` omitting this knob matches `llm_reasoning`, so I left it alone rather than making this PR the place that diverges.

https://claude.ai/code/session_01Cf5zk52eebyLhd24Hxu5u4

PR #187, opened 2026-09-07
fix(llm): classify truncation before payload inspection on bedrock + anthropic

GraDKh:
Ports the truncation-ordering fix merged for OpenAI in #157 (`c51a4aa9`) to the two
adapters that never received it.

## The bug

Found while investigating a real BYOD deployment: cognify against Bedrock
(Claude Sonnet 4.5) stalled ~4 minutes and returned HTTP 500, on 3 of 4 runs,
different chunks each time.

Both adapters tested truncation only on a **present** payload:

```rust
let truncated = response.is_truncated();
match response.structured_payload(&self.caps) {
    Some(_) if truncated => { /* raise / terminal */ }
    Some(payload)        => { /* validate */ }
    None                 => { /* corrective re-ask, SAME budget */ }
}
```

Sonnet 4.5 sets `supports_native_structured_output`, so `structured_payload`
parses the response **text** as JSON — and text cut off at `maxTokens` never
parses. The truncation arm was therefore **dead code on that branch**. Every
stall was filed as *"did not contain parseable JSON"*, re-asked at the identical
budget, re-truncated at the same point, and burned a full generation plus an 8s
backoff per attempt until `MaxRetriesExceeded`.

This was **not** a missing token cap. `effective_max_tokens` was correctly
applying 16384; the classification was wrong.

## Three deliberate decisions

**A caller-supplied `max_tokens` is terminal on Bedrock, not raised.** Before
this change the truncation arm was unreachable on the native branch, so a caller
budget was never rewritten there. Classifying correctly would have started
silently re-issuing a client's 1000-token request at the 16384 ceiling — a 16×
larger bill on a path where it never happened. #157 refuses this for exactly this
reason: the HTTP custom-prompt route forwards client budgets straight into
structured output. Anthropic's documented divergence — it *does* raise over a
caller budget — is left alone.

**The Bedrock retry log no longer renders `last_error`.** `DeserializationError`
embeds the raw Converse body verbatim, which on the cognify path is model output
derived from the user's ingested documents. At `warn` that would put user content
into operator logs. Only the variant is logged unconditionally; the message only
for errors this loop synthesises itself.

**`warn` is reserved for a truncation re-ask.** That is the expensive case worth
seeing at INFO. A validator- or parse-driven repair is routine — this loop runs
once per chunk — so promoting those would emit a `warn` per repaired chunk on an
ordinary ingest. Anthropic gains the same truncation-specific `warn`, so its
stall is no longer invisible either.

Untouched, both being deliberate Python parity: `extraction_options()`'s
`max_tokens: None` and `retry.rs`'s 8s backoff floor.

## Tests

8 new tests across both adapters, covering both truncation disguises (present but
partial payload, and no payload at all).

**All 8 fail against the pre-fix adapters** — verified by reverting only the two
adapter files and keeping the tests — with the exact production symptom: N
identical re-asks where 1 was expected, dying with the misclassified error string.

```
a_truncated_native_answer_is_reported_as_truncation_not_as_unparseable_json  FAILED (3 calls, expected 1)
a_blank_truncated_native_answer_is_reported_as_truncation                    FAILED (3 calls, expected 1)
truncation_without_a_tool_use_block_is_reported_as_truncation                FAILED (5 calls, expected 1)
```

They run under a normal workspace build: `bedrock` is not a default feature of
`cognee-llm`, but `cognee-components` default-enables `cognee-llm/bedrock` and
cargo unifies features, so `cargo test --workspace` executes all three Bedrock
tests. Confirmed by running it.

## What this does and does not fix

It converts a ~4-minute hang into an **immediate, correctly-named error**. It does
not make the extraction succeed.

With no caller budget, `current` always equals `effective_output_cap()`
(`min(model cap, ceiling)`), so a truncated extraction now fails terminally on the
first attempt instead of on the third. Making it *succeed* needs more output room,
and the obvious knob is a trap worth recording: `auto_chunk_size` is
`max_completion_tokens / 2`, so raising `LLM_MAX_COMPLETION_TOKENS` also doubles
chunk size and therefore output size. It fights itself. That is a separate design
question and not in scope here.

## Reviewer notes

- The raise branch on Bedrock is currently unreachable for a `None` caller budget
  (`current == cap` always). It is retained for symmetry with Anthropic and
  OpenAI rather than deleted, but a maintainer may prefer it gone.
- `is_truncated()` is `stopReason == "max_tokens"`. It is correctly *false* for
  `content_filtered` / `guardrail_intervened`, where raising the budget would not
  help and the corrective re-ask remains the right handling.
- Non-truncated paths are byte-identical; only their reachability narrows. A
  response unparseable for a non-truncation reason still gets the corrective
  re-ask.

### DIFF
--- a/crates/llm/src/adapters/anthropic.rs
+++ b/crates/llm/src/adapters/anthropic.rs
@@ -548,6 +548,9 @@ impl AnthropicAdapter {
 
         let mut last_error =
             LlmError::InvalidResponse("No structured-output attempt made".to_string());
+        // Set only by the truncation branch, so the retry log can distinguish an
+        // expensive truncation loop from a routine corrective re-ask.
+        let mut truncation_retry = false;
 
         for attempt in 0..self.structured_output_retries {
             if attempt > 0 {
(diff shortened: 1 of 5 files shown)


GraDKh:
Review addressed in `de7f2c4a`. Three findings were correct and all three are fixed; the first two are corrections to my own additions, not to the workflow's original port.

**The caller-budget refusal is reverted.** You were right that it breaks `a_truncated_answer_is_re_asked_with_a_raised_budget` — I confirmed the failure locally, and the tree was briefly holding two tests asserting opposite contracts. I had run only the truncation suites and `cargo test -p cognee-llm`, which does not exercise `bedrock_integration` with the bedrock feature; that is on me.

Your fifth finding is what settles it rather than the broken test: `GenerationOptions::default()` sets `max_tokens: Some(DEFAULT_MAX_COMPLETION_TOKENS)`, so `..Default::default()` is indistinguishable from a deliberate budget. "The caller set it explicitly" is not a signal this adapter can read. Any new call site forgetting `max_tokens: None` would have silently lost truncation recovery on Bedrock — a worse failure than the one I was guarding against, and invisible.

Bedrock now raises on the native branch exactly as it already did on the fallback branch. That also disposes of your second finding: the error naming a possibly-clamped, never-sent budget is gone with the branch that produced it.

**The privacy guard is now actually enforced.** You are right that `InvalidResponse` is not restricted to self-synthesised errors — `converse::map_error` returns it carrying the raw HTTP body for a 400/ValidationException, so my `InvalidResponse(msg) => msg` arm reopened the exact leak the hunk existed to close. Fixed with `LlmError::log_kind()`, a payload-free discriminant over all 16 variants; the retry log now renders only that. The truncation case keeps its own dedicated message, which quotes no payload.

**The latch is cleared per attempt** in both adapters, so a truncation on attempt 0 no longer mislabels a later validator-driven re-ask — which, as you note, was reintroducing precisely the WARN-per-chunk noise the change exists to avoid.

Verification after the fixes: `bedrock_integration` 12 passed (was 11/1), `bedrock_truncation_retry` 3 passed, `anthropic_truncation_retry` 5 passed; fmt, check and `clippy -D warnings` clean both with and without `--features bedrock`.

**One thing I have not changed, deliberately.** The network retry ladders at `bedrock/mod.rs:327` and `anthropic.rs:351` still log `error = %last_error` and have the same payload-disclosure shape. Both are pre-existing on `main` and outside this diff, so I left them rather than widen the change — but they should get the same `log_kind()` treatment in a follow-up, and I would rather record that here than have it look overlooked.

PR #190, opened 2026-09-08
fix(llm): omit maxTokens when the caller sets none, and drop the 0.1 extraction temperature

GraDKh:
Two Python-parity fixes for the Bedrock path, plus the instrumentation needed to
tell which one actually mattered.

## The evidence

A real BYOD cognify run returned **HTTP 500**: 3 of 55 chunks truncated at 16384
output tokens, and `RollbackScope::WholeRun` discarded the other 52. Python, on the
same document with the same model, **succeeded in 74s**.

A subagent captured Python's actual Converse body through the real
`litellm 1.100.0` + `instructor 1.15.1` stack:

```json
{"messages":[…], "inferenceConfig": {}, "system":[…], "toolConfig":{…}}
```

Python's Bedrock adapter accepts `max_completion_tokens: int = 16384`, stores it, and
**never serialises it** — a grep for the name in that file returns only the parameter
and the assignment. AWS documents the omitted case: *"The default value is the maximum
allowed value for the model that you are using"* — **64000** for Sonnet 4.5.

Rust sent `"maxTokens": 16384` on every call. It was running the same extraction with
**3.9× less headroom**, and because its ceiling equalled its retry cap, the truncation
ladder added in #187 had **nothing to climb into** and failed terminally on attempt 1.

Worth noting what this is *not*: Python's chunks are 8191 tokens against Rust's ~690,
and Python made ~5 extraction calls to Rust's 55. Rust's input per call is ~12×
**smaller**. The 171 `Completed graph extraction for DataPoint` records in Python's log
are graph-storage nodes from an in-memory traversal, not LLM calls.

## Fix A — `None` now means None on the wire

`effective_max_tokens` substituted the configured ceiling for a caller that passed no
budget, reversing the caller's stated intent. `extraction_options()` sets
`max_tokens: None` with a comment saying it means *"use the model's full default output
budget, for Python parity"* — and the adapter turned that into 16384.

It now returns `Option<u32>`, and `inference_config` omits the key entirely when `None`.
`effective_output_cap` takes the options too: with no caller budget the model's own
maximum was already in force, so a truncation there is genuinely terminal rather than
something to re-ask at the identical number.

**Deliberate policy change:** `llm_max_completion_tokens` no longer bounds a caller that
passes no budget. That is exactly Python, where the setting only sizes input chunks
(`cognee/infrastructure/llm/utils.py`) and never reaches the request. An explicit caller
budget is still clamped by the ceiling then the model cap, and `transcribe_image` still
clamps because it always sends an explicit budget.

## Fix B — extraction no longer sends `temperature: 0.1`

Python declares `llm_temperature` and never passes it to any adapter, so its body carries
no temperature and the model default applies. Very low temperature is a known driver of
degenerate repetition — the leading suspect for why three ~690-token chunks demanded
>16384 output tokens *at all*.

Summarization's `0.3` is left alone: it wasn't implicated, and changing it would blur the
experiment below.

## Telling the two apart

Shipping both together makes a green run uninformative, so the structured-output loop now
logs input/output tokens, the truncation flag and the budget actually sent, at INFO:

- **output_tokens near the expected ~5–6k** → the runaway was temperature-driven; Fix B is
  what mattered.
- **some calls legitimately exceed 16384** → the runaway is real; Fix A is what saved them.

Without this, one run cannot distinguish them.

## Tests

- An absent budget is pinned at the wire level: the block equals `{}`, matching Python.
- The ceiling-clamp guarantee for **explicit** budgets is pinned separately, so Fix A
  can't quietly remove operator control where it still applies.
- The #187 truncation tests now assert the model cap (64000) rather than 16384.
- `generate(..., None)` correspondingly sends no budget; its test is renamed and
  re-pinned rather than deleted, since the old name asserted the behaviour being changed.

`cargo test -p cognee-llm --features bedrock`: 12 suites, 0 failures.
`cargo test -p cognee-cognify --lib`: 325 passed. fmt and `clippy -D warnings` clean.

## Deliberately not in scope

Whether a truncated chunk should fail the whole run. A parallel investigation found
Python does the same thing — `asyncio.gather` **without** `return_exceptions=True`,
whole-run rollback in `rollback.py`, HTTP 500 — and that parity is already pinned by the
executed harness in `e2e-cross-sdk/failure-parity/`. It is parity, not a bug.

Still open and worth its own change: Rust discovers the failure only when the whole batch
joins (184s/301s observed) where Python fails at the first bad chunk.


GraDKh:
## Correction: the Fix B justification was wrong, the fix is not

I claimed Python ignores `llm_temperature`. That is **false against latest main**, and I have corrected the code comment in `c5b0e873`.

What happened: the clone I checked was `dd8a1659d` — exactly tag **`v1.4.0`**, the image this divergence was measured against. Latest main is `e93a4f0c` / **1.5.4**, and upstream fixed this deliberately in **1.5.0** (PR #4477, COG-6174, "Plumb LLM_TEMPERATURE into calls, add LLM_SEED"), with nine tests in `tests/unit/infrastructure/llm/test_llm_config.py` pinning the behaviour.

The mechanism is not a named argument, which is why grepping the adapters for `temperature=` returns nothing and looks conclusive. A `@model_validator` folds it into `llm_args` (`config.py:260-290`), and every adapter already merges `{**self.llm_args, **kwargs}`. Captured request bodies on 1.5.4:

| config | Bedrock `inferenceConfig` on the wire |
|---|---|
| `LLM_TEMPERATURE=0.123` | `{"temperature": 0.123}` |
| unset | `{}` |
| `LLM_ARGS={"temperature":0.456}` | `{"temperature": 0.456}` |

So `LLM_ARGS` was a working escape hatch even on 1.4.0 — "ignored" was too strong even there.

**The fix itself stands.** On both 1.4.0 and 1.5.x the *default* path sends no temperature, and no deployment we ship sets `LLM_TEMPERATURE`. Rust pinned `0.1` on every extraction call, so the divergence is real either way. What narrows is the reasoning: not "Python ignores it" but "Python omits it when unset, and it is unset here".

**A gap this leaves, recorded rather than quietly fixed:** nothing threads Rust's configured `llm_temperature` (`crates/lib/src/config.rs:151`) into cognify, so an operator who sets it is ignored. Matching 1.5.x properly means honouring an explicitly-set value and omitting it otherwise — which needs the same explicitly-set-vs-default distinction `llm_max_completion_tokens` also lacks, so it belongs in its own change.

Fix A is unaffected: Python's Bedrock adapter demonstrably ships `"inferenceConfig": {}` when no budget is set, on both versions.

GraDKh:
## Added: `llm_temperature` is now honoured when explicitly set

`5bb0ad62`. This completes the temperature half — dropping the hardcoded `0.1` fixed the *default* path, but the *configured* path did not work on any surface.

Python's rule has three cases and Rust implemented none of them:

| | Python 1.5.x | Rust before | Rust now |
|---|---|---|---|
| caller passes one | wins | ignored (0.1 pinned) | wins |
| unset caller, operator set | operator's value | ignored (0.1 pinned) | operator's value |
| neither | key absent | 0.1 sent | key absent |

### `Option<f64>`, because "explicitly set" has to be representable

`llm_temperature: f64` defaulting to `0.0` cannot distinguish "unset" from "set to zero". It is now `Option<f64>`, and `None` is **not** the same as `Some(0.0)` — Python draws that exact distinction for a concrete reason: the default gpt-5 family rejects any temperature but the provider's own, so a silently-defaulted `0.0` breaks those models.

Threading: `Settings` → `LlmInputs.temperature` → `BedrockAdapter::with_default_temperature` → `converse::inference_config`.

### The HTTP server needed its own field, and that is the point

It had **no temperature field at all**. The deployed BYOD stack runs that surface, so without this the setting would have been plumbed everywhere except the one place it was wanted — and the change would have looked correct while doing nothing. It now reads `LLM_TEMPERATURE` and lowers it like the CLI/SDK path.

### Scope, deliberately narrow

**Bedrock only.** The same builder would suit the Anthropic and OpenAI adapters, but neither was implicated here and neither has been checked against upstream, so widening it belongs in its own change. My first attempt at this patch accidentally landed on the Anthropic factory and failed to compile — a useful reminder that the two call sites are not interchangeable.

Also still absent: Python's **local-inference carve-out**, where an unset temperature becomes `0.0` for ollama / llama_cpp / lm_studio (`is_local_llm`). Rust has no equivalent notion, so that is a separate piece of work.

### One precision detail

`GenerationOptions.temperature` is `f32`, so an operator setting `0.9` reaches the wire as `0.8999999761581421`. Bedrock accepts it. The new test uses f32-exact values (0.25, 0.5, 0.0) and documents why, rather than asserting on round decimals and appearing to fail.

### Gate

`cognee` 126, `cognee-http-server` 280, `cognee-cli` 15, all `cognee-llm --features bedrock` suites green; fmt and `clippy -D warnings` clean. The three-way contract is pinned, including `Some(0.0)` as a real value.

GraDKh:
All ten review findings addressed — `6fed5fcf` and `1a06e0a`.

## The one that mattered

**Finding 1 was correct and would have silently reverted this PR for every existing user.** `Settings` derives `Serialize` with no `skip_serializing_if` and `save_config` writes the whole struct, so anyone who ever ran `cognee-cli config set <anything>` has `"llm_temperature": 0.0` on disk — I confirmed it on this machine at `~/.config/cognee-rust/config.json:19`.

Under the new `Option` that reads back as `Some(0.0)`: an explicit temperature of zero on every Bedrock call, chosen by nobody, and the *extreme* of the setting this release stopped sending by default.

`ConfigDocument` gains **v3**, mapping a persisted default back to unset while keeping a value the user actually chose. Two tests pin both directions. This is the same trap the v1→v2 `chunk_size` migration exists for, one field over — the reviewer spotting the precedent is what made the fix obvious.

## Docs that contradicted the code

`effective_max_tokens`' rustdoc still claimed the ceiling "bounds every path"; the field doc still described `min(this, model cap)`. Both now state the behaviour **and the consequence**: `llm_max_completion_tokens` does not bound a cognify call, so a runaway can bill to the model maximum with no knob. That is the sentence an operator needs before relying on the setting, and `cargo doc` is a CI job.

The CHANGELOG documented Bedrock taking the configured ceiling — which this PR reverses. Corrected, and both breaking changes are now written up, including the `Option<f64>` type change that breaks Rust SDK consumers.

## Scope corrected rather than expanded

`LLM_TEMPERATURE` reaches the wire on **Bedrock only**; the OpenAI-compatible, Azure and Anthropic factories ignore `LlmInputs::temperature`. Both doc sites now say so instead of implying the setting is honoured generally.

`AnthropicAdapter::effective_max_tokens` has the identical absent-budget bug and reproduces this incident on `LLM_PROVIDER=anthropic` with `claude-sonnet-4-5`. It cannot be fixed the same way — Anthropic *requires* `max_tokens`, so omission is not an option and resolving to the model cap is a behaviour change deserving its own tests. Recorded at the exact site rather than left to be rediscovered.

## The rest

- `with_default_temperature` had been inserted between `with_extra_args` and its doc comment; reordered. My reorder then left an orphaned blank line that `clippy -D warnings` rejected — fixed in `1a06e0a`, and it would have failed the CI lint job.
- The two older retry logs read `maxTokens` with `unwrap_or(0)`, so an absent budget logged `max_tokens = 0` — misleading in exactly the run this PR wants to diagnose. They now log the `Option`, matching the new usage line.
- `config get` emitted `null` but `config set` rejected it, so there was no way back to unset. Both the CLI and `ConfigManager::set` now accept it, mirroring `chunk_size`.
- `docs/configuration.md` still listed the default as `0.0`.

## Not changed

Finding 8 (per-call INFO instrumentation volume) is deliberate and temporary: it is the only thing that distinguishes which of the two fixes did the work, and one production run settles it. Worth demoting to `debug!` immediately afterwards — noted so it does not become permanent by inattention.

## Gate

`cognee-cli` 17, `cognee` 126, `cognee-http-server` 280, all `cognee-llm --features bedrock` suites green; fmt and `clippy -D warnings` clean.

PR #203, opened 2026-09-09
fix(bedrock): count real tokens, and survive a runaway generation

GraDKh:
Five consecutive AWS BYOD deployment runs failed to cognify a single PDF. This is
the fix, in four commits, each of which was proven necessary by a run that failed
*after* the previous one landed.

## What was wrong

`TokenCounterKind::from_env` selects a BPE counter for the openai and bedrock
providers. Nothing in the server build compiled one, so `build()` handed back a
`WordCounter` and said so only on raw stderr.

That is not a rounding error, it is a unit change. `auto_chunk_size` derives its
number from the **embedding model's token limit**, so a word counter spends a
token budget in the wrong currency. On Bedrock with Titan v2 the 8191-token
budget became 8191 *words* — ~11149 real tokens against an 8192 cap — and every
embedding call returned HTTP 400. `RollbackScope::WholeRun` then swept the 99
nodes and 335 edges extraction had already produced.

Separately, ~4.5% of Bedrock structured-output calls run away and emit until they
hit the model maximum. At the cap the adapter returned terminally, so two bad
chunks out of 55 discarded all 53 good extractions.

## The commits

1. **`fix(llm)`** — re-ask a cap truncation instead of failing the run. The
   failure is stochastic, not a property of the input: the runaway chunk ids
   differ between runs of the same document, and in one run the transport layer
   re-POSTed an identical body and got a normal 686-token answer 11s later.
   Bounded by `structured_output_retries`.
2. **`fix(chunking)`** — dispatch bedrock to a BPE counter, and convert a token
   budget before a word counter spends it.
3. **`fix(chunking)`** — `TokenCounterKind::effective()`. Commit 2 was correct and
   inert: it reasoned about the *requested* counter while `build()` degrades to
   whitespace at runtime. Sizing now follows what will actually run.
4. **`fix(http-server)`** — `tiktoken` in the default feature list, and the three
   fallback notices promoted from `eprintln!` to `tracing::warn!`.

## Where the default does *not* go

Not on `cognee-chunking`. That was the first attempt and `scripts/check_all.sh`
rejected it: `tiktoken-rs` is on the forbidden list for `cognee-cli/android-default`
alongside `lancedb`, `pdfium-render` and `aws-config`, because a leaf-crate default
flows past every `default-features = false` upstream of it. The guard's own
wording — *"a feature forward has stopped subtracting"* — describes exactly what
that change did.

`cognee` (lib) has carried `"tiktoken"` in its defaults all along, so the library
was never the gap: the server image builds `--no-default-features` with an
explicit list, and `cognee-http-server` had no `tiktoken` feature to name.

## Verification

Run 6 on AWS BYOD, same stack and document as the five failures:

| | run 5 | run 6 |
|---|---|---|
| `falling back to WordCounter` | present | 0 |
| Titan input-cap 400s | every embed call | 0 (peak input 7290 / 8192) |
| cognify | HTTP 500 | **HTTP 200 in 486s** |
| searches | never reached | 4/5 HTTP 200, grounded answers |

Local: `scripts/check_all.sh` — Rust compile/clippy/all feature variants, the
android forbidden-list **and both sentinels**, C API (9 steps + slim-build
`CG_ERR_FEATURE_NOT_BUILT`), Python (128 passed / 100 skipped), TS. Java skipped
locally (no JDK); CI covers it. cognify 327 green in all three feature states.

## Known and not addressed here

One runaway generation still cost ~352s of run 6's 486s — 72% of wall-clock —
because extraction passes `max_tokens: None` as a deliberate Python-parity
invariant. Python carries the same `None` and does not hit this, so the useful
question is what protects Python, not what number to pick for Rust. Left alone
rather than guessed at.

The deployment image passes `--no-default-features`, so it still needs `tiktoken`
in its `BASE_FEATURES` for the new default to reach it.

### DIFF
--- a/crates/chunking/src/config.rs
+++ b/crates/chunking/src/config.rs
@@ -195,27 +195,33 @@ impl TokenCounterKind {
             // the user that their configured tokenizer is not active.
             #[cfg(not(feature = "hf-tokenizer"))]
             TokenCounterKind::HuggingFace { model_id: _ } => {
-                eprintln!(
+                tracing::warn!(
                     "cognee-chunking: HuggingFace tokenizer requested but `hf-tokenizer` \
-                     feature is not enabled — falling back to WordCounter"
+                     feature is not enabled — falling back to WordCounter. A token budget \
+                     spent in words undercounts real tokens by ~35%, which can \
+                     exceed an embedding model's input cap."
                 );
                 Ok(Box::new(WordCounter))
             }
 
             #[cfg(not(feature = "hf-tokenizer"))]
             TokenCounterKind::HuggingFaceFile { path: _ } => {
-                eprintln!(
+                tracing::warn!(
                     "cognee-chunking: HuggingFaceFile tokenizer requested but `hf-tokenizer` \
-                     feature is not enabled — falling back to WordCounter"
+                     feature is not enabled — falling back to WordCounter. A token budget \
+                     spent in words undercounts real tokens by ~35%, which can \
+                     exceed an embedding model's input cap."
                 );
                 Ok(Box::new(WordCounter))
             }
 
             #[cfg(not(feature = "tiktoken"))]
             TokenCounterKind::TikToken => {
-                eprintln!(
+                tracing::warn!(
                     "cognee-chunking: TikToken tokenizer requested but `tiktoken` feature is \
-                     not enabled — falling back to WordCounter"
+                     not enabled — falling back to WordCounter. A token budget spent \
+                     in words undercounts real tokens by ~35%, which can exceed an \
+                     embedding model's input cap."
                 );
                 Ok(Box::new(WordCounter))
             }
--- a/crates/chunking/src/config.rs
+++ b/crates/chunking/src/config.rs
@@ -96,18 +96,11 @@ impl TokenCounterKind {
                 TokenCounterKind::Word
             }
             "openai" | "openai_compatible" => TokenCounterKind::TikToken,
-            // Bedrock previously fell through to `Word`, and whitespace counting
-            // against a token budget is not a rounding error — it is a different
-            // unit. `auto_chunk_size` derives its number from the EMBEDDING
-            // MODEL'S TOKEN LIMIT (8191 for Titan v2), so counting words spent
-            // that budget in the wrong currency: 8191 words measured ~11100
-            // Titan tokens on real text, and every embed call 400d with
-            // "Too many input tokens. Max input tokens: 8192".
-            //
-            // Titan's own tokenizer is not published, so no counter here is
-            // exact. cl100k BPE is nonetheless far closer to it than whitespace
-            // (~1.3 tokens/word vs 1.0 by construction), and the clamp in
-            // `CognifyConfig::with_auto_chunk_size` covers the residual error.
+            // Titan's tokenizer is not published, so no counter here is exact.
+            // cl100k BPE is the closest available proxy: both are byte-level BPE
+            // over a ~100k vocabulary and agree within a few percent on prose,
+            // where whitespace counting is off by ~35% and in the wrong unit
+            // entirely. `fit_token_budget` absorbs the residual difference.
             "bedrock" => TokenCounterKind::TikToken,
             "ollama" => {
                 if let Ok(model_id) = std::env::var("HUGGINGFACE_TOKENIZER")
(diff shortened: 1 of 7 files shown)


crates/chunking/src/config.rs
```diff
@@ -96,6 +96,19 @@ impl TokenCounterKind {
...
+            // MODEL'S TOKEN LIMIT (8191 for Titan v2), so counting words spent
+            // that budget in the wrong currency: 8191 words measured ~11100
+            // Titan tokens on real text, and every embed call 400d with
+            // "Too many input tokens. Max input tokens: 8192".
+            //
+            // Titan's own tokenizer is not published, so no counter here is
+            // exact. cl100k BPE is nonetheless far closer to it than whitespace
+            // (~1.3 tokens/word vs 1.0 by construction), and the clamp in
+            // `CognifyConfig::with_auto_chunk_size` covers the residual error.
+            "bedrock" => TokenCounterKind::TikToken,
```
GraDKh: This text is too vague, just put a comment why the tiktoken is an proper choice for the bedrock. Also tiktoken may be disabled by feature, need a fallback with a warning or error in that case

crates/chunking/src/config.rs
```diff
@@ -108,6 +121,46 @@ impl TokenCounterKind {
...
+    /// asking for `TikToken` in an image built without the `tiktoken` feature
+    /// yields whitespace counting, and any token budget handed to it is spent in
+    /// the wrong unit.
+    ///
+    /// That is not hypothetical — it shipped. A build requested TikToken for
+    /// Bedrock, degraded silently to `WordCounter`, and the resulting 8191-word
+    /// chunks measured ~11149 real tokens against an 8192-token embedder limit,
+    /// failing every embedding call with HTTP 400.
+    #[must_use]
+    pub fn effective(&self) -> TokenCounterKind {
```
GraDKh: Oh, I see, here is a fallback. Do we need to have it in public? I'd rather have a single function that takes config and allways returns effective

crates/cognify/src/config.rs
```diff
@@ -654,10 +654,49 @@ impl CognifyConfig {
         embedding_engine: &dyn EmbeddingEngine,
         llm: &dyn Llm,
     ) -> Self {
-        self.max_chunk_size = Some(Self::auto_chunk_size(embedding_engine, llm));
+        let budget = Self::auto_chunk_size(embedding_engine, llm);
+        // `effective()`, not the configured kind: a requested BPE counter degrades
+        // to whitespace when its cargo feature is absent, and sizing against the
+        // request rather than the reality is how an 8192-token embedder limit was
+        // blown by an 8191-"token" budget that was really 8191 words.
```
GraDKh: Remove this comment

crates/chunking/src/config.rs
GraDKh: Trimmed to the actual rationale: cl100k and Titan are both byte-level BPE over a ~100k vocabulary and agree within a few percent on prose, where whitespace counting is off by ~35% and in the wrong unit entirely. The incident retelling is gone.

On the disabled-feature case — you found it in your next comment: `build()` falls back to `WordCounter`. Two things changed there. It now warns through `tracing::warn!` instead of `eprintln!`, which is why five deployment runs of structured logs never showed it; and `fit_token_budget` resolves the *effective* counter, so the fallback no longer silently keeps a token-unit budget.

I went with warn rather than error deliberately: `--no-default-features` builds (the wasm32 CI lane, the Android profile) legitimately have no BPE counter, and erroring would break them.

crates/chunking/src/config.rs
GraDKh: Done, and this was the right call — `effective()` was an API that asked every caller to remember an extra step, and forgetting it *was* the bug.

`effective()` is now private. The single public entry point is `TokenCounterKind::fit_token_budget(token_budget)`: it resolves the counter that will really run and converts the budget in one call. There is no longer a way to size a budget against a counter that will not run.

`CognifyConfig::fit_budget_to_counter` moved into cognee-chunking with it, next to the counter whose units it reasons about, and its tests came along. The cognify call site is now just `self.token_counter_kind.fit_token_budget(budget)`.

crates/cognify/src/config.rs
GraDKh: Removed — and it stopped being needed rather than just deleted. With the conversion behind `fit_token_budget`, the call site is one line that says what it does:

```rust
self.max_chunk_size = Some(self.token_counter_kind.fit_token_budget(budget));
```

PR #205, opened 2026-09-09
fix(llm): pace and bound Bedrock dispatch like the other providers (SDK-612)

GraDKh:
Closes **SDK-612**.

## Why

SDK-503 wired a token bucket and a 900s overload cooldown into the OpenAI-compatible, Anthropic and Azure adapters. **Bedrock got neither.** The factory never installed a pacer, and the adapter never consulted one — `grep -rn 'pacer' crates/llm/src/adapters/bedrock/` returned nothing.

So on Bedrock, throttling was retried with backoff but never *paced*: no episode opened, and the next wave went straight back at a provider that had just said it was saturated. That is the failure mode SDK-503 exists to prevent — the calibration run lost 218 of 1,950 summaries to rate limiting — and Bedrock is a first-class native provider (`3d1982e`), not a compatibility shim.

`acquire_in_flight` was missing here too. All three other adapters take it, so Bedrock also sat outside the process-wide concurrency ceiling. Same protection layer, so both are fixed together.

## What

Both halves mirror the Anthropic adapter, which is the closest reviewed template:

- **`BedrockAdapter`** gains `pacer`, `with_pacer`, and a `pacer()` accessor falling back to the process-wide `llm_pacer()`. `call_converse` admits per attempt → takes the in-flight permit → re-gates before the send when the first admission was the free fast path, so an attempt never spends two tokens. Overload is recorded from a transport timeout and from the shared status classifier.
- **`BedrockLlmFactory::build`** calls `install_pacer`, which installs the pacer *and* the in-flight ceiling. First-call-wins, so it is a no-op in a mixed process and the only thing that installs anything in a Bedrock-only one.

Two things worth a reviewer's eye:

- **No Bedrock-specific overload notion was needed.** Bedrock reports `ThrottlingException` as 429 and a busy or not-yet-ready model as 503, both already classified by `crate::retry::overload_reason`. I checked before adding a parallel concept.
- **Recording happens before the terminal check.** A 429 that turns out to be terminal still means the provider is saturated right now.

Unlike the OpenAI and Anthropic loops there is **no queued-time bookkeeping**: this is a fixed attempt count with no minimum-elapsed floor, so time spent queuing cannot cut the retry budget short. Adding the subtraction here would be cargo-culting.

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets --features bedrock -- -D warnings` — clean.
- `bedrock_pacing` (4 new) + `bedrock_integration` — 16 passed, 3 ignored (live).
- **Mutation-checked in both directions.** Forcing the reason to `None` fails exactly the two positive cases; forcing it to `Some` fails the 400. The negative cases are the point — testing only that a 429 paces would pass just as well against an unconditional `record_overload`.

New tests use the existing `bedrock_integration.rs` harness shape: a real adapter pointed at `httpmock` through `AwsInputs::bedrock_runtime_endpoint`, so the real endpoint chain runs. The pacer is injected with `with_pacer` rather than installed process-wide, so these cases do not need a process of their own the way `pacing_after_in_flight_queue.rs` does.

## Unrelated failure seen while testing

`integration_openai::test_simple_text_generation` fails on this branch **and with these changes reverted** — verified. It is a live-API test asserting non-empty content while the configured model (`gpt-oss-120b`) spends all 10 allowed tokens on reasoning and returns empty content. Pre-existing and environmental; not touched here, but worth a ticket.

https://claude.ai/code/session_01BfBnTT1Ya6K6vGsD2GZnyU


GraDKh:
## Rebased onto `main` — `047672e`

#215 (SDK-624) landed on both files this PR touches, so it had gone conflicting. Rebased rather than merged, and force-pushed with `--force-with-lease`.

**Both conflicts were purely additive** — #215 added `build_http_client`, `with_http_timeouts`, `with_request_deadline` and `deadline_exceeded` at the same point this adds `with_pacer`/`pacer()`, and a `started` local where this adds a `pacer` local. Neither side replaces the other; both kept.

**The gap this PR fixes is still real on current `main`** — I checked rather than assuming, since #215's title mentions Bedrock:

- `install_pacer` is called by the OpenAI-compatible factory (`:122`), its transcriber (`:166`), Anthropic (`:210`) and Azure (`:275`). `BedrockLlmFactory` begins at `:322` and does not call it.
- `grep -n 'pacer|llm_pacer|acquire_in_flight' crates/llm/src/adapters/bedrock/mod.rs` at `origin/main` returns nothing.

#215 gave Bedrock the *time* bounds (deadline, connect/request timeouts). It did not give it rate limiting or the in-flight ceiling, which is what this adds. The two are complementary: a deadline caps how long one call may take, the pacer caps how often calls start and opens the cooldown after a throttle.

One thing worth noting for a reviewer: #215 renamed the retry knobs (`LLM_MAX_RETRIES` / new `LLM_NETWORK_RETRIES`) and the deadline now clamps the backoff to the remaining budget. The admission point here sits *after* that clamp and before the send, so a paced wait is inside the deadline it is measured against — which is the correct side, since the alternative would let pacing silently extend a budget the operator set.

### Re-verified after the rebase

`fmt --check` and `clippy --all-targets --features bedrock -- -D warnings` clean; `bedrock_pacing` + `bedrock_integration` 16 passed, 3 ignored (live).

Mutations re-run against the reshaped loop, not assumed to still hold: forcing the reason to `None` fails exactly the 429 and 503 cases; forcing it to `Some` fails the 400. Same result as before the rebase.

GraDKh:
## Review findings addressed — `0fe0a53`

**The medium one was a real gap, and my own reasoning caused it.** My commit message argued no queued-time bookkeeping was needed here. That was right for the mechanism I had in mind and wrong for this one — there are two. The `queued_for_permit` subtraction protects a minimum-elapsed retry *floor*, which this ladder genuinely lacks. The post-wait guard enforces the aggregate *deadline*, which Bedrock gained in SDK-624 — the PR this branch was rebased onto hours earlier. The argument held before the rebase and stopped holding after it.

Ported. Without it, `LLM_REQUEST_DEADLINE_SECONDS` does not bound a Bedrock call the way `docs/configuration.md` says it does.

The predicate is **extracted rather than inlined**, because its difficulty is entirely in what it must *not* do: three of its four conditions exist to exclude a case that has to keep running — the first attempt, a budget already spent before the wait began (the top-of-loop guard owns that and clamps the backoff instead), and a disabled deadline. Six unit tests pin those plus the boundary instant.

Two things worth recording that the review prompted me to check:

- **This guard is asserted by no test in the OpenAI or Anthropic adapters either.** `openai_request_deadline.rs` covers the cascade guard and the transport-ladder guard, not this one. So the six tests here are the first coverage it has anywhere.
- **It detects rather than preempts, in all three adapters.** It runs after the wait returns, so a wait far longer than the budget still blocks for its full length and *then* reports. Bounding the wait itself needs a timeout on `admit` and `acquire_in_flight`, which none of them has. Noted in the helper's docs rather than left to be rediscovered.

Also fixed: the `network_retries` doc deferred unifying the retry ladder to SDK-612, which is this branch. Restated as a deliberate divergence — a retry floor here would *lengthen* Bedrock calls, the opposite of what the deadline is for.

**Not fixed here:** the Bedrock embedding engine is still unpaced. That is a real gap on the sibling path and the PR body overclaimed by implying otherwise; it wants its own ticket rather than scope creep into an LLM-adapter change.

`fmt --check`, `clippy --all-targets --features bedrock -- -D warnings` clean; 190 passed, 3 ignored across the Bedrock suites.

PR #208, opened 2026-09-09
feat(cli): pipeline-unblock — clear what a killed run leaves behind (SDK-616)

GraDKh:
Closes **SDK-616**.

## Why

The exclusive-run claim is released on every graceful exit path, errors included. A process that is **killed** — SIGKILL, OOM, a hard crash — cannot release, so the row outlives it and refuses every later run on that `(dataset, pipeline)` pair. `CLAIM_STALE_AFTER` is **24 hours**, and there was no way to clear it: no CLI command, no HTTP route, no flag. The existing release is scoped to the holder's `claim_id`, which is precisely what an operator staring at a dead process does not have.

This compounds with **SDK-507**: if a large cognify is OOM-killed — the scenario that ticket exists for — the dataset is locked for a day.

## The design decision, and why it isn't a "reset" command

A claim **older** than the staleness window is already reclaimed by `try_claim_pipeline_run` without anyone asking. So the only claim this command can ever find is a *young* one — either a live run, or a holder that died minutes ago. Nothing available here distinguishes them, and releasing a live claim re-admits the concurrent run the claim exists to prevent.

An age-based safety rail would therefore refuse **exactly** the case the command is for. So instead: **report by default, release only on `--release`**, with the operator asserting the holder is gone. The report names the holder and the claim's age, because the age is the only evidence available.

## What

Two `PipelineRunRepository` methods, both with default implementations so the three existing implementors keep compiling:

| Method | Note |
|---|---|
| `get_pipeline_run_claim` | Holder and `claimed_at`. A row whose `claim_id` does not parse is an **error**, not `None` — reporting a corrupt claim as absent would tell an operator the pair is free while the row keeps refusing their runs. |
| `force_release_pipeline_run_claim` | Unscoped by holder (the point), but still scoped by dataset **and** pipeline. Logs a warning naming the risk; nothing on the run path may call it. |

```
cognee-cli pipeline-claim -d <dataset> [--pipeline <name>] [--release]
```

Rejected inside `run-sequence` — a sequence that pauses to ask a human whether a holder is dead is not a sequence.

`PipelineAlreadyRunning` now names the remedy. The most common way to reach that error is not a concurrent run at all, and the operator who has just had a cognify killed should not have to find this command by reading source.

## Verification

- `cargo fmt --check`, `cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings` — clean.
- `cognee-database` claim suite: **35 passed** (7 new). `cognee-cli`: 68 passed. `cognee-cognify`: 331 passed.
- **The two scoping tests are mutation-checked, each on its own.** Dropping the dataset filter fails `force_release_does_not_touch_another_dataset`; dropping the pipeline filter fails `force_release_does_not_touch_another_pipeline`. Without them an unscoped delete would quietly clear every claim in the table and the happy-path test would still pass — which is the failure mode worth guarding here, since this is the one call that can re-admit a concurrent run.

## Deliberately out of scope

Shortening or making `CLAIM_STALE_AFTER` configurable. Lowering it re-admits a concurrent run into a legitimately long one — a trade-off the trait docs already warn about, and separate from giving operators an escape hatch.

`chrono` is added to `cognee-cli` (workspace version) to render the claim age.

https://claude.ai/code/session_01BfBnTT1Ya6K6vGsD2GZnyU


GraDKh:
## Reworked after review — `fc5d605`

A code review found that the first version **does not unwedge the scenario it was built for**. It was right. Verified before fixing.

**There are two gates, not one.** `check_pipeline_run_qualification` reads the latest `pipeline_runs` row for the pair and rejects a `Started` one — *before* the claim is consulted. A killed process leaves that row behind exactly as it leaves its claim. So releasing the claim left the next run refused by the first gate with the identical error, right after the command printed "can run again".

**And the two do not expire alike.** The claim ages out after a day. The orphaned run row has **no staleness window at all**: the only sweep that retires one is `reset_orphans`, reached solely from `http-server/src/state.rs:158`. A CLI-only deployment never reaches it, so a killed cognify wedges the dataset *permanently*, not for 24 hours. My ticket understated the bug.

The command now reports and clears both, in the order cognify checks them, and is renamed **`pipeline-unblock`** — it is no longer about claims. New `reset_orphan_run` retires one pair's orphan the way `reset_orphans` retires every pair's: an `Errored` successor rather than a delete, so the audit trail survives and qualification sees the state it treats as proceed.

### Three more from the same review

- **Force-release is scoped to the observed holder.** The unscoped delete had a TOCTOU — if the holder finished and a new run claimed the pair between the read and the release, it killed the live newcomer while reporting the dead holder's id. The `claim_id` was already in hand.
- **A claim past the staleness window is now reported as already reclaimable.** `get_pipeline_run_claim` applies no staleness filter, so the old wording told an operator to leave alone (or release) a claim the next run reclaims by itself — and left them believing the real blocker was gone.
- **`--pipeline` is validated against the three real names.** An unrecognised value looked up a pair that cannot exist and asserted, confidently, that nothing was blocking. The doc also omitted `temporal-cognify`, so an operator wedged on a temporal run could not find the right name.

### Also

Both `PipelineAlreadyRunning` messages now describe the cause **without prescribing a binary** — the earlier wording pointed HTTP and binding callers at a CLI they do not have, for a blocker that on the raising path is not the claim. Memify's variant gained the same text; it claims under its own name and wedges the same way. `docs/tools/cli.md` gained the command.

### Verification

`fmt --check`, `check --all-targets`, `clippy --all-targets -- -D warnings` clean. **41** database cases (13 new), **71** CLI, **331** cognify.

The six new repository cases cover the orphan reset's happy path, that it leaves a completed run alone, that it is scoped by pipeline *and* by dataset, and the TOCTOU. The four CLI cases pin the report wording, including that a stale claim is named as *not* the blocker — the judgement an operator makes from those lines is the whole point, and getting it backwards sends them to do work that changes nothing.

GraDKh:
## All six review findings addressed — `19ed227`

Two were substantive.

**The orphan retire had the TOCTOU I had just fixed one method over.** `try_release_pipeline_run_claim` is scoped to the observed holder precisely so a run that takes the pair between an operator's report and their clear is not killed. `reset_orphan_run` was scoped to the pair alone — so the same sequence would write an `Errored` successor over a healthy in-flight run, marking it failed and re-opening the gate it had legitimately closed. It now takes the observed `pipeline_run_id`. Mutation-checked: removing the guard fails the new case and nothing else.

**`Initiated` was reported as a blocker and is not one.** `check_pipeline_run_qualification` maps it to `Proceed`; only `Started` is refused. I read that mapping while writing the command and filtered on both anyway, so it reported — and with `--clear` "fixed" — a blocker that never existed.

The other four:

- **`--pipeline` no longer defaults to cognify.** That hid a wedged `temporal-cognify` behind "nothing is blocking" — the exact failure `KNOWN_PIPELINES` exists to prevent — and the failing run's error names no pipeline, so an operator could not know what to pass. Omitting it now checks all three.
- **A claim past the staleness window is no longer a blocker.** It was described as "not what is blocking you" and then the operator was told to re-run with `--clear` anyway, and the "nothing is blocking" path could never fire for it. `is_reclaimable` is shared with the report so the two cannot drift.
- **"Cleared … can run again" no longer contradicts the line above it.** A contradicted clear now says so and points at a fresh report.
- **`force_release_…` is now `try_release_…`.** The prefix implied holder-unscoped deletion — the PR description asserted that as fact — while the method is scoped and returns `Ok(false)` without the holder's id. The docs now say there is deliberately no unscoped variant.

`fmt --check`, `check --all-targets`, `clippy --all-targets -- -D warnings` clean. 42 database cases, 71 CLI.

Issue #212, opened 2026-09-10
Tracking (index): leftovers from the Bedrock runaway investigation


GraDKh:
Everything left over from the investigation that landed #210. The cassette-workflow breakage is tracked separately in #211 and is not repeated here.

> **Revision 2026-09-11 — restructured as an index.** This issue previously carried the full analysis inline. That detail now lives in Linear, one ticket per work item, for everything that lands in **this** repo. Items belonging to `cognee-cloud-rs` and `infra` keep their detail here, since they have no ticket yet. The previous full-text revision is preserved in this issue's edit history.
>
> Every claim was re-verified against `cognee-rs` `main` `82ec4379` (past `f0b62a86`, so #209 and #210 are in). **Five claims did not survive that check** — see [Corrections](#corrections) at the bottom. Evidence tags are **MEASURED** (read from a log or reproduced), **VERIFIED** (read from the source cited), **INFERRED** (follows from code, not observed).

---

## A. `cognee-rs` — this repo

Full detail in each ticket. Ordered for execution: impact first, then file collisions, with cheap noise-clearing work early.

- [x] **1.** [SDK-634](https://linear.app/cognee/issue/SDK-634) — Delete three stale experiment branches, including the dropped chunk-failure-tolerance work · `A6` `A2` · **S** — **done**, all three deleted from `origin`; no commit.
- [x] **2.** [SDK-623](https://linear.app/cognee/issue/SDK-623) — #204 ACL follow-ups: swallowed grant failures, missing tenant filter, recall `datasetIds`, stale docs · `B8` · **M** — **done**, merged as `1998ddf2` (PR #214). Spawned items 13 and 14 below.
[...]

PR #224, opened 2026-09-15
feat(llm): add a json_schema constrained-decoding mode with per-schema demotion (SDK-630)

GraDKh:
Closes [SDK-630](https://linear.app/cognee/issue/SDK-630). Item **9** of the tracking index in #212.

Rust had no constrained / grammar-guided structured-output mode on the OpenAI-compatible path: `StructuredOutputMode` was `Auto | Tools | Functions | Json` and nothing ever emitted `response_format: json_schema`. Python **does** emit it on its default `litellm_native` path whenever `litellm.supports_response_schema(model)` is true, so this is a parity gap — and a quality one. Re-tallied from the 22 captured payloads in the `ab_out/` harness: the unconstrained `tools` arms were missing a `target_node_id` in **8 of 11** payloads, while the strict arms were **5/5 clean**.

`LLM_STRUCTURED_OUTPUT_MODE=json_schema` now sends

```json
"response_format": {"type": "json_schema",
                    "json_schema": {"name": "extract_structured_data",
                                    "strict": true, "schema": {…}}}
```

ahead of the `tools` → `functions` → `json` cascade, carrying the all-required / `additionalProperties: false` rewrite OpenAI requires alongside `strict`.

---

### 1. The Bedrock half was already done — the ticket is stale

SDK-630 says Converse's native `outputConfig` is *"**new work**, the adapter has no such support today"*. VERIFIED on `c6bedefa`: it is not. `converse::output_config_block` (`crates/llm/src/adapters/bedrock/converse.rs`) emits `outputConfig.textFormat.jsonSchema` today, selected by `converse::apply_structured_output` from `caps.supports_native_structured_output` — and `bedrock/caps.rs` **is** the Rust equivalent of litellm's model table, so both Anthropic models cognee ships already take the constrained branch and `amazon.nova-lite-v1:0` already falls back to the synthetic `json_tool_call` tool.

So the default-path parity gap the ticket opens with is **already closed on Bedrock**. Nothing is added there, and Bedrock deliberately does not read this knob — it has the table, so asking an operator would be strictly worse.

What this PR does take from that side is the *traversal*: `force_additional_properties_false` moves into `crate::schema` and the new strict transform is built on the same walker, so the two providers' recursion cannot drift when one grows a new combinator. Bedrock's wire shape is unchanged, and a new unit test asserts the part that must **not** change with it (that it leaves `required` alone and honours an explicit `additionalProperties: true`).

### 2. `json_schema` is opt-in, and is not a pin

**`auto` is unchanged and never sends the shape.** This is a deliberate departure from the ticket's "parity gap on the default path" framing, and it should be reviewed as a decision, not read as an oversight.

The reasoning: Python's default path is gated on `litellm.supports_response_schema(model)` — a **model-capability table lookup**. Python does *not* send constrained output to a model it does not know. Rust has that table for Bedrock and uses it; for OpenAI-compatible endpoints there is no table and anything can sit behind `LLM_ENDPOINT` — including Baseten's `gpt-oss-120b`, which answers **HTTP 501** to a constrained request. So the faithful Rust analogue of the Python gate is either a model table or an operator declaration, and this PR takes the declaration. If you would rather have the table (host-gated `api.openai.com` / `*.openai.azure.com` plus a known-family list, mirroring `bedrock/caps.rs`), say so and I will add it — it is additive to everything here.

Turning the knob on is safe even on a heterogeneous fleet, because the mode **demotes** rather than failing — which is the ticket's own "demotion is mandatory, not optional", and is why this value is not a hard pin like the other three:

1. `strict: true` with the strict-rewritten schema;
2. on rejection, the same envelope **without** `strict`, carrying the ordinary shallow schema — for gateways that reject the keyword but handle the response format (older Azure api-versions, some vLLM builds);
3. on rejection again, **out of the mode**: the ordinary cascade runs and behaves exactly as under `auto`.

Both knobs move together at step 2 on purpose. The only MEASURED evidence about this endpoint class is Baseten's, where the recursive all-required rewrite 501s **with or without** `strict: true` (recorded in `structured_output_impl`'s existing comment) — so a demotion that kept the strict schema and dropped only the keyword would be a request nobody has ever seen accepted.

## What moves the ladder, and what does not

Only an outright refusal of the request **shape** — HTTP **400** or **501**. A 429, a non-501 5xx, a network error, a deadline, a bad answer and a model *refusal* all leave the memo untouched, because none of them is evidence about whether the endpoint can parse the body. Getting this wrong in the other direction would disable constrained decoding for the life of the process over one bad minute.

The memo is keyed on a **canonical fingerprint of the schema**, not held as one flag per endpoint — mirroring what `litellm_native` caches per `(model, response_model)`. The two rejections have different scopes and one flag would get one of them wrong:

- Baseten refuses constrained decoding *at all*, so every schema demotes and the endpoint pays **one wasted request per distinct schema per process** (cognify uses a handful);
- a provider that refuses only a *particular* schema (grammar too deep, a recursive `$ref`, too many enum members) demotes just that one and keeps constrained decoding for the rest — which an endpoint-wide flag would have thrown away.

## HTTP 501 becomes terminal — and this is load-bearing

`send_chat_request_before` now maps **501 → `LlmError::FeatureNotSupported`** and treats it as non-retryable, alongside the existing `400..=402 | 404` set. HTTP defines 501 as the server lacking the capability, so no wait can make it succeed.

It is not just tidiness: a retried 501 arrives as `MaxRetriesExceeded`, which `structured_output_impl` treats as **fatal** (correctly — the transport budget now has a time floor). Without this change the demotion would have been **unreachable**, and the mode would have failed the very call the fallback exists to rescue, after burning the whole ladder.

This also changes `generate()`'s behaviour for a 501: an immediate typed error instead of `MaxRetriesExceeded` after the full transport budget.

## Also here

`message.refusal` is now deserialized. Structured outputs return it *instead of* `content`, so without the field a refusal was indistinguishable from an empty answer and the re-ask asked the identical question. The text goes into the corrective instruction and **only** there — it is model output derived from the user's ingested documents, which this crate keeps out of log sinks (`LlmError::log_kind`), so the warning records that a refusal happened, not what it said.

## Tests

`crates/llm/tests/openai_json_schema_mode.rs` — 7 mock-server cases: the wire shape (asserting the optional property really is `required` and every node closed, so the assertion is not vacuous), both rungs of the ladder, the fall-through to the cascade, the per-schema memo (including that a *different* schema is still probed), that a 401 does not demote, that a refusal does not demote, and that `auto` sends no constrained request at all.

Unit tests for `strict_json_schema` (all-required at every depth, through `$defs` reached via an array's `items` — the shape schemars gives `Vec<Edge>` on `KnowledgeGraph`), for the shared Bedrock traversal, and for the fingerprint's key-order insensitivity (the workspace enables `serde_json/preserve_order`, so two logically identical schemas are genuinely different `Map`s) and array-order sensitivity.

Two couplings a refactor elsewhere could break silently are pinned by behaviour and named in the test module doc: the `"Bad request:"` error-string spelling the demotion matches by prefix, and 501's terminality.

**Not covered:** the ticket asks for "a live check that demotion fires on Baseten's 501". That needs a Baseten credential and is not something CI can assert; the 501 path is covered against a mock instead, and the real-world claim rests on the MEASURED evidence already recorded in the code.

## Verification

`scripts/check_all.sh`: every Rust stage green (fmt, clippy `-D warnings`, all-targets check, the telemetry / no-default-features / Postgres-only / slim-CLI / wasm32 / PDF pure-rust / lancedb lanes), C API green. The gate then stops at the known environment failure — `python/scripts/check.sh` calls bare `pytest`, exit 127 — so the last three stages were run by absolute path: Python **128 passed / 100 skipped**, TS and Java as noted below.

Targeted: 429 `cognee-llm` + `cognee-components` tests, 367 all-features unit tests, 19 Bedrock tests under `--features bedrock` (that suite is `#![cfg(feature = "bedrock")]` with no `required-features`, i.e. SDK-666 — run explicitly so the shared-traversal change is actually exercised rather than compiling to a green zero).


GraDKh:
Review round addressed in `bdde7827`. Six findings were real and are fixed; one is answered rather than changed.

**Fixed**

- **Concurrent probes had no single-flight** (`openai.rs:253`). Correct, and it made the "one wasted request per distinct schema per process" claim in the docs false under the exact workload the feature is for — cognify shares one adapter and one schema across a wave of chunks, so every task read "not yet tried" and walked its own ladder. A fingerprint-keyed reservation now bounds it to one probe; the rest of the wave uses the cascade for that call. `JsonSchemaSupport` gained an `Unknown` state to distinguish "never sent" from "sent and accepted" (same request, but only the first needs coordinating), success is now recorded, and the reservation is released by a `Drop` guard so a `?`, a panic or a dropped future cannot leak it and disable the mode for the process.

  Worth noting for the reviewer: the obvious test for this would not have worked. Against a refusing endpoint, "one strict request" is also what six *serialised* calls produce, so `concurrent_callers_share_one_probe` makes the probe **succeed** instead — then six-strict, one-strict-plus-five-cascade, and never-overlapped all separate.

- **`default` left in the strict schema** (`schema.rs:295`). Fixed — stripped at every node. The reasoning is slightly stronger than "OpenAI rejects it": under strict decoding every property is `required`, so a default can never apply, which makes the keyword dead weight even where it is tolerated. Note the deliberate asymmetry with the shallow `recompute_top_level_required`, which reads a literal `default` as the marker for *optional* — correct there (it reproduces instructor's non-strict rewrite), wrong here.

- **Tuple positions not traversed** (`schema.rs:230`). Fixed — `prefixItems` and array-valued `items` both recurse now, not just an object-valued `items`.

- **Bare `{"type": "object"}` skipped** (`schema.rs:295`). Fixed — a property-less object node is closed. One unreached node is enough to put the document outside the subset and get the request rejected, which the memo would then have remembered as "this endpoint refuses constrained decoding".

- **Memo shared across a clone that retargets `api_version`** (`openai.rs:448`). Fixed by resetting on `with_api_version` rather than by widening the key — the api-version is part of the endpoint's identity and both memories describe an endpoint, so starting fresh is always safe and the worst case is re-discovering what was known. `cascade_probe` is reset alongside it: the same argument applies to it verbatim and it was already shared this way before this PR.

- **Two stale doc contracts** (`config.rs:151`, `context.rs:502`). Both fixed — `LlmInputs::structured_output_mode` and `Settings::set_llm_structured_output_mode` each still said every non-`auto` value pins.

- **`docs/configuration.md` "a 5xx"** — reworded to "any **other** 5xx", since 501 is itself a 5xx and is the documented fallback case.

**Answered, not changed: the 400 arm is broad on purpose** (`openai.rs:317`).

The finding is accurate — any 400 demotes, not only one naming `response_format`. It is deliberate, and the alternative is worse. Narrowing means classifying provider error prose, and a shape-rejecting server whose 400 does not name the field would then never demote and would re-probe on *every* call: unbounded waste on a deployment that is otherwise working, which is the exact failure this ladder exists to prevent.

The over-breadth costs little in exchange, because the constrained request differs from the cascade's **only** in `response_format`. A 400 caused by anything else — an invalid budget, a malformed message — fails every other mode too, so that call was lost whatever this predicate returns; the only cost is a poisoned memo entry in a process that is already failing, and the memo is process-local. Reasoning is now recorded on `is_request_shape_rejection` and in `docs/configuration.md` rather than left implicit.

GraDKh:
Round two addressed in `ad71c18b`. Five real findings fixed, one re-raised item answered again.

**The mid-ladder race is the important one, and the finding is exactly right.** The reservation was taken only while the level read `Unknown`, and the owner records its demotion *while still holding the guard* — so a caller arriving between the two rungs saw a settled state, skipped `try_begin_probe`, and sent its own non-strict probe. Against an endpoint that refuses both rungs that is the same burst the reservation exists to prevent, one rung later.

Fixed by giving the state machine the distinction it was missing: `NonStrictPending` (strict refused, demoted rung not yet tried) is now separate from `NonStrict` (demoted rung confirmed working), and the reservation keys off `is_discovering()`, which covers both incomplete states rather than just the first. Settled states still take no reservation, so once a schema is discovered concurrent calls go out at full rate.

**The corrective re-ask named a function that was not in the request** — also right, and I had missed it. Every retry went through the shared "call the `extract_structured_data` function again" wording, but a `response_format: json_schema` request carries no tool and no function. It now uses the shared *preamble* `corrective_reason_detail` — which exists for exactly this case, since Bedrock's native `outputConfig` branch has no tool either — with a directive the request can satisfy.

**And the point about the refusal text never being exercised was well made:** with `structured_output_retries(1)` the mode falls through before a second request is built, so the text could have been dropped entirely with every test still green. `a_retry_carries_the_failure_reason_and_names_no_function` uses two attempts and requires the refusal verbatim in the follow-up body, plus asserts the function wording is absent — covering both findings in one case.

**`required` order in the fingerprint** — agreed, it is a set and the old test was locking in the extra probing. Now sorted before hashing; every other array keeps its order, since `anyOf` precedence and `prefixItems` positions are significant. Added a case that a normalised `required` cannot collide with a same-valued `enum`.

**Stale `required` with empty `properties`** — fixed, the entry is dropped rather than left to demand a property the model is forbidden to emit. Worth noting this one had a second-order cost: the rejection would have been memoised as the *endpoint* refusing constrained decoding.

**Lazy fingerprint** — fixed; it was hashed on every structured-output call including the default `auto` path and the three cascade pins, none of which consult the memo.

**`call_api`'s retry-contract doc comment** — fixed; it still said all 5xx are retried and only 400/401 are immediate.

**Unchanged, and answered in the previous round: the 400 classifier.** Narrowing it means classifying provider error prose, and a shape-rejecting server whose 400 does not name `response_format` would then never demote and would re-probe on *every* call — unbounded waste on a working deployment, which is the failure this ladder exists to prevent. The over-breadth costs little in return because the constrained request differs from the cascade's only in `response_format`: a 400 from anything else fails every other mode too, so the call was lost regardless, and the memo is process-local. Recorded on `is_request_shape_rejection` and in `docs/configuration.md`.

Verification on this commit: 445 `cognee-llm` + `cognee-components` tests, 31 Bedrock tests under `--features bedrock`, clippy `-D warnings` clean.

GraDKh:
A `/code-review` pass raised five findings. Four were real and are fixed in `91c85716`; one is a documentation correction rather than a code change.

**The map-clobbering one is the serious find, and it was a silent-wrong-data bug.** `strict_json_schema` set `additionalProperties: false` unconditionally, including over the *schema* form of that keyword — which is how JSON Schema spells a map (`HashMap<String, String>` → `{"type": "object", "additionalProperties": {"type": "string"}}`). The rewritten document is one the provider **accepts** and which forbids every key, so the model returns a valid, permanently empty map: no rejection, so nothing demotes, nothing in the logs, wrong data. It is reachable from any caller-supplied schema, including cognify's configurable `summary_schema`.

A schema-valued `additionalProperties` is now left alone. That leaves the node outside OpenAI's strict subset, so the request is refused, the ladder demotes, and the cascade returns the right answer. Between a silent wrong result and a noisy fallback, the fallback is the one to take — and that principle is now written on `close_object_node`, because it is the reason the traversal should *not* simply be widened (see below). A boolean `additionalProperties: true` is still overwritten; it is not satisfiable alongside `strict` and keeping it buys nothing.

**The 400 argument was wrong, and I had defended it twice.** My reasoning was: the constrained request differs from the cascade's only in `response_format`, so a 400 from anything else fails every mode and that call is lost regardless. That covers the failing call and misses the ones after it. The memo is monotonic and process-wide, so one oversized chunk answered `400 context_length_exceeded` in a long-lived server strips constrained decoding from every later extraction — including the smaller chunks that would have succeeded. A silent quality regression, not a lost call, and on the measured numbers that is the 8-of-11 vs 5/5 difference.

A 400 now moves the ladder only when its body names the thing being refused (`response_format` / `json_schema` / `schema` / `strict`). The residual is the opposite risk I was worried about — a server refusing the shape without naming it gets re-probed each call — and that is the narrower one, since the refusing backend this feature exists for answers **501**, which is typed and needs no prose matching at all.

**Per-rung reservation.** The reservation was taken once before the ladder, so a demotion *away from a settled state* (api-version rollover, gateway upgrade) had every concurrent caller walk the whole ladder itself. It is now re-checked at the top of each rung. The new test states the limit honestly rather than overclaiming: the first rung's burst is **irreducible** — nothing can know the endpoint changed until a request comes back refused — so six concurrent callers send six strict requests; what the reservation prevents is that burst being doubled, and the demoted rung goes out exactly once. `docs/configuration.md` now says this too, since the old "one wasted request per distinct schema, once" was only true of the cold path.

**Non-object schema root.** `routers::llm::string_response_schema` is `{"type": "string"}`, and `response_format.json_schema` requires an object root, so both HTTP LLM routes were burning two rungs on a request invalid by construction. Now skipped outright rather than demoted — there is no endpoint behaviour to remember.

**Nullable object nodes** are closed: `declares_object` reads the array form of `type`, which is what schemars emits for a nullable object.

**Not changed: widening the traversal** (`patternProperties`, `not`, `if`/`then`/`else`, `contains`, `propertyNames`, `dependentSchemas`, the schema form of `additionalProperties`). The finding is accurate that they are unvisited, but reaching into them cannot make such a schema strict-valid — OpenAI's strict subset does not accept those keywords at all — so it would only change *which* part the provider refuses. A schema using them is meant to be refused and demoted to the cascade, which answers it correctly. The map case above is the proof that "close every node we can reach" is the wrong goal: doing it there produced an accepted request with the wrong meaning. `rewrite_object_nodes`'s docstring now states the covered set explicitly and gives this reasoning, instead of reading as though it were exhaustive. The note that Bedrock shares the same gap pre-existing this PR is also recorded there.

Verification on `91c85716`: 450 `cognee-llm` + `cognee-components` tests, 31 Bedrock tests under `--features bedrock`, clippy `-D warnings` clean. Four new cases cover the map, the nullable object, the unrelated 400, the non-object root, and the settled-state transition.

GraDKh:
Copilot round three addressed in `71983582`. All six findings were real and are fixed — nothing argued away this time.

**The rung decision was not atomic** (`openai.rs:2281`). Correct, and it undercut the guarantee the previous commit added: reading the level and reserving were two separate lock acquisitions, so a caller could read a settled `Strict`, be descheduled while another recorded `NonStrictPending` and took the reservation, then re-read `NonStrictPending` and send a second, unreserved probe. `levels` and `probing` now share one mutex, and `JsonSchemaMemo::begin_rung` is the only entry point — it reads the level and reserves inside a single critical section, returning `Send { strict, guard } | Deferred | Unsupported`. The ladder can no longer observe a level it did not also reserve against.

**Unbounded memo growth** (`openai.rs:281`). Also correct: the map is keyed on caller-supplied schemas and `create_structured_output_raw` takes an arbitrary `Value`. Capped at 1024 distinct fingerprints — roughly 200x any real workload, since cognify uses a handful — and past the cap the map is dropped and rebuilt rather than evicted from. There is no useful recency signal to order an eviction by (every entry is equally true), and the cost of being wrong is a re-probe, so the crude reset is the right shape. It warns when it fires, because reaching it means a caller is generating a distinct schema per request.

**`additionalProperties: true` was being closed** (`schema.rs:399`). This is the same silent-loss shape as the map case from the last round, and I had explicitly justified the boolean carve-out — wrongly. `true` says the author wants keys beyond `properties`; rewriting it to `false` gives a request the provider *accepts* while dropping exactly those keys. Only an **absent** keyword is filled in now, which has the pleasant side effect of making the strict rewrite identical to the Bedrock one it shares a traversal with. Both regression tests updated to assert the preservation.

**`object_root` inferred objecthood from `properties`** (`openai.rs:2247`). Right, including the specific pointer to `validate_summary_schema` accepting a type-less schema. `properties` constrains an instance only *if* it is an object and asserts nothing on its own, the strict rewrite preserves the omission, and the probe would be refused anyway. Of the two options offered I took the conservative one — require the root to declare itself, via `schema::declares_object_root` — because synthesising the missing `type` would narrow a schema its author chose to leave open, which is precisely the mistake `close_object_node` now exists to avoid.

**Two stale contracts**, both fixed: the `structured_output_mode` field rustdoc still said every non-`Auto` value pins, and `.env.example` opened by describing the knob as a pin two lines before explaining that one of its values is not one.

New unit coverage for `begin_rung`'s reservation states (first caller reserves, second deferred, owner keeps its guard across the demotion, settled levels take none, `Unsupported` short-circuits) and for `record`'s monotonicity and cap behaviour.

Verification on `71983582`: 452 `cognee-llm` + `cognee-components` tests, 31 Bedrock tests under `--features bedrock`, clippy `-D warnings` clean.

GraDKh:
Copilot round four addressed in `78eac81a`. Four findings, all real, all fixed.

**Nullable root admitted to constrained decoding** (`schema.rs:379`). Right, and it is a distinction I collapsed when I introduced `declares_object_root` last round: I reused the nested predicate, which accepts `{"type": ["object", "null"]}`. That is correct *inside* a document — a nullable object node still has to be closed — and wrong at the root, where it permits a bare `null` response that structured outputs do not accept. The root now has to say `"type": "object"` and nothing else; the nested predicate is untouched, and the two are documented against each other so the next reader does not re-merge them.

**`{"type": "object", "required": ["x"]}` with no `properties`** (`schema.rs:345`). Correct — I fixed this for *empty* `properties` two rounds ago and left the *absent* case one branch away doing the opposite. Closing the node while keeping the entry demands a key the model is forbidden to emit, so the provider rejects it and the schema is demoted for a fault the rewrite introduced rather than one it had. Both branches now reconcile, and the regression test covers both spellings.

**`force_additional_properties_false` skipped nullable object nodes** (`schema.rs:295`). Also right, and it affects Bedrock rather than the new mode: the helper matched only the string form of `type`, so a nullable object was left open in violation of its own every-object-node contract, and Bedrock's native schema validation is what rejects that. It now shares `declares_object` and `close_object_node` with the strict rewrite — which has the further benefit that an explicit `additionalProperties` is preserved on both paths for the same reason, rather than the two drifting again. Verified against the Bedrock suites including the SigV4 golden tests (42 passing).

**501 was only observed indirectly** (`openai.rs:1771`). Fair: mapping it to `FeatureNotSupported` is a public `generate` contract, and the json_schema tests only see it through the fallback, so a regression could have kept demotion working while `generate` started retrying or returned a different terminal error. `assert_terminal` now returns the error so callers can assert the variant, and 501 joins the terminal matrix in `retry_parity.rs` alongside 401/402/404 with an explicit `FeatureNotSupported` assertion.

Verification on `78eac81a`: 455 `cognee-llm` + `cognee-components` tests, 42 Bedrock tests under `--features bedrock`, clippy `-D warnings` clean.

PR #239, opened 2026-09-18
feat(cognify): add a ChunkGraphExtractor seam for non-LLM graph backends

GraDKh:
## What

Adds a pluggable seam so graph extraction is no longer hard-wired to the LLM-backed
`FactExtractor`, plus the ontology surface such a backend needs to build its schema.

Two coherent halves:

1. **`cognify`** — a `ChunkGraphExtractor` trait, injected through `CognifyConfig` exactly the
   way `Transcriber` already is, with two branch points in `tasks.rs`.
2. **`ontology`** — `OntologyTerms`, exposing OWL classes and object properties with their
   labels and comments, behind a defaulted `OntologyResolver::terms()`.

## Why

Today `cognify` can only build a graph by calling an LLM. There is no way to run extraction
offline, in a data-residency-constrained deployment, or with a local NER model without forking
`tasks.rs` — and forking it means reimplementing the abort partition, failure accounting, edge
deduplication and graph writes that live around the extraction call, none of which are
extraction-specific.

This PR moves the extraction call behind a trait and leaves everything around it untouched.

The ontology half stands on its own: `build_lookup` already walks the triples, but only surfaces
classes and individuals, and `uri_to_key` lowercases and replaces spaces, so `worksAt` becomes
`worksat`. Anything that needs the original URI, label or comment — schema generation, docs,
UI — currently cannot get them.

## Why this ships with no in-tree implementor

The only implementor in this PR is `MockChunkGraphExtractor`. That is deliberate, and the
module docs say so under *"Why this ships with no in-tree implementor — and must not be
deleted"*.

Being straightforward about it: the first real implementor is out-of-tree. That is already the
pattern for premium adapters in this project, and the trait is written to be backend-neutral —
nothing in it assumes a particular model, runtime, or vendor.

The risk with an extension point that has no in-tree user is that someone later reads it as dead
code and removes it. `CognifyConfig::custom_chunker` is the cautionary case in this very repo: a
config field with a builder and **no read site anywhere**. The module docs make the distinction
explicit — *unreachable* versus *implemented elsewhere* — and the mock is a genuinely exercised
implementor, so the seam cannot silently rot.

If reviewers would rather see a reference implementation land alongside it, a rule-based
extractor over the ontology terms is a reasonable follow-up and the author is happy to write it.

## Public API and stability

Five new or changed public types carry `#[non_exhaustive]`, each with a `# Stability` rustdoc
block following the `EdgeResolutionStats` precedent:

| Type | Rationale |
|---|---|
| `ExtractedGraphData` | already taking a field addition here; constructed only inside the crate |
| `GraphBackendError` | on the enum, not its variants — backends still construct them freely |
| `ChunkRef`, `ExtractionContext` | plus `new()`, since out-of-tree backends must build both to unit-test |
| `OntologyTerms` | plus `new()`; it is a return type, so the attribute *protects* implementors — adding `owl:DatatypeProperty` later leaves their call sites compiling |

## Backwards compatibility

- `OntologyResolver::terms()` is a **default method**, so every existing implementor keeps
  compiling unchanged.
- `ExtractedGraphData` gains `backend_summaries`, and is constructed only within `cognify`.
- No behaviour changes when no backend is configured: the LLM path is byte-for-byte the one that
  runs today.

## Testing

`cargo test`: **577 passed, 0 failed, 13 ignored** across 43 targets, including two new suites —
`graph_backend_seam.rs` (11) and `ontology_terms_test.rs` (10), neither with ignored tests.

The seam tests assert the things that would otherwise rot silently: summary identity is checked
by recomputing `uuid5(chunk_id, b"TextSummary")` independently rather than by re-calling
`TextSummary::new`, and the "no LLM was called" assertion is proven non-vacuous by companion
tests that assert the same counter reads `1` and `> 0` on the LLM path.

`cargo clippy -D warnings` clean; `cargo doc` adds no new warnings.

## Known limitations, stated rather than hidden

- `max_parallel_extractions` is documented as LLM-path-only — the backend seam is sequential.
  Making it concurrent is deliberately out of scope here.
- `MockChunkGraphExtractor` is compiled unconditionally rather than behind `feature = "testing"`.
  Gating it would force either `required-features` on the seam suite, removing its only
  end-to-end coverage by default, or a self dev-dep that re-arms the suites #238 just gated off.
  The reasoning is recorded in the module docs.

## Note for reviewers

This is two separable halves in one branch. The ontology commits stand alone and could be split
into their own PR if you would prefer to review them independently — say the word and I will
split it.


GraDKh:
## Review fixes pushed — the trait's failure granularity changed

Three findings addressed in `d4fe581`, `4a71e98`, `fabb983`. **If you are mid-review of the diff, re-read `crates/cognify/src/graph_backend/mod.rs` and the `extract_graphs_via_backend` loop in `tasks.rs` first** — the trait's public signature is different from the one that was reviewed.

### F1 (medium/high) — `extract_graphs` returned one `Result` per *batch*

This was the real one. `DEFAULT_CHUNKS_PER_BATCH` is 2000, so a realistic dataset is a single batch, and `tasks.rs` charged `fails_item: true` to **every chunk in the batch** on an `Err`. One transient error — one bad chunk, one session fault — marked every document in the run as failed, where the LLM path loses exactly one chunk. It also meant the abort-time partition this PR advertises as "identical to the LLM path" could never preserve anything: with every file in the batch failed, the "complete" side of the three-way split was always empty.

The batch size is pre-existing and documented (`docs/configuration.md`, `tests/failure_at_default_batch.rs`); the coarse granularity was ours. So the trait changed, not the batch size:

```rust
async fn extract_graphs<'c, 'x>(
    &self,
    chunks: &[ChunkRef<'c>],
    ctx: &ExtractionContext<'x>,
) -> Result<Vec<ChunkGraphResult>, GraphBackendError>;

pub type ChunkGraphResult = Result<KnowledgeGraph, ChunkExtractionError>;
```

Two granularities, deliberately distinct:

* **Per chunk** — `Err(ChunkExtractionError)` in that chunk's own slot. A new `#[non_exhaustive]` error struct with a `new` constructor (backends build it, the seam reads it). One `StageFailure` against that chunk and its document; siblings in the same batch keep their graphs and are persisted.
* **Whole batch** — the outer `Err(GraphBackendError)`, unchanged in meaning: the call could not be served at all (model won't load, runtime missing). Still charged to every chunk in the batch. `ArityMismatch` keeps its separate hard-error path at the top of the match, exactly as before.

**The arity contract is untouched and still enforced.** A failed chunk *occupies* its slot with an `Err` rather than being omitted, so `results.len() != refs.len()` is still an `ArityMismatch`, and the positional zip onto the batch is still load-bearing. The seam now accumulates a `batch_failed` flag across both granularities and checks FailFast once after the batch is fully accounted for — the same "collect the whole batch, then decide" shape the LLM loop already had, so a FailFast abort still reports every failure in the batch that tripped it.

`MockChunkGraphExtractor` gains `failing_chunk(chunk_id)` for the per-chunk shape, alongside `set_failure_after(n)` for the whole-batch one. Two new integration tests pin the new behaviour at the **default** batch size, which is the whole point:

* `one_failing_chunk_does_not_fail_its_batch` — three one-chunk files, one batch (asserted: `extract_calls() == 1`), middle chunk fails. Exactly one `StageFailure`, charged to that file only; `alpha` and `gamma` still reach the entity list, `beta` does not.
* `the_abort_partition_keeps_the_batch_siblings` — same batch under FailFast + FailedItems: the partition now keeps the two complete files.

Both were run against a simulated pre-fix seam (results collapsed back to batch granularity) and fail there: 3 `StageFailure`s instead of 1, and 0 surviving files instead of 2. `backend_failure_is_recorded_per_chunk` was renamed `whole_batch_backend_failure_is_charged_to_every_chunk` — it now pins the outer-`Err` path specifically, and its doc says why charging the whole batch is correct *there*.

Rustdoc updated throughout, including the `# Stability` blocks: the module docs gain a "The two failure granularities" section, `GraphBackendError` says what it is *not* for, and the crate README carries the new signature.

### F10 (low) — `summarize_text`'s short-circuit vs. `make_summarize_text_task`'s docs

`summarize_text` keys its short-circuit off `config.graph_backend`; `make_summarize_text_task` advertised itself as fusable "against a different sibling". In that combination a summarizing backend is configured, the LLM summarizer switches off, and nothing produces `backend_summaries` — every summary vanishes behind one `info!`.

The short-circuit cannot detect its sibling, and that is by construction: the two branches run concurrently under `try_parallel` and cannot negotiate, which is why the decision is per run rather than per chunk in the first place. So the advertisement is what goes. The supported pairings are now named explicitly — fused by `make_extract_graph_and_summarize_task`, or as a separate stage next to `make_extract_graph_task` — and both put the task alongside `extract_graph_from_data`. `CognifyConfig::graph_backend` now states that it means "this run's extraction stage is the backend" and that `extract_graph_from_data` is its only reader. The short-circuit's `info!` names the backend and the chunk count and says in-line that the extraction stage must be running that same backend, so a run that produces no summaries leaves a traceable record of why.

### F11 (low) — `backend_summaries` ordering

It was documented as "in `chunks` order", which reads as positional parity with the public `chunks` field next to it. It is sparse: DLT chunks never reach the backend, failed chunks have no graph to summarize, and a declined chunk (empty `summarize_chunk`) records nothing. The doc now says so, states `backend_summaries.len() <= chunks.len()`, and names `TextSummary::made_from` as the way to associate a summary with its chunk (`base.id` is `uuid5(chunk_id, b"TextSummary")`, so the id works too).

### Verification

`cargo test -p cognee-cognify -p cognee-ontology`: **580 passed, 0 failed, 13 ignored** across 43 targets (was 577/0/13 — +3: two seam tests and one mock unit test; `graph_backend_seam` 11 → 13). `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean, `cargo doc` still 24 warnings, all pre-existing.

### Note for downstream implementors

`extract_graphs` is a breaking change for anyone who has already written a backend: wrap each produced graph in `Ok(...)`, and return `Err(ChunkExtractionError::new(name, reason))` in a chunk's slot instead of failing the whole call. Our own out-of-tree GLiNER backend is being updated accordingly.

